PR #1, opened 2020-11-26
Initial implementation of the operator-rs library

lfrancke:
This is bare-bones!
It mostly consists of extracted methods from the zookeeper-operator that can be reused.
The ergonomics could be a lot better but it's a starting point.


src/finalizer.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Adds our finalizer to the list of finalizers.
+pub async fn add_finalizer<T>(
+    client: Client,
+    resource: &T,
+    finalizer: &str
+) -> Result<T, Error>
+    where T: k8s_openapi::Resource + Clone + Meta + DeserializeOwned
+{
+    let name = &Meta::name(resource);
+    let namespace = &Meta::namespace(resource).expect("Resource is namespaced");
```
soenkeliebau, line 27: Is this error message correct? To me this is triggered, when the namespace option is None i.e. the resource doesn't have a namespace..
Also, I am hesitant to panic in libraries, shouldn't we let the binary decide how to handle this error?

src/finalizer.rs
```diff
@@ -0,0 +1,113 @@
...
+    // It would be preferable to use a strategic merge but that currently (K8S 1.19) doesn't
+    // seem to work against custom resources.
+    // This is what the patch could look like
+    // ```
+    //         "metadata": {
+    //             "$deleteFromPrimitiveList/finalizers": [FINALIZER_NAME.to_string()]
+    //         }
+    // ```
+    let name = Meta::name(resource);
+    let namespace = Meta::namespace(resource).expect("Custom Resource is namespaced");
```
soenkeliebau, line 65: See above

src/finalizer.rs
```diff
@@ -0,0 +1,113 @@
...
+                // And then remove the finalizer from the list.
+
+                finalizers.swap_remove(index);
+                let new_metadata = serde_json::to_vec(&json!({
+                    "metadata": {
+                        "finalizers": finalizers
+                    }
+                }))?;
+
+                let patch_params = PatchParams {
```
soenkeliebau, line 88: Would we potentially restore a finalizer here that was present when we retrieved our finalizer list, but has been deleted in the meantime, or is this handled by the Kubernetes api to ensure we only update the same version of the object that we retrieved?

soenkeliebau (changes requested):

src/crd.rs
```diff
@@ -0,0 +1,117 @@
...
+        const RESOURCE_NAME: &'static str = "tests.test.stackable.de";
+        const CRD_DEFINITION: &'static str = "FOOBAR";
+    }
+
+    // TODO:
+    #[test]
+    fn test_exists()  {
+        let client = tokio_test::block_on(kube::Client::try_default()).expect("Client creation should not fail");
+        let result = tokio_test::block_on(exists::<Test>(client));
+        assert_err!(result);
```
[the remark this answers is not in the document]
lfrancke: Yes...this is where my question came from on how to mock a Client :(
I don't have a good solution here I'm afraid

src/finalizer.rs
lfrancke: Yes to the latter. We post the resourceVersion we based our patch on and if that doesn't match the current one it'll fail (and hopefully retry?)

src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+        Some(ControllerAction::Update)
+    } else {
+        // The object is being deleted but we've already finished our finalizer
+        // So there's nothing left to do for us on this one.
+        None
+    };
+}
+
+/// Creates a vector of tolerations we need to work with the Krustlet
+pub fn create_tolerations() -> Vec<Toleration> {
```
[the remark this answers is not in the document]
lfrancke: Probably a good idea, yes.
This could be that common library though? What do you think?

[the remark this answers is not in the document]
soenkeliebau: Fine for me, yes. I initially thought that we'd be pulling a lot of stuff in that for example the agent doesn't really need, but have no really strong opinion either way to be honest.

src/client.rs
```diff
@@ -0,0 +1,98 @@
...
+        Client {
+            client,
+            post_params: PostParams::default(),
+            patch_params: PatchParams::default(),
+        }
+    }
+
+    /// Returns a [kube::client::Client]] that can be freely used.
+    /// It does not need to be cloned before first use.
+    pub fn kube_client(&self) -> KubeClient {
```
soenkeliebau, line 29: I've looked at https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv and won't pretend to fully understand it, but could we maybe rename this to .into_inner() or .into_kube_client() ?
I _think_ that into usually consumes the wrapping object though, maybe _as_ is more appropriate here?

src/client.rs
```diff
@@ -0,0 +1,98 @@
...
+        T: Clone + DeserializeOwned + Meta + Serialize,
+    {
+        self.get_api(Meta::namespace(resource))
+            .replace(&Meta::name(resource), &self.post_params, resource)
+            .await
+            .map_err(Error::from)
+    }
+
+    /// Returns an [kube::Api] object which is either namespaced or not depending on whether
+    /// a resource is passed in or not and whether that has a namespace or not.
```
soenkeliebau, line 74: Is the comment about the resource having a namespace or not triggering whether the API is namespaced correct? It look to me like the parameter controls that?

src/client.rs
```diff
@@ -0,0 +1,98 @@
...
+    where
+        T: Meta,
+    {
+        match namespace {
+            None => self.get_all_api(),
+            Some(namespace) => self.get_namespaced_api(&namespace),
+        }
+    }
+
+    fn get_all_api<T>(&self) -> Api<T>
```
soenkeliebau: nit: is this really worth extracting into a function of its own?

src/client.rs
```diff
@@ -0,0 +1,98 @@
...
+    where
+        T: k8s_openapi::Resource,
+    {
+        Api::all(self.client.clone())
+    }
+
+    fn get_namespaced_api<T>(&self, namespace: &str) -> Api<T>
```
soenkeliebau: nit: is this really worth extracting into a function of its own?

src/client.rs
```diff
@@ -0,0 +1,98 @@
...
+    client: KubeClient,
+    post_params: PostParams,
+    patch_params: PatchParams,
+}
+
+impl Client {
+    pub fn new(client: KubeClient) -> Self {
+        Client {
+            client,
+            post_params: PostParams::default(),
```
soenkeliebau, line 22: Do we expect at some point to need different sets of params with the same client? If yes, maybe add an Option<...> to the interface already so we do not need to change it later on?

soenkeliebau (approved):
Left some comments, nothing major, feel free to address those or merge as is.

src/crd.rs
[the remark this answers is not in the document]
soenkeliebau: We'll come up with something.

src/client.rs
lfrancke: I don't think it's too important for post & patch, no. For List maybe. I'd keep it like this for now.

src/client.rs
lfrancke: Hmm I have to admit I forgot about into_* but I don't think that makes sense here. But maybe as_kube_client or to_kube_client. Preferences?

src/client.rs
lfrancke: Comment is outdated, thanks for catching that.

src/client.rs
soenkeliebau: Based on the docs I would say that "as" is correct, as this decreases the level of abstraction. "to" would indicate staying at the same level of abstraction and just changing the representation.

soenkeliebau (approved):
lgtm

PR #3, opened 2020-12-04
Addresses build problems

lfrancke:
Github Actions fails because of duplicate features in clippy check and a bad doc test.

This happened because the first commit only added the Github Actions definition.


soenkeliebau (approved):
lgtm

PR #7, opened 2020-12-09
Support list and two different patch strategies

lfrancke:


src/client.rs
```diff
@@ -11,16 +11,29 @@ use serde::Serialize;
 #[derive(Clone)]
 pub struct Client {
     client: KubeClient,
+    merge_patch_params: PatchParams,
```
soenkeliebau, line 14: I think I commented on this before, but can't really remember... 
Is it possible that we'll need to support additional patch / list / ... parameter sets? If so, would it maybe make sense to not add these here, but insert a  register_params(id: &str, params: PatchParams) method, internally store those in a HashMap and reference them when patching? It is not really an issue, because all of this is abstracted away by the new method.
Just a thought, in case there is a chance that we have to add a third set of params to this in a few weeks time.

soenkeliebau (approved):
LGTM overall, left one question

PR #8, opened 2020-12-11
Introduce result type

lfrancke:
This introduces a "OperatorResult" type which uses our own Error type.
I decided to name it `OperatorResult` instead of reusing the `Result` name.
Personally, I find it confusing to have `Result` mean different things in different places and I tend to get confused.

This also replaces `map_err` with a different way to returning our errors.

This PR depends on #7 so that needs to be merged in first!


soenkeliebau (approved):
LGTM

PR #9, opened 2020-12-11
Adds a `delete` method.

lfrancke:
There are some things (like Pods) that return the object itself when delete is called, others return a `Status` object instead.
So we have to return an `Either` here and the client will have to know what to expect.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,5 +1,6 @@
 use crate::error::OperatorResult;
-use kube::api::{ListParams, Meta, PatchParams, PatchStrategy, PostParams};
+
+use kube::api::{DeleteParams, ListParams, Meta, PatchParams, PatchStrategy, PostParams};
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -120,6 +123,22 @@ impl Client {
             .await?)
     }
 
+    /// Which of the two results this returns depends on the API.
+    /// Take a look at the Kubernetes API reference.
```
soenkeliebau, line 127: Nit: maybe add a link?

soenkeliebau (approved):
LGTM, just added one suggestion

src/client.rs
lfrancke: I would but I can't find a stable link that doesn't link to a specific version.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/ there is no 'latest' or anything like that.

PR #10, opened 2020-12-11
One minor one major fix

lfrancke:
Minor: This allows using the `object_to_owner_reference` method with more objects.
`Meta` requires `Resource` but we only need `Resource`.

Major: This also fixes ConfigMap creation to use the namespace of the resource that's passed in.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@ use crate::error::{Error, OperatorResult};
 pub use crd::CRD;
 use k8s_openapi::api::core::v1::{ConfigMap, Toleration};
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::OwnerReference;
+use k8s_openapi::Resource;
 use kube::api::{Meta, ObjectMeta, PatchParams, PatchStrategy};
 use kube::Api;
 use kube_runtime::controller::Context;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -98,7 +99,7 @@ pub fn create_tolerations() -> Vec<Toleration> {
     ]
 }
 
-pub fn object_to_owner_reference<K: Meta>(meta: ObjectMeta) -> OperatorResult<OwnerReference> {
+pub fn object_to_owner_reference<K: Resource>(meta: ObjectMeta) -> OperatorResult<OwnerReference> {
```
soenkeliebau, line 102: I won't pretend to fully understand the difference, but could you just briefly comment on why the other Metas don't need to be substituted by "Resource"?

soenkeliebau (commented):
Left one quick question

src/lib.rs
lfrancke: The only thing we access from `K` are the associated consts `API_VERSION` and `KIND` and they are from `Resource`.

`Resource` is a supertrait for `Meta` so this works as well but we don't need to be as _tight_.
Every `Meta` is also a `Resource` but not every `Resource` does have to be `Meta`.

soenkeliebau (approved):
LGTM

PR #12, opened 2020-12-11
Adds some more utility functions for Pods

lfrancke:


src/podutils.rs
```diff
@@ -0,0 +1,176 @@
...
+    Unknown,
+}
+
+impl fmt::Display for PodPhase {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(self, f)
+    }
+}
+
+pub enum PodConditionType {
```
soenkeliebau, line 20: Is there a specific reason that Debug and Display are not derived/implemented for this?

soenkeliebau (approved):
LGTM, left just one minor question

src/podutils.rs
lfrancke: It's not actually used anywhere yet....I need to figure out how to use it and then it makes sense to add that, yes.

soenkeliebau (approved):
LGTM

PR #16, opened 2020-12-20
Flattens the code a bit by introducing early returns

lfrancke:
This was a suggestion from the Ferrous Rust Experts and it makes the code nicer to read.


soenkeliebau (approved):
LGTM

PR #17, opened 2020-12-20
Change CI to use the actions-rs actions everywhere and restructure a bit

lfrancke:


lfrancke:
Because the "old" checks are required to merge we need to manually disable that for this commit once it's approved.

soenkeliebau (approved):
LGTM 

full disclosure: i've not tested it, I assume we'll do that as we go along :)

PR #18, opened 2020-12-21
TIL about assert!

lfrancke:
This changes `assert_eq!(true, ...)` to `assert!(...)` and `assert_eq!(false, ...)` to `assert!(!...)`


soenkeliebau (approved):
LGTM

PR #21, opened 2020-12-21
Merge test & build step

lfrancke:
It probably makes more sense to run the release build only when the tests finished successfully.
Should we want to publish nightly builds this probably makes sense.


soenkeliebau (approved):
LGTM

lfrancke:
@soenkeliebau  I'm wondering: Does it make sense to run a build --release at all after a test run?

soenkeliebau:
I think something broke with the checks-not sure if just for this run..

PR #24, opened 2020-12-23
Let dependabot also check for updates in Github Actions

lfrancke:


soenkeliebau:
The github-actions module seems to still be in beta, I'm not saying we should not use it because of that, just an observation.

Could you please update the branch before I approve?

lfrancke:
Done, thanks

PR #42, opened 2021-01-28
Move to recommended "best practice" on imports.

lfrancke:
See https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#creating-idiomatic-use-paths

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,7 +1,9 @@
 use crate::error::OperatorResult;
 
+use either::Either;
+use k8s_openapi::Resource;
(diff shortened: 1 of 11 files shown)


maltesander (approved):
Ok!

lfrancke:
Thanks for the review @maltesander 
Unfortunately there was a conflict. I'll push a fix shortly.

soenkeliebau (approved):
LGTM at a quick glance.
I've built it and run clippy both of which work (apart from a clippy issue I missed in another PR).

PR #44, opened 2021-01-29
Cleans up a few things from the main crate

lfrancke:
Cleans up a few things from the main crate

These are methods that are not needed anymore. There are some left but they might be useful again. Probably though they can also be deleted.


maltesander (approved):

soenkeliebau (approved):
LGTM

PR #45, opened 2021-01-29
Adds a method to build an ObjectMeta object

lfrancke:
This also fixes a bug (not sure how that happened) in the object_to_owner_reference method.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,9 @@ tokio = { version = "0.2", features = ["macros", "rt-threaded"] }
 tracing = "0.1"
 tracing-subscriber = "0.2"
 
+[dev-dependencies]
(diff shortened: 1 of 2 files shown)


maltesander (approved):
Just reminder for all of us. Like clippy suggests with the warnings, we barely need "unwrap_or" and should default to "unwrap_or_else" (difference is when the "or" part is evaluated. With "unwrap_or" always, with "unwrap_or_else" only when the else part is happening. So if we only create default values like empty string etc it should always be unwrap_or_else for not creating objects we do not need.

lfrancke:
Yes, that's a good point. I believe I fixed that warning in my other open PR #44

soenkeliebau (approved):
LGTM

PR #46, opened 2021-02-02
Updates kube-rs to latest version

lfrancke:
This requires some major changes in the structure of the operator.


lfrancke:
@maltesander If you like you're welcome to review this but mostly I'm interested in your feedback about the controller changes.

If you could look at the documentation and the example here: https://github.com/stackabletech/operator-rs/pull/46/files#diff-1fd48ab302efbb91fe84507671007262db18094467e4ad707d036f60ff22e0ea

And then say if this makes sense to you. Should require only minimal changes in your operator I hope.

Cargo.toml
```diff
@@ -7,18 +7,19 @@ edition = "2018"
...
 serde_json = "1.0"
 serde_yaml = "0.8"
 thiserror = "1.0"
-tokio = { version = "0.2", features = ["macros", "rt-threaded"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
 tracing = "0.1"
 tracing-subscriber = "0.2"
 
 [dev-dependencies]
+k8s-openapi = { version = "0.11", default-features = false, features = ["v1_20"] }
```
soenkeliebau, line 22: This might be my lack of understanding, but why is this feature needed for dev but not otherwise?

src/controller.rs
```diff
@@ -1,3 +1,118 @@
...
+//!
+//! impl ReconciliationState for FooState {
+//!     type Error = String;
+//!
+//!     fn reconcile(
+//!         &mut self,
+//!     ) -> Pin<Box<dyn Future<Output = Result<ReconcileFunctionAction, Self::Error>> + Send + '_>>
+//!     {
+//!         Box::pin(async move {
+//!             self.test1()
```
soenkeliebau, line 78: It looks a bit weird the way that subsequent calls are at a different level than the first one due to them being chained after that one.
But that really is just an observation, no need to do anything about this!

My first instinct was to give the ReconcileFunctionAction an associated function to execute all of them, but most probably that would reintroduce the exact problem that you worked hard to avoid, so disregard this ;)

src/controller.rs
```diff
@@ -166,33 +285,20 @@ where
...
+            trace!(?duration, "Reconciler loop: Requeue");
+            return Ok(ReconcilerAction {
+                requeue_after: Some(duration),
+            });
+        }
+        Ok(action) => {
+            trace!("Reconciler loop: {:?}", action);
+        }
+        Err(err) => {
+            error!("Error reconciling [{:?}]", err);
```
soenkeliebau, line 301: Previously we requeued for an error, was this a conscious decision to do it differently?

soenkeliebau (dismissed):
LGTM overall, just some questions that shouldn't prevent merging, unless Lars says so :)

Cargo.toml
lfrancke: This comes straight from the docs of kube-rs.
The operator-rs library should not prescribe the Kubernetes version for all users but we need _some_ for our own integration tests etc.

src/controller.rs
lfrancke: I understand about the initial issue but I don't understand your proposed solution but I have a feeling that your hunch might be correct :) Lots of hunches though!

src/controller.rs
lfrancke: I'd like to say yes.


But that'd be a lie.

src/controller.rs
soenkeliebau: As I said, can be ignored :)

src/controller.rs
lfrancke: Fixed

soenkeliebau (approved):
LGTM

maltesander (approved):

PR #48, opened 2021-02-04
Remove FOSSA action

lfrancke:
It always fails and the bot still runs automatically. Not sure what this is good for.


soenkeliebau (approved):
LGTM

PR #49, opened 2021-02-04
Fix build_metadata method.

lfrancke:
It would take the name of the passed in object which makes for correct metadata but it doesn't help us at all.
We usually want to create metadata for a new "child" object that should take a new name.


soenkeliebau (approved):
LGTM

PR #54, opened 2021-02-09
Improve logging

lfrancke:
Create a span for each new reconcile, assign a unique id to each request and improve logging in general

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -308,9 +308,9 @@ where
                 ?duration,
                 "Reconciliation finished successfully (it is normal to see this message twice)"
             );
-            return Ok(ReconcilerAction {
+            Ok(ReconcilerAction {
                 requeue_after: Some(duration),
-            });
+            })
         }
         Ok(action) => {
             trace!(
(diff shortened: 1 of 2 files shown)


src/controller.rs
```diff
@@ -243,8 +244,12 @@ where
             .run(reconcile, error_policy, context)
             .for_each(|res| async move {
                 match res {
-                    Ok(o) => info!("Reconciled {:?}", o),
-                    Err(ref e) => error!("Reconcile failed: {:?}", e),
+                    Ok(o) => trace!(resource = ?o, "Reconciliation finished successfully (it is normal to see this message twice)"),
+                    Err(ref err) => {
+                        // If we get here it means that our `reconcile` method returned an error which should never happen because
+                        // we convert all errors to requeue operations.
+                        error!(?err, "Reconciliation finished with an error, this should not happen, please file an issue")
```
soenkeliebau, line 251: Random brainfart, we are calling for people to file issues in some places, should we keep track of these places somewhere and give them an identifier that people can reference or something similar?
Nothing to do about this here, just a thought to be discussed.

soenkeliebau (approved):
LGTM - just one note that we might think about but that is just because it occurred to me here

lfrancke:
Oh, I'm sorry, not sure why I added you as a reviewer. This is still a draft but it looks good so it can probably go in as well.

src/controller.rs
lfrancke: Yeah...we should reactivate our common logging thing but that's probably very much outdated by now. It's a good idea though.

Issue #55, opened 2021-02-10
list_pods needs to use a label selector to only get pods for a specific resource


lfrancke:
Currently this will probably mean that one operator deletes pods that don't even belong to it.

lfrancke:
I believe this is already fixed -> investigate if that's true

lfrancke:
I investigated, it's already fixed.

PR #56, opened 2021-02-10
This filters pods by owner reference of the current object.

lfrancke:
This is related to #55 which talks about using labels.
That, however, requires everyone to actually set labels so this is a first leightweight approach that fixes the issue itself but not the performance implications of fetching all pods and then filtering on the client side.


maltesander (approved):
I think we should still implement the label approach asap. Not much effort and lots of traffic reduction in production?

lfrancke:
Agreed. I lef the ticket for the label approach open. Thank you for the review.

PR #60, opened 2021-02-11
feature: ControllerStrategy::init_reconcile_state is now async and returns Result

lfrancke:


maltesander (approved):
I am not sure how we handle external dependencies so far. Since the operator framework will be the core of many many products, maybe we should evaluate external dependencies a little more in terms of future development and prospect (async_trait is currently version 0.1.42).
I did not go into it deeper, whether there is no clean "rust" way and this is the only option we have, then we have no choice. Since rust is still developing, smaller (3rd party) repositories may stop the support/development which may break the operator's upwards compatibility. Since we always want to adapt to major rust releases, this can lead to a lot of adaptations in the operator (and every product depending on it).

lfrancke:
Yeah this is a limitation in Rust unfortunately and `async_trait` is the accepted workaround. It's used in a lot of projects and written by the same guy as serde.

I'd also like to use as little dependencies as possible but it's not easy....

PR #65, opened 2021-02-18
feature: Adds a `maybe_delete` function 

lfrancke:
that checks whether a `deletionn_timestamp` has been set before attempting to delete.

Trying to delete an already deleted/non-existant resource leads to an API error.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,4 +1,6 @@
 use crate::error::OperatorResult;
+use crate::finalizer;
+use crate::podutils;
 
 use either::Either;
 use k8s_openapi::Resource;
@@ -7,6 +9,7 @@ use kube::client::{Client as KubeClient, Status};
 use kube::Api;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
+use tracing::trace;
 
 /// This `Client` can be used to access Kubernetes.
 /// It wraps an underlying [kube::client::Client] and provides some common functionality.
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -137,6 +140,28 @@ impl Client {
             .await?)
     }
 
+    /// This potentially deletes a resource _if it is not already deleted_.
+    /// It checks by looking at the `deletion_timestamp` of the resource.
+    /// If that is the case it'll return a `Ok(None)`.
+    pub async fn maybe_delete<T>(&self, resource: &T) -> OperatorResult<Option<Either<T, Status>>>
```
maltesander, line 146: I don't like "maybe": The outcome is a deleted resource.
What about delete_dangling or delete_safely (in terms of we don't delete again if it is already up for deletion?)

maltesander (changes requested):

src/client.rs
lfrancke: Yeah, that's a valid point.... I'll think about it. Thanks!

lfrancke: @soenkeliebau pinging to see if you have any ideas for a better name

soenkeliebau: no definitive ideas ... something like "ensure_deleted"  / "delete_if_no_deletion_timestamp" ...

That being said, is there a specific reason to make this a separate function, when would we use `delete` over this one? To my mind the behavior in this function should be the default behavior in delete. Could we maybe change the return value of delete to include the deletion timestamp and a flag if it was already present or something similar?

maltesander (approved):

PR #68, opened 2021-02-23
Removes the `derive` feature from kube

lfrancke:
This is not needed here and can lead to problems because it pulls in kube-derive with default features which doesn't allow downstream users to easily opt-out of the schema feature.


lfrancke:
This contains some new dependencies. That was unintentional, I'll push a fix.

maltesander (approved):

PR #76, opened 2021-03-03
Helper method to convert LabelSelector to query string

lfrancke:
Assigning both @soenkeliebau and @maltesander .

Malte: I don't like the tests....there must be an easier way with rstest or something similar, no?
Any ideas?


soenkeliebau (approved):
LGTM

PR #78, opened 2021-03-03
Many new methods to deal with common Operator operations

lfrancke:
Many new methods to deal with common things we found while implementing Operators.

This is almost certainly not perfect and i assume some signatures should change etc. but before it gets too big I'd like to get it in now so it can be improved upon later.

### DIFF
--- a/src/podutils.rs
+++ b/src/podutils.rs
@@ -141,7 +141,6 @@ pub fn is_pod_assigned_to_node(pod: &Pod, node: &Node) -> bool {
 ///
 /// assert!(!podutils::pod_matches_multiple_label_values(&pod, &required_labels));
 /// ```
-// TODO: Tests
 pub fn pod_matches_multiple_label_values(
     pod: &Pod,
     required_labels: &BTreeMap<String, Option<Vec<String>>>,
@@ -273,11 +272,11 @@ pub fn pod_matches_labels(pod: &Pod, expected_labels: &BTreeMap<String, Option<S
 mod tests {
     use super::*;
     use crate::test;
-    use crate::test::PodBuilder;
+    use crate::test::{NodeBuilder, PodBuilder};
     use k8s_openapi::api::core::v1::{Pod, PodCondition, PodStatus};
 
     #[test]
-    fn test_is_pod_assigned_to_node() {
+    fn test_is_pod_assigned_to_node_name() {
         // Pod with no node_name
         let pod = PodBuilder::new().build();
         assert!(!is_pod_assigned_to_node_name(&pod, "foobar"));
@@ -290,6 +289,24 @@ mod tests {
         // Pod with empty spec doesn't match
         pod.spec = None;
         assert!(!is_pod_assigned_to_node_name(&pod, "foobar"));
+    }
+
+    #[test]
+    fn test_is_pod_assigned_to_node() {
+        // Pod with no node_name
+        let pod = PodBuilder::new().build();
+        let node = NodeBuilder::new().name("foobar").build();
+        let node2 = NodeBuilder::new().name("barfoo").build();
+        assert!(!is_pod_assigned_to_node(&pod, &node));
+
+        // Pod with node_name, matches one but not the other
+        let mut pod = PodBuilder::new().node_name("foobar").build();
+        assert!(is_pod_assigned_to_node(&pod, &node));
+        assert!(!is_pod_assigned_to_node(&pod, &node2));
+
+        // Pod with empty spec doesn't match
+        pod.spec = None;
+        assert!(!is_pod_assigned_to_node(&pod, &node));
     }
 
     #[test]
@@ -553,5 +570,38 @@ mod tests {
             1,
             find_pods_that_are_in_use(&nodes, &existing_pods, &expected_labels).len()
         );
+    }
+
+    #[test]
+    fn test_pod_matches_multiple_label_values() {
+        let pod = PodBuilder::new().build();
+
+        let mut required_labels = BTreeMap::new();
+        required_labels.insert("foo".to_string(), None);
+
+        // Pod has no labels but we want one but don't care about the value
+        assert!(!pod_matches_multiple_label_values(&pod, &required_labels));
+
+        // Pod has only the required label
+        let pod = PodBuilder::new().with_label("foo", "bar").build();
+        assert!(pod_matches_multiple_label_values(&pod, &required_labels));
+
+        // Pod has multiple labels
+        let pod = PodBuilder::new()
+            .with_label("foo", "bar")
+            .with_label("bar", "foo")
+            .build();
+        assert!(pod_matches_multiple_label_values(&pod, &required_labels));
+
+        // Pod has correct label but wrong value
+        required_labels.insert("bar".to_string(), Some(vec!["baz".to_string()]));
+        assert!(!pod_matches_multiple_label_values(&pod, &required_labels));
+
+        // Pod cas correct label and also one of the correct values
+        required_labels.insert(
+            "bar".to_string(),
+            Some(vec!["baz".to_string(), "foo".to_string()]),
+        );
+        assert!(pod_matches_multiple_label_values(&pod, &required_labels));
     }
 }
--- a/src/podutils.rs
+++ b/src/podutils.rs
@@ -187,6 +187,8 @@ pub fn pod_matches_multiple_label_values(
     pod: &Pod,
     required_labels: &BTreeMap<String, Option<Vec<String>>>,
 ) -> bool {
+    // TODO: This method currently will abort on the first error, we could extend this
+    //  (or add a second function) to return all "validation" results instead.
     let pod_labels = &pod.metadata.labels;
 
     for (expected_key, expected_value) in required_labels {
--- a/src/podutils.rs
+++ b/src/podutils.rs
@@ -203,9 +203,76 @@ pub fn is_valid_pod(pod: &Pod, required_labels: &BTreeMap<String, Option<Vec<Str
     ) && pod_matches_multiple_label_values(pod, required_labels)
 }
 
+/// This method can be used to find Pods that are not needed anymore.
+///
+/// For this to work we'll compare a list of all Pods against a list of Pods that are actively being used.
+/// We'll do this for an arbitrary number of Node lists and match labels.
+// TODO: Test and docs
+pub fn find_excess_pods<'a>(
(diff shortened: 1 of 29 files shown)


lfrancke:
I think I forgot to remove the `finalizer_name` method which should not be needed anymore now.
This is a TODO for me.

src/podutils.rs
```diff
@@ -93,46 +96,229 @@ where
...
+/// #            spec: None,
+/// #            status: None,
+/// #        };
+///
+/// let mut required_labels = BTreeMap::new();
+/// required_labels.insert("foo".to_string(), Some("bar".to_string()));
+///
+/// assert!(!podutils::pod_matches_labels(&pod, &required_labels));
+/// ```
+pub fn pod_matches_labels(pod: &Pod, required_labels: &BTreeMap<String, Option<String>>) -> bool {
```
soenkeliebau, line 142: I've commented somewhere else if we could introduce a type alias for `BTreeMap<String, Option<String>>` - after seeing this, we could think that a bit further and make that `BTreeMap<String, Option<Vec<String>>>` instead and implement `From<BTreeMap<String, Option<String>>` for that, which would allow us to remove this function and have `pod_matches_multiple_label_values` take something implementing Into<LabelList> or whatever we decide to call it.

Just a thought, and doesn't need to be addressed for this PR though.

src/podutils.rs
```diff
@@ -93,46 +96,229 @@ where
...
+/// let mut required_labels = BTreeMap::new();
+/// required_labels.insert("foo".to_string(), None);
+///
+/// assert!(!podutils::pod_matches_multiple_label_values(&pod, &required_labels));
+/// ```
+pub fn pod_matches_multiple_label_values(
+    pod: &Pod,
+    required_labels: &BTreeMap<String, Option<Vec<String>>>,
+) -> bool {
+    let pod_labels = &pod.metadata.labels;
```
soenkeliebau, line 190: I've thought about this function for a little while and can't shake the feeling that we could do 'better' with a two step .iter().filter(...) approach to first check for missing keys and then for matching values.  
That would also allow to give a full report of the non-matching labels in the debug message instead of just the first non-matching item.

That being said, this is not something we should spend time on currently!

soenkeliebau (changes requested):
Looks good overall, I've requested changes mostly because of the usage of RoleGroup in reconcile.rs

src/reconcile.rs
```diff
@@ -75,6 +89,29 @@ impl<T> ReconciliationContext<T> {
...
+    pub async fn wait_for_running_and_ready_pods(&self, pods: &[Pod]) -> ReconcileResult<Error> {
+        for pod in pods {
+            if !podutils::is_pod_running_and_ready(pod) {
+                return Ok(ReconcileFunctionAction::Requeue(self.requeue_timeout));
+            }
+        }
+        Ok(ReconcileFunctionAction::Continue)
+    }
+
+    // TODO: Docs & Test
```
lfrancke, line 102: I forgot this, maybe I forgot this whole file?

src/role_utils.rs
```diff
@@ -0,0 +1,127 @@
...
+//! Nested under a role there can be multiple role groups, each with its own LabelSelector and configuration.
+//!
+//! ## Example
+//!
+//! Role Groups: `default`, `20-cores`, `gpu`
+//!
+//! # Pod labels
+//!
+//! Each Pod that Operators create needs to have a common set of labels:
+//! * <TODO>/name - The name of the parent resource, this is useful so an operator can list all its pods by using a LabelSelector
```
lfrancke, line 40: Opinions on how these labels should look like? @soenkeliebau

src/role_utils.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use crate::role_utils::find_excess_pods;
+    use crate::test::{NodeBuilder, PodBuilder};
+    use std::collections::BTreeMap;
+
+    #[test]
+    fn test_find_excess_pods() {
```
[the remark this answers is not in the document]
lfrancke: I'm open for suggestions but I actually like it at the moment.
It just defines the stuff we actually need.

src/podutils.rs
lfrancke: Do you have a suggestion for a type alias?
I thought about it as well/had one in for a while but nothing really descriptive came to mind.

src/role_utils.rs
soenkeliebau: Not really, for the Kafka operator my current idea is:

```
kafka.stackable.tech/service
kafka.stackable.tech/role-group
kafka.stackable.tech/role
```

But nothing here is anything that I am passionate about..

src/podutils.rs
soenkeliebau: I've spent 10 minutes with my idea, which of course wasn't all that easy, because you cannot implement traits for an alias, so it had to be a new type..

I ended up with this:

```rust
struct LabelAndValueList {
    list: BTreeMap<String, Option<Vec<String>>>,
}

impl From<&BTreeMap<String, Option<Vec<String>>>> for LabelAndValueList {
    fn from(value: &BTreeMap<String, Option<Vec<String>>>) -> Self {
        LabelAndValueList {
            list: value.clone(),
        }
    }
}

impl From<&BTreeMap<String, Option<String>>> for LabelAndValueList {
    fn from(value: &BTreeMap<String, Option<String>>) -> Self {
        let result = value
            .iter()
            .map(|(key, value)| {
                (
                    key.clone(),
                    value.as_ref().map(|string| vec![string.clone()]),
                )
            })
            .collect::<BTreeMap<_, _>>();
        LabelAndValueList { list: result }
    }
}
```

Which would allow having only one function that can take both types of BTreeMaps: 

```rust
pub fn pod_matches_multiple_label_values<T>(pod: &Pod, required_labels: T) -> bool
where
    T: Into<LabelAndValueList> + Sized,
{
    let required_labels = required_labels.into();
    // ...
}
```

But probably has other issues. I only tested this until I got one test running :)

src/reconcile.rs
```diff
@@ -133,6 +178,129 @@
...
 
+    /// Finds nodes in the cluster that match a given LabelSelector
+    /// This takes list of RoleGroup and returns
+    /// a map with found nodes per String
+    ///
+    /// This will only match Stackable nodes (Nodes with a special label).
+    /// TODO: Docs & Tests
+    pub async fn find_nodes_that_fit_selectors(
+        &self,
+        role_groups: Vec<RoleGroup>,
```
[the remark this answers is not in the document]
lfrancke: You're correct. That shouldn't belong here but....I think I'm in favor of leaving it in anyway. At the moment we're the only ones using this library and I can't think of a really good way of abstracting this while still keeping it simple to use.

src/reconcile.rs
```diff
@@ -271,9 +417,68 @@ fn pod_owned_by(pod: &Pod, owner_uid: &str) -> bool {
...
+/// WARN: Should a label "type" already be used this will be overridden!
+/// If this is really needed add a match…expression
+///
+/// We will not however change the original LabelSelector, a new one will be returned.
+fn add_stackable_selector(selector: &LabelSelector) -> LabelSelector {
+    let mut selector = selector.clone();
+    selector
+        .match_labels
+        .get_or_insert_with(BTreeMap::new)
+        .insert("type".to_string(), "krustlet".to_string());
```
lfrancke: @soenkeliebau Is this the correct label to target? Can you tell me (again) which would be best?

src/podutils.rs
lfrancke: Interesting. I have to think about it.


I remember  a problem that one list should not be used in place of the other because of different semantic meaning.

I have to admit that I forgot parts of it but one example:
Let's say you have a bunch of role-groups A, B and C and now you find a group "D" in your labels.
This is probably not illegal. More likely is that "D" did exist in your spec and the user removed it.

I'm not sure if this is true here. These things are not easy :)

src/reconcile.rs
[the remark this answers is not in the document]
lfrancke: I've moved it to role_utils but that means we need to pass in a client etc.

I thought about using an Extension Trait but ReconciliationContext is not a trait...another TODO maybe?

src/podutils.rs
lfrancke: After discussing this offline we decided to not do this because it might mean that someone passes in multiple label values where only one is allowed.
By having two separate methods the distinction becomes clearer.

src/podutils.rs
soenkeliebau: I'm not sure what you mean by "validate method", but still think we shouldn't spend time on this at the moment :)

src/reconcile.rs
soenkeliebau: https://github.com/deislabs/krustlet/blob/2c01e500b9cd906837b114ca00b401fefb21f363/crates/kubelet/src/node/mod.rs#L445

lfrancke: Yeah but won't that hit any Krustlet? (On the off-chance that someone has another krustlet with WASM running for example).

Shouldn't we use something stackable specific?

soenkeliebau: That is true, but would require a change in the agent at the same time. I think I'd prefer to merge this as is and make that change at a later time (doesn't have to be much later).

soenkeliebau: Come to think of it, we could have label creation shared between the agent and this to ensure it remains the consistent. I am hesitant to make the agent depend on this framework solely for that reason though.. then again, won't hurt much after the compiler did it's job optimizing stuff out I think ..

lfrancke: So we don't have any identifying label that we can use? Don't we have one with "stackable-linux"?

soenkeliebau (approved):
LGTM

src/reconcile.rs
soenkeliebau: This is what a node currently has:

```
Labels:             beta.kubernetes.io/arch=stackable-linux
                    beta.kubernetes.io/os=stackable-linux
                    kubernetes.io/arch=stackable-linux
                    kubernetes.io/hostname=soenke-laptop
                    kubernetes.io/os=stackable-linux
                    type=krustlet
Annotations:        node.alpha.kubernetes.io/ttl: 0
                    volumes.kubernetes.io/controller-managed-attach-detach: true
Taints:             kubernetes.io/arch=stackable-linux:NoExecute
                    kubernetes.io/arch=stackable-linux:NoSchedule
                    node.kubernetes.io/unreachable:NoSchedule
```

PR #79, opened 2021-03-05
The 'ensure_crd_created' method retries on failure.

Pscheidl:
Attempt to implement CRD creation retry. Added an optional timeout for that. One frequent reason why CRDs fail to be created are permissions - the retry won't help here, it will simply time out, unless permissions are adjusted in time.

## Tests

There is a basic `test_ensure_crd_created` introduced. It is **not** ignored by default and it requires Kubernetes and `KUBECONFIG` environment variable set. Could be ignored by default to make tests runnable locally by anyone. Running Kubernetes cluster in GitHub actions is simple - there is a ready `debianmaster/actions-k3s` available. Exposes the `KUBECONFIG` var automatically, with `clusteradmin` permissions enabled by default (all permissions basically).

If left enabled, anyone who'll do `cargo test` will see tests failed, unles the above mentioned condition is met. This is easy to do, e.g. `https://k3s.io/` is as simple as this on linux:

```bash
curl -sfL https://get.k3s.io | sh -
export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
```
(sometimes, chown or simply `sudo chmod 777 /etc/rancher/k3s/k3s.yaml` on the kubeconfig file is required).

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -26,7 +26,7 @@ pub struct Client {
     post_params: PostParams,
     delete_params: DeleteParams,
     /// Default namespace as defined in the kubeconfig this client has been created from.
-    default_namespace: String,
+    pub default_namespace: String,
 }
 
 impl Client {
@@ -321,14 +321,39 @@ impl Client {
         Api::namespaced(self.client.clone(), namespace)
     }
 
-    /// Waits indefinitely until given resource with metadata is created in Kubernetes. If the resource
-    /// is already present, this method just returns. Makes no assumptions about resource's state,
+    /// Waits indefinitely until resources matching given `ListParams` are created in Kubernetes.
+    /// If the resource is already present, this method just returns. Makes no assumptions about resource's state,
     /// e.g. a pod created could be created, but not in a ready state.
     ///
     /// # Arguments
     ///
     /// - `namespace` - Optional namespace to look for the resources in.
     /// - `lp` - Parameters to filter resources to wait for in given namespace.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use kube::api::ListParams;
+    /// use std::time::Duration;
+    /// use tokio::time::error::Elapsed;
+    /// use k8s_openapi::api::core::v1::Pod;
+    /// use stackable_operator::client::{Client, create_client};
+    ///
+    /// #[tokio::main]
+    /// async fn main(){
+    /// let client: Client = create_client(None).await.expect("Unable to construct client.");
+    /// let lp: ListParams =
+    ///         ListParams::default().fields(&format!("metadata.name=nonexistent-pod"));
+    ///
+    /// // Will time out in 1 second unless the nonexistent-pod actually exists
+    ///  let wait_created_result: Result<(), Elapsed> = tokio::time::timeout(
+    ///          Duration::from_secs(1),
+    ///          client.wait_created::<Pod>(Some(client.default_namespace.clone()), lp.clone()),
+    ///      )
+    ///      .await;
+    /// }
+    /// ```
+    ///
     pub async fn wait_created<T>(&self, namespace: Option<String>, lp: ListParams)
     where
         T: Meta + Clone + DeserializeOwned + Send + 'static,
--- a/src/client.rs
+++ b/src/client.rs
@@ -4,8 +4,8 @@ use crate::label_selector;
 use crate::podutils;
 
 use either::Either;
-use k8s_openapi::apimachinery::pkg::apis::meta::v1::{Condition, LabelSelector};
 use futures::StreamExt;
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::{Condition, LabelSelector};
 use k8s_openapi::Resource;
 use kube::api::{DeleteParams, ListParams, Meta, Patch, PatchParams, PostParams};
 use kube::client::{Client as KubeClient, Status};
--- a/src/client.rs
+++ b/src/client.rs
@@ -9,10 +9,11 @@ use futures::StreamExt;
 use k8s_openapi::Resource;
 use kube::api::{DeleteParams, ListParams, Meta, Patch, PatchParams, PostParams};
 use kube::client::{Client as KubeClient, Status};
-use kube::Api;
+use kube::{Api, Config};
 use kube_runtime::watcher::Event;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
+use std::convert::TryFrom;
 use tracing::debug;
 use tracing::trace;
 
@@ -24,10 +25,16 @@ pub struct Client {
     patch_params: PatchParams,
     post_params: PostParams,
     delete_params: DeleteParams,
+    /// Default namespace as defined in the kubeconfig this client has been created from.
+    default_namespace: String,
 }
 
 impl Client {
-    pub fn new(client: KubeClient, field_manager: Option<String>) -> Self {
+    pub fn new(
+        client: KubeClient,
+        field_manager: Option<String>,
+        default_namespace: String,
+    ) -> Self {
         Client {
             client,
             post_params: PostParams {
@@ -42,6 +49,7 @@ impl Client {
                 ..PatchParams::default()
             },
             delete_params: DeleteParams::default(),
+            default_namespace,
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/client.rs
```diff
@@ -250,6 +253,36 @@ impl Client {
     {
         Api::namespaced(self.client.clone(), namespace)
     }
+
+    /// Waits until given resource with metadata is recognized as applied by Kubernetes API
```
lfrancke, line 257: Only when I saw the usage in crd.rs did I understand how this is to be used.
When you say `given resource` the way it is _given_  is via a `field_selector` in the `ListParams`.

An example would be great here or at least a quick sentence about that.

Your method is definitely the most flexible solution so I'm happy to merge as is.
For convenience we could add a method that wraps this because I assume watching by name will be the most common scenario. But we can do that once we needed.

src/client.rs
```diff
@@ -250,6 +253,36 @@ impl Client {
...
+    where
+        T: Meta + Clone + DeserializeOwned + Send + 'static,
+    {
+        let api: Api<T> = self.get_api(namespace);
+        let mut watcher = kube_runtime::watcher(api, lp).boxed();
+
+        while let Some(result) = watcher.next().await {
+            match result {
+                Ok(event) => match event {
+                    Event::Applied(_) | Event::Restarted(_) => {
```
lfrancke, line 273: `Restarted` means the watch stream was restarted, no?
That in itself doesn't carry much information about whether the object is ready or now.
I'd have assumed that we only need to wait for `Applied`?

src/client.rs
```diff
@@ -250,6 +253,36 @@ impl Client {
...
+                        break;
+                    }
+                    Event::Deleted(_) => {
```
lfrancke, line 276: Can this end in an endless wait/loop?
If I create a resource, immediately delete it and then call `wait_ready` (for whatever reason) could it be that I only receive the `Deleted` Event?
Thinking out loud: I guess not because the first thing watcher does is to list all current resources which'd mean we'd see it before it's deleted.

src/crd.rs
```diff
@@ -74,8 +82,22 @@ where
...
+                if let Ok(res) = create::<T>(client.clone()).await {
+                    break res;
+                }
+                tokio::time::sleep(Duration::from_millis(100)).await;
+            }
+            wait_ready::<T>(client.clone()).await?;
+            Ok(())
+        };
+        if let Some(timeout) = timeout {
+            tokio::time::timeout(timeout, crd_created).await?
```
lfrancke, line 97: TIL, thank you! That's great.

lfrancke (commented):
Thank you for the Pull Request, this is great functionality to have.

I just have a few minor comments.
I love the K3s integration. It's been on our list for a while and we didn't get around to it so far

src/client.rs
Pscheidl: Yes it can, unles there's a timeout set up by the caller on this whole `wait_ready` function. Not sure if this is a good pattern, but using e.g. tokio, you can always timeout futures yourself, no need for this function to do it for you. I use it 
on this line: https://github.com/stackabletech/operator-rs/pull/79/files#diff-31b064fae044a3049a84f1d3379839a417abdbf581ef11e516aa2a94eeb34a40R97

src/client.rs
teozkr: We don't send an individual `Applied` event if it is already covered by the `Restarted`. So the correct pattern here would be something like `Event::Applied(_) | Event::Restarted([_, ..])`.

Pscheidl: Looks like `Event::Restarted` carries a Vec. `Event::Restarted([_, ..]` results in the following error:

```
error[E0529]: expected an array or slice, found `std::vec::Vec<T>`
   --> src/client.rs:273:58
    |
273 |                     Event::Applied(_) | Event::Restarted([_, ..]) => {
    |                                                          ^^^^^^^ pattern cannot match with input type `std::vec::Vec<T>`
```

Pscheidl: I wonder `restarted` can give me an empty Vec provided the `ListParams` matches nothing. I'll write a test for that.

Pscheidl:
@lfrancke The K3S integration is something that worked for me very well in the past. However, it has to be said (and I will document this) that the default KUBECONFIG produced by K3S has all the permissions enabled. So what works by default on K3S may fail due to permissions on other clusters for common users.

src/client.rs
Pscheidl: The necessity of catching the `restarted` event is nicely demonstrated by this test: https://github.com/stackabletech/operator-rs/pull/79/files#diff-7f93c4e263c4e9ec748f804c7fd04a3b2fde86ffd741fb5516d67e1097bae4c1R366

if the resource is already created and there are no changes, it might come to us as `restarted.

Pscheidl: The answer is yet, restarted can give empty Vec if ListParams filters nothing. Fixed in the code and added test for that.

src/client.rs
Pscheidl: This is great observation, thank you. I reworded the whole documentation and added a usage example that shows full usage, including timeout.

Pscheidl: Regarding the convenience method - I'll happily add it if required. I usually add stuff once needed, so I'd just do it like that personally.

src/crd.rs
```diff
@@ -74,8 +82,22 @@ where
...
+            loop {
+                if let Ok(res) = create::<T>(client.clone()).await {
+                    break res;
+                }
+                tokio::time::sleep(Duration::from_millis(100)).await;
+            }
+            wait_ready::<T>(client.clone()).await?;
+            Ok(())
+        };
+        if let Some(timeout) = timeout {
```
lfrancke, line 96: Would it make sense to extract this to a `wait_created_with_timeout` method next to the `wait_created` one? This seems like a super useful pattern to me and it's not limited to CRDs

src/crd.rs
```diff
@@ -90,3 +112,18 @@ where
     let crd: CustomResourceDefinition = serde_yaml::from_str(T::CRD_DEFINITION)?;
     client.create(&crd).await.and(Ok(()))
 }
+
+/// Waits until CRD of given type `T` is applied to Kubernetes.
+pub async fn wait_ready<T>(client: Client) -> OperatorResult<()>
```
lfrancke, line 117: This sa also not tied to Crds really....but let's leave it here for now.

The whole Crd trait is not needed anymore I believe since the schemars integration.

lfrancke (approved):
This looks good to me.
I've left one or two comments but they are optional.
I'd be happy to merge as is but one thing would be great: If you could copy your comments from the PR about how to run the tests locally to the README?

If you don't want to that's fine and I'll create a follow-up PR myself.

I'll leave the decision up to you. Tell me to and I'll merge as is.

Thank you for the contribution!
I'd just need you to accept the CLA if possible.

Pscheidl:
@lfrancke I'll definitely do the guides, please don't merge yet.

Pscheidl:
Made sure the ignored tests actually run by looking into the workflow logs.

src/crd.rs
Pscheidl: First, the naming is wrong and it should be aligned with `wait_created` on Client. So I renamed it.

`wait_created` and its documentation clearly says the resources is created in Kubernetes and in some context (CRDs) recognized by Kubernetes APIs. Nothing more. There are no implications about resource's state. If that was a pod, it could be created, yet the underlying container could be crashed. I **avoided** the word `ready`, because readiness has different meaning in Kubernetes, not to say the meaning changes with context. E.g. a ready pod is generally understood to be a pod that's has `status.Ready=true`.

Pscheidl:
The name `wait_ready` of one of the functions is wrong and it should be aligned with `wait_created` on Client. So I renamed it.

`wait_created` and its documentation clearly says the resources is created in Kubernetes and in some context (CRDs) recognized by Kubernetes APIs. Nothing more. There are no implications about resource's state. If that was a pod, it could be created, yet the underlying container could be crashed. I **avoided** the word `ready`, because readiness has different meaning in Kubernetes, not to say the meaning changes with context. E.g. a ready pod is generally understood to be a pod that's has `status.Ready=true`.

src/crd.rs
Pscheidl: This could be one generic method. But hey, it's already there. The `Option<Duration>` passed as argument into this method is only used in `tokio::time::timeout`. What if we just let the callers handle the timeout themselves ? 

Instead of calling the function with Duration,

```rust
    ensure_crd_created::<TestCrd>(client.clone(), Some(Duration::from_secs(10)))
        .await
```

the function signature is simplified and the timeout is delegated to that dedicated function, which to me is `tokio::time::timeout` and we'd just duplicate it's functionality. It looks like this.

```rust
    tokio::time::timeout(
        Duration::from_secs(30),
        ensure_crd_created::<TestCrd>(client.clone()),
    )
    .await
```
Also, the code is cleaner and it clearly states there is a timeout.

Pscheidl: I did the above mentioned change for you to see the effects. It is contained in a completely separate commit https://github.com/stackabletech/operator-rs/pull/79/commits/3b2e5b078332a11df2adf67152fcae0803310442 - we can always roll this one back.

src/crd.rs
Pscheidl: Ok, I'll leave that one as is now, as I don't have much context. If there are bigger changes coming, let's not waste time on intermediate changes.

src/crd.rs
lfrancke: Makes sense to me! We can always add something later.

lfrancke (approved):
I'm fine with merging this.

Thank you for your contribution!
Is there anything else you want changed? If not I'll merge it later today.

Pscheidl:
@lfrancke Sorry for the late reply, no, definitely nothing I want changed. Thanks for the review and the help. Let's merge.

PR #81, opened 2021-03-08
feat: Move the finalization handling out of the Controller

lfrancke:
BREAKING CHANGE: This means that adding and removing finalizers is now optional

### DIFF
--- a/src/finalizer.rs
+++ b/src/finalizer.rs
@@ -22,6 +22,7 @@ where
 /// and will update the resource in Kubernetes.
 ///
 /// It'll return `true` if we changed the object in Kubernetes and `false` if no modification was needed.
+/// If the object is currently being deleted this _will_ return an Error!
 pub async fn add_finalizer<T>(
     client: &Client,
     resource: &T,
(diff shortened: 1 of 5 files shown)


src/reconcile.rs
```diff
@@ -173,6 +241,26 @@ where
         );
         self.set_condition(condition).await
     }
+
+    /// A reconciler function to adsour finalizer to the list of finalizers.
```
maltesander, line 245: Typo:
 /// A reconciler function to add our finalizer to the list of finalizers.

src/finalizer.rs
```diff
@@ -17,25 +18,45 @@ where
...
+    Ok(true)
 }
 
 /// Removes our finalizer from a resource object.
 ///
 /// # Arguments
-/// `name` - is the name of the resource we want to patch
-/// `namespace` is the namespace of where the resource to patch lives
-pub async fn remove_finalizer<T>(client: Client, resource: &T, finalizer: &str) -> OperatorResult<T>
+///
```
maltesander, line 52: Missing client argument comment. I know its self explanatory but you commented it above. Just to be consistent.
/// - `client` - The Client to access Kubernetes

maltesander (changes requested):
Only a typo and missing comment argument (when you provide arguments). Did not mention missing argument comments where you did not specify them at all.

maltesander (approved):

PR #87, opened 2021-03-24
Generic command controller for custom resource command objects; kube version bump to 0.51

maltesander:


src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+use std::pin::Pin;
+use std::time::Duration;
+
+const FINALIZER_NAME: &str = "command.stackable.tech/cleanup";
+
+/// The Owner trait represents our main controller object. This is required to be able to search
+/// for the main controller object and extract metadata information to set the owner reference
+/// in our command object. This must be implemented by every command custom resource.
+pub trait CommandOwner {
+    type Owner: Meta + Clone + DeserializeOwned + Debug + Send + Sync + Crd;
```
lfrancke, line 116: We already talked about this offline: I believe we don't need this type in here.
It can be moved out and be another generic parameter for the methods below.

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    async fn init_reconcile_state(
+        &self,
+        context: ReconciliationContext<Self::Item>,
+    ) -> Result<Self::State, Error> {
+        Ok(CommandState { context })
+    }
+}
+
+/// This creates an instance of a [`Controller`] which waits for incoming events and reconciles them.
```
lfrancke, line 203: This comment can be more specific. It's the main entry point into this module.
Maybe at least a hint about the module level documentation.

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+///
+/// This is an async method and the returned future needs to be consumed to make progress.
+pub async fn create_command_controller<T>(client: Client)
+where
+    T: CommandOwner + Meta + Clone + Debug + DeserializeOwned + Send + Sync + 'static,
+{
+    let command_api: Api<T> = client.get_all_api();
+    let owner_api: Api<T::Owner> = client.get_all_api();
```
lfrancke, line 211: Help me please...why do we even need to watch the Owner?

As mentioned above this should probably just take a second generic parameter instead of using T::Owner

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+#[async_trait]
+impl<T> ControllerStrategy for CommandStrategy<T>
+where
+    T: CommandOwner + Meta + Clone + DeserializeOwned + Send + Sync,
+{
+    type Item = T;
+    type State = CommandState<T>;
+    type Error = Error;
+
+    fn finalizer_name(&self) -> String {
```
lfrancke, line 191: This is a TODO for me. In the #78 PR this should be removed!

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+
+struct CommandState<T> {
+    context: ReconciliationContext<T>,
+}
+
+impl<T> CommandState<T>
+where
+    T: CommandOwner + Clone + DeserializeOwned + Meta,
+{
+    /// This controller sets the owner reference in our custom resource object. We need to
```
lfrancke, line 129:
```suggestion
    /// This controller sets the owner reference in our custom resource. We need to
```

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+/// # Arguments
+/// * `client` - Kubernetes client
+/// * `metadata_name` - The metadata.name value to be matched
+///
+async fn find_owner<T>(client: &Client, metadata_name: &str) -> OperatorResult<T>
+where
+    T: Clone + DeserializeOwned + Meta,
+{
+    let lp = ListParams::default().fields(&format!("metadata.name={}", metadata_name));
+    let mut owners: Vec<T> = client.list(None, &lp).await?;
```
lfrancke, line 233: This should use the namespace from the command object

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    if owners.is_empty() {
```
lfrancke, line 235: Good question on what to do with commands like this.
We could just delete them but that'd be hard to debug....
At least flagging them for deletion would probably be good.
Nothing to do now but a TODO might be good

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+        return Err(Error::MissingCustomResource {
+            name: metadata_name.to_string(),
+        });
+    }
+
+    Ok(owners.pop().unwrap())
```
lfrancke, line 241: You could change this (including the is_empty) to a match on pop() (but I'm fine with the way it is now)

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+where
+    T: CommandOwner + Clone + DeserializeOwned + Meta + Send + Sync,
+{
+    type Error = Error;
+
+    fn reconcile(
+        &mut self,
+    ) -> Pin<Box<dyn Future<Output = Result<ReconcileFunctionAction, Self::Error>> + Send + '_>>
+    {
+        Box::pin(async move { self.set_owner_reference().await })
```
lfrancke, line 164: I suggest adding a second method here that first checks if an ownerReference is already set

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+///
+async fn patch_owner_reference<T>(
+    client: &Client,
+    resource: &T,
+    owner_reference: &OwnerReference,
+) -> OperatorResult<T>
+where
+    T: Clone + DeserializeOwned + Meta,
+{
+    // TODO: Check for existing owner references. As of now we just override.
```
lfrancke, line 259: Please add that we can use the json_path method for that purpose

src/command_controller.rs
```diff
@@ -0,0 +1,288 @@
...
+        Box::pin(async move { self.set_owner_reference().await })
+    }
+}
+
+#[derive(Debug)]
+struct CommandStrategy<T> {
+    // TODO: Better workaround for PhantomData?
+    // We use it here cause we need to make CommandStrategy generic to be able to do:
+    // impl<T> ControllerStrategy for CommandStrategy<T>
+    _ignore: Option<std::marker::PhantomData<T>>,
```
lfrancke, line 173: This makes me wonder if my architecture with the strategy etc. is not a good one....
I see why you need it and I don't see a good way to remove it.

lfrancke (changes requested):
Not 100% done reviewing but have to leave for now.

src/command_controller.rs
lfrancke: I've updated my hackathon PR so that this method is not needed anymore

src/command_controller.rs
```diff
@@ -196,23 +263,29 @@ where
...
+            context,
+            owner: None,
+            owner_reference: None,
+        })
     }
 }
 
-/// This creates an instance of a [`Controller`] which waits for incoming events and reconciles them.
+/// This creates an instance of a [`Controller`] which waits for incoming commands.
+/// For each command, we set the owner reference of our main controller custom resource, to
```
lfrancke, line 275: I know what you want to say here but I don't understand this sentence.

Something along these lines maybe?
```suggestion
/// For each command, we try to find the referenced resource and will set the Owner Reference of the command to this referenced resource.
/// If we can't find the referenced object we TODO <- Malte, what do we do then?.
/// This means that the controller of the parent resource can now watch for commands and this helper controller will make sure that they trigger a reconcile for the parent by setting the OwnerReference.
```

src/command_controller.rs
```diff
@@ -248,34 +326,45 @@ where
 /// * `resource` - The resource where to set the OwnerReference
 /// * `owner_reference` - The OwnerReference to add
 ///
-async fn patch_owner_reference<T>(
+async fn patch_owner_reference<C>(
```
lfrancke, line 329: I'm in a rush but I think this can be much simpler by using json_patch.
If you like I can make a suggestion next week.

lfrancke (changes requested):

src/command_controller.rs
```diff
@@ -225,20 +298,25 @@ where
 /// * `client` - Kubernetes client
 /// * `metadata_name` - The metadata.name value to be matched
 ///
-async fn find_owner<T>(client: &Client, metadata_name: &str) -> OperatorResult<T>
+async fn find_owner<O>(
```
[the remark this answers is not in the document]
maltesander: Yeah i think it can be useful for multiple operators, so i would rather add it here.

src/command_controller.rs
```diff
@@ -1,102 +1,12 @@
 //! Generic controller to add command CRDs for restart, start, stop (...) command operations as
 //! specified in [Stackable ADR010](https://github.com/stackabletech/documentation/blob/main/adr/ADR010-command_pattern.adoc).
-//!
-//! # Example
```
lfrancke: Having an example would be useful, no?

lfrancke (commented):
Looks good to me now!
We just need to wait for the hackathon PR

lfrancke:
Instead of waiting for the hackathon PR (could be a bit longer) you could also rebase this on main. Do you depend on anything from the new stuff?

soenkeliebau:
I think the test failure is due to the `ignore` [here](https://github.com/stackabletech/operator-rs/blob/934cee51e4ab611507ba37b1f62f01077e8c525d/src/command_controller.rs#L6) 

afaik that code is ignored by default, but by explicitly marking it _ignore_ you bring it into scope when we run `cargo test -- --ignore` and that has stuff which the compiler complains about.

src/command_controller.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+impl<C, O> CommandState<C, O>
+where
+    C: Command + Clone + DeserializeOwned + Meta,
+    O: Clone + DeserializeOwned + Meta,
+{
+    /// Check if our custom resource command already has the owner reference set to the main
+    /// controller custom resource. If so we can stop the reconcile.
+    async fn owner_reference_existing(&mut self) -> ReconcileResult<Error> {
```
lfrancke, line 162: Can you use controller_ref::get_controller_of here?
Because yours doesn't check for controller = true

That'd get you the controller and then you could check kind & name on that one.

src/command_controller.rs
```diff
@@ -0,0 +1,329 @@
...
+///
+pub async fn list_commands<C>(client: &Client) -> OperatorResult<Vec<C>>
+where
+    C: Meta + Clone + DeserializeOwned,
+{
+    let command_api: Api<C> = client.get_all_api();
+    let list = command_api
+        .list(&ListParams {
+            ..ListParams::default()
+        })
```
lfrancke, line 323:
```suggestion
        .list(&ListParams::default())
```

lfrancke (changes requested):

lfrancke (approved):

PR #92, opened 2021-04-08
Rename podutils to pod_utils

lfrancke:
To be more consistent with the other modules

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ serde_yaml = "0.8"
 thiserror = "1"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 tracing = "0.1"
+tracing-futures = "0.2"
 tracing-subscriber = "0.2"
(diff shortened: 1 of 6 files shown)


soenkeliebau (dismissed):
LGTM overall and checks all passed as well

My only concern is, that it doesn't build for me locally due to 

```
error[E0433]: failed to resolve: use of undeclared crate or module `tracing_futures`
   --> src/controller.rs:310:1
    |
310 | / {
311 | |     debug!(?resource, "Beginning reconciliation");
312 | |     let context = context.get_ref();
313 | |
...   |
363 | |     }
364 | | }
    | |_^ not found in `tracing_futures`
    |
help: consider importing this trait
    |
114 | use tracing::Instrument;
    |

error: aborting due to previous error

```

But I expect that this is  due to an error on my part, as I can see that the Github actions didn't use the tracing-futures dependency either and it worked perfectly fine.

soenkeliebau (approved):
LGTM

PR #93, opened 2021-04-08
feature: Namespace support

lfrancke:
* Supports getting namespaces to watch for operators using an environment variable
* Supports validating namespace names

### DIFF
--- a/src/namespace.rs
+++ b/src/namespace.rs
@@ -1,5 +1,4 @@
 //! This module provides helpers and constants to deal with namespaces
-//!
 use crate::error::{Error, OperatorResult};
 use crate::validation::validate_namespace_name;
 use std::env;
--- a/src/namespace.rs
+++ b/src/namespace.rs
@@ -1,3 +1,7 @@
+//! This module provides helpers and constants to deal with namespaces
+//!
+use crate::error::{Error, OperatorResult};
+use crate::validation::validate_namespace_name;
 use std::env;
 use std::env::VarError;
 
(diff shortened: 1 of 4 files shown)


src/validation.rs
```diff
@@ -69,6 +78,24 @@ pub fn is_rfc_1123_subdomain(value: &str) -> Vec<String> {
     errors
 }
 
+/// Tests for a string that conforms to the definition of a label in DNS (RFC 1035).
+pub fn is_rfc_1035_label(value: &str) -> Vec<String> {
```
maltesander, line 82: (For subdomain as well,) I do not like the "is_xyz" not returning a bool. 
For the usage: If the return value is empty, it is a valid label. I find that a bit strange.
I see that you wanna collect and propagate the errors.

What about a tuple:
pub fn is_rfc_1035_label(value: &str) -> (bool, Option<Vec<String>>)
Result with custom error containing a vec?
pub fn is_rfc_1035_label(value: &str) -> Result<bool, RfcError)

maltesander (approved):
I have some troubles with the error vec return type. But since i know that you know what you are doing, it is approved. Just have a look at my comment and maybe clarify it for me.

src/validation.rs
siegfriedweber: What about `Result<(), Vec<String>>`?

The best would be to avoid boolean blindness and parse instead of validate the input, e.g.
```
struct Label(String);

type ParseErrors = Vec<String>;

pub fn parse_rfc_1035_label(value: &str) -> Result<Label, ParseErrors> {
    …
}
```

soenkeliebau: I'm mostly with Sigi here, Result<(),Vec<String>> would have been my first choice. 
Parsing seems a bit overkill to me, as we basically return a String if I am not mistaken? Defining an extra struct for that seems like overkill to me.

lfrancke: I agree that this isn't a great API.
I copied it almost verbatim from Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L187

I agree that it'd be great - in theory - to have all usages automatically validated by using those special types but I'm not sure how realistic that is considering that our input and output is always "untyped".....but I like it.

I think for now the best would be to either keep it as it is for maximum "compatibility" with K8S or use the `Result<()...>` thing.

soenkeliebau: Do we gain anything by sticking to the way Kubernetes does it?
Unless there is a clear benefit I'd be in favor of doing this the rusty way and using a Result

soenkeliebau:
A general thought that occurred to me this morning, would it make sense to try and commit these validations upstream in kube-rs? I'd assume those would be useful to others as well..

src/namespace.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn test_parse_watch_namespaces() {
+        assert!(matches!(get_watch_namespace(), Ok(WatchNamespace::All)));
```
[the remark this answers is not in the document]
lfrancke: The unique name won't work here because it's hardcoded in "get_watch_namespace".
Could be pulled out into yet another method in theory if if becomes a problem.

lfrancke:
I thought the same, yes. I think it makes sense to talk with the kube-rs folks to see if they are interested in any of the stuff we're doing.

src/namespace.rs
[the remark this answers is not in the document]
soenkeliebau: As long as we keep all tests in one test method this should never become an issue, but deleting it when the test starts and adding a comment to keep it in mind might still make sense.

maltesander (approved):

soenkeliebau (approved):
LGTM

PR #95, opened 2021-04-08
Added automatic support for app.kubernetes.io/instance label in build_metadata

maltesander:


src/metadata.rs
```diff
@@ -5,13 +5,28 @@ use k8s_openapi::Resource;
 use kube::api::{Meta, ObjectMeta};
 use std::collections::BTreeMap;
 
+/// The name of the application e.g. "mysql"
+pub const APP_KUBERNETES_IO_NAME: &str = "app.kubernetes.io/name";
```
lfrancke, line 9: We have all of these in labels.rs already

src/metadata.rs
```diff
@@ -21,8 +36,14 @@ pub fn build_metadata<T>(
 where
     T: Meta,
 {
+    let mut merged_labels = get_recommended_labels(resource)?;
+
+    if let Some(provided_labels) = labels {
+        merged_labels.extend(provided_labels);
```
lfrancke, line 42: This would overwrite existing ones?

If so: Could you add a comment to the docs about that?

src/metadata.rs
```diff
@@ -33,6 +54,27 @@ where
...
 
+/// Create the kubernetes recommended labels:
+/// - app.kubernetes.io/name - The name of the application e.g. mysql
+/// - app.kubernetes.io/instance - A unique name identifying the instance of an application e.g. mysql-abcxzy
+/// - app.kubernetes.io/version - The current version of the application (e.g., a semantic version, revision hash, etc.) e.g. 5.7.21
+/// - app.kubernetes.io/component - The component within the architecture e.g. database
+/// - app.kubernetes.io/part-of - The name of a higher level application this one is part of e.g. wordpress
+/// - app.kubernetes.io/managed-by - The tool being used to manage the operation of an application e.g. helm
+///
+fn get_recommended_labels<T>(resource: &T) -> OperatorResult<BTreeMap<String, String>>
```
lfrancke, line 65: Could you move this to labels.rs (if you think it makes sense) and maybe remove the labels that are not set?

src/metadata.rs
```diff
@@ -33,6 +54,27 @@ where
...
+where
+    T: Meta,
+{
+    let mut recommended_labels = BTreeMap::new();
+    // TODO: throw error if no name found? Can that even happen?
+    if let Some(name) = &resource.meta().name {
```
lfrancke, line 71: No, that can't happen. I recommend using `Meta::name(resource)` instead.

lfrancke (changes requested):

lfrancke (approved):

PR #96, opened 2021-04-09
Adapt Crd to latest kube-rs changes which automatically generates a schema

lfrancke:


src/crd.rs
```diff
@@ -20,34 +23,48 @@ pub trait Crd {
...
+    ///
+    /// ```text
+    ///     fn crd() -> CustomResourceDefinition {
+    ///         MyCustomResource::crd()
+    ///     }     
+    ///
+    /// ```
+    fn crd() -> CustomResourceDefinition;
+
+    /// Generates a YAML CustomResourceDefinition and writes it to a `Write`r.
```
maltesander, line 43: typo: `Writer`

maltesander (changes requested):
Just a type in comment.

src/crd.rs
maltesander: Typo still here :)

maltesander (changes requested):
LGTM. Just a small typo.

maltesander (approved):

PR #97, opened 2021-04-15
Code optimisation

maltesander:
Removed some unnecessary clones and made the kubernetes client a reference where possible.
This will require adaptation in the operators.


lfrancke (approved):
Thank you!

PR #116, opened 2021-05-20
BREAKING: Make list_pods more generic and rename to list_owned

lfrancke:


src/reconcile.rs
```diff
@@ -132,16 +132,20 @@ where
...
+    /// This lists all Resources that have an OwnerReference that points to us (the object from `self.resource`)
     /// as its Controller.
     ///
     /// Unfortunately the Kubernetes API does _not_ allow filtering by OwnerReference so we have to fetch
-    /// all Pods and filter them on the client.
+    /// all Resources and filter them on the client.
     /// To reduce this overhead a LabelSelector will be included that uses the standard
     /// `app.kubernetes.io/instance` label and will use the name of the resource in this context
     /// as its value.
     /// You need to make sure to always set this label correctly!
```
maltesander, line 143: Maybe a hint for labels::get_recommended_labels? And you could use that instead of (further down):
`let mut labels = BTreeMap::new();`
`labels.insert(labels::APP_INSTANCE_LABEL.to_string(), self.resource.name());`

maltesander (dismissed):
Just one comment. We have the get_recommended_labels method you could use. For now it only builds the instance label. But if we use it consistently for everything (once more labels are added) it should be safe to use here as well?

src/reconcile.rs
lfrancke: Good point, thank you.

lfrancke: Well....that fails the build. I'll need to understand what's going on.

lfrancke: Take a look please. It seems to build now.

maltesander (approved):
Looks good!

PR #120, opened 2021-05-28
Update kube* from 0.52 to 0.56

lfrancke:
There are a few breaking changes in there but no actual code changes outside of those changes.

https://github.com/clux/kube-rs/blob/master/CHANGELOG.md


Cargo.toml
```diff
@@ -10,28 +10,28 @@ version = "0.1.0-nightly"
...
-lazy_static = "1"
-regex = "1"
+kube = { version = "0.55", default-features = false, features = ["jsonpatch"] }
+kube-runtime = "0.55"
+lazy_static = "1.4"
+regex = "1.5"
 schemars = "0.8"
-serde = "1"
+serde = "1.0"
 serde_json = "1"
```
maltesander, line 23: Add minor version to serde_json as well to be consistent?

maltesander (changes requested):
Just a minor version missing in the toml.

maltesander (approved):
Looks good!

PR #131, opened 2021-06-02
Moved exists method from crd.rs to client.rs. 

maltesander:
We use the client get method to simulate an exists, which currently is neither offered by kubernetes or kube-rs.


razvan (approved):
Looks good.

PR #145, opened 2021-06-15
Added Conditions to common status traits

maltesander:


src/status.rs
```diff
@@ -0,0 +1,7 @@
+//! This module provides structs and trades to generalize the custom resource status access.
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Condition;
+
+pub trait Conditions {
+    fn conditions(&self) -> Option<&Vec<Condition>>;
```
lfrancke, line 5: Should this not return:

```suggestion
    fn conditions(&self) -> Option<&[Condition]>;
```

lfrancke (approved):

PR #146, opened 2021-06-17
Update kube-rs to 0.57 and k8s-openapi to 0.12

lfrancke:
This has some breaking changes for downstream users but will allow us to treat Crds generic

https://github.com/clux/kube-rs/blob/master/CHANGELOG.md
https://github.com/Arnavion/k8s-openapi/blob/master/CHANGELOG.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,9 @@ const_format = "0.2"
 either = "1.6"
 futures = "0.3"
 json-patch = "0.2"
-k8s-openapi = { version = "0.11", default-features = false }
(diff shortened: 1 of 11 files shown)


maltesander (dismissed):
Looks good!

lfrancke:
Thanks. Do you think we should merge it now?
It'll probably require changes in all operators.

maltesander:
Yes maybe we wait until we have the other bigger changes merged from product config etc.

maltesander (approved):
LGTM

Issue #147, opened 2021-06-17
Consider giving `client` a configurable timeout value


soenkeliebau:
TODO: Is this still valid and/or can kube-rs specify a timeout now and do we already do so?

I ran some tests today with a weird network setup which basically caused my connection attempts to Kubernetes to never return anything. I think routing was borked and the configured ip was routed somewhere where someone just dropped packages... or something

This caused my test to hang and never go into the retry behavior I wanted to test. 

I probably just didn't wait long enough to hit an internal timeout in kube-rs somewhere, but to make things like this more predictable it might make sense to equip our client with a configurable timeout value after which Futures get cancelled and an error returned.

lfrancke:
This might not be needed anymore after the operator rework

teozkr:
The rework doesn't really change anything here.

lfrancke:
True, but you don't use the client anymore do you? I mean "our" client.
You use kube-rs client directly, no?

lfrancke:
@teozkr What's the status for kube-rs now? Does it now support specifying timeouts?

lfrancke:
Replaced by stackabletech/issues#199

PR #152, opened 2021-06-21
Adds function to check by name if all required CRDs have been registered in Kubernetes

soenkeliebau:
The function also allows specifying a timeout and delay between retries to optionally allow operators to wait until CRDs have been created and then continue their startup.

partially fixes https://github.com/stackabletech/issues/issues/27

follow up PRs are needed in all operators for to fully fix the issue.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -15,6 +15,7 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::convert::TryFrom;
 use std::fmt::Debug;
+use std::time::Duration;
 use tracing::trace;
 
 /// This `Client` can be used to access Kubernetes.
@@ -320,6 +321,31 @@ impl Client {
             Ok(Some(
                 api.delete(&resource.name(), &self.delete_params).await?,
             ))
+        }
+    }
+
+    /// This deletes a resource _if it is not deleted already_ and waits until the deletion is
+    /// performed by Kubernetes.
+    ///
+    /// It calls `delete` to perform the deletion.
+    ///
+    /// Afterwards it loops and checks regularly whether the resource has been deleted
+    /// from Kubernetes
+    pub async fn ensure_deleted<T>(&self, resource: T) -> OperatorResult<()>
+    where
+        T: Clone + Debug + DeserializeOwned + Resource,
+        <T as Resource>::DynamicType: Default,
+    {
+        self.delete(&resource).await?;
+
+        loop {
+            if !self
+                .exists::<T>(&resource.name(), resource.namespace().as_deref())
+                .await?
+            {
+                return Ok(());
+            }
+            tokio::time::sleep(Duration::from_secs(5)).await;
         }
     }
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -337,7 +337,10 @@ impl Client {
         T: Clone + Debug + DeserializeOwned + Resource,
         <T as Resource>::DynamicType: Default,
     {
-        let mut backoff_strategy = ExponentialBackoff::default();
+        let mut backoff_strategy = ExponentialBackoff {
+            max_elapsed_time: None,
+            ..ExponentialBackoff::default()
+        };
 
         self.delete(&resource).await?;
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -3,6 +3,8 @@ use crate::finalizer;
 use crate::label_selector;
 use crate::pod_utils;
 
+use backoff::backoff::Backoff;
+use backoff::ExponentialBackoff;
 use either::Either;
 use futures::StreamExt;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::{Condition, LabelSelector};
@@ -15,8 +17,7 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::convert::TryFrom;
 use std::fmt::Debug;
-use std::time::Duration;
-use tracing::trace;
+use tracing::{error, info, trace};
 
 /// This `Client` can be used to access Kubernetes.
 /// It wraps an underlying [kube::client::Client] and provides some common functionality.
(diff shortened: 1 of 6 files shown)


src/crd.rs
```diff
@@ -60,6 +62,109 @@ where
...
+
+        // Checks done, now we
+        //   1. return ok(()) if all CRDs are present
+        //   2. return an error if CRDs are missing and the timeout has expired
+        //   3. queue another loop iteration if an error occurred and the timeout has not expired
+        match check_result {
+            Ok(()) => return Ok(()),
+            Err(err) => {
+                info!(
+                    "Error occurred when checking if all required CRDs are present: [{}]",
```
lfrancke, line 135: If I understand correctly this will show even for missing CRDs which does not feel like an "error" to me.

It'd be good if you could distinguish between actual errors and missing CRDs.
Errors should probably be `error!` as well while missing CRDs should probably print which CRDs are still missing (which this will do)

tests/test_crd.rs
```diff
@@ -65,3 +65,99 @@ async fn k8s_test_test_ensure_crd_created() {
         .await
         .expect("CRD should be created"))
 }
+
+#[tokio::test]
+#[ignore = "Tests depending on Kubernetes are not ran by default"]
+async fn k8s_test_wait_for_crds() {
+    // TODO: Switch this to using TemporaryResource from the integration-test-commons crate
```
lfrancke, line 72: Every TODO should have a ticket (I'm not sure when the best time is for adding those references either)

tests/test_crd.rs
```diff
@@ -65,3 +65,99 @@ async fn k8s_test_test_ensure_crd_created() {
         .await
         .expect("CRD should be created"))
 }
+
+#[tokio::test]
+#[ignore = "Tests depending on Kubernetes are not ran by default"]
+async fn k8s_test_wait_for_crds() {
```
lfrancke, line 71: A test with two (or more) existing CRDs would be good to have

lfrancke (changes requested):

src/crd.rs
soenkeliebau: True, I'll split that into two different match branches. 

Regarding the log level, I've followed the principle of logging at debug and info here and leaving anything more severe up to the code using this library. Otherwise we'll probably have duplicate logging once here and then the operator will log an error again.
That being said, this might be a bit of a special case since we probably won't return control to the operator here until all CRDs are present (unless the operator specifies a timeout and adds a loop of its own) so logging at warn might make sense..

soenkeliebau:
I've changed the conflicting test cases to run serially and added a cleanup method that waits until resources are fully deleted by Kubernetes, which should hopefully resolve the flakiness.

src/client.rs
```diff
@@ -323,6 +324,31 @@ impl Client {
...
     }
 
+    /// This deletes a resource _if it is not deleted already_ and waits until the deletion is
+    /// performed by Kubernetes.
+    ///
+    /// It calls `delete` to perform the deletion.
+    ///
+    /// Afterwards it loops and checks regularly whether the resource has been deleted
+    /// from Kubernetes
+    pub async fn ensure_deleted<T>(&self, resource: T) -> OperatorResult<()>
```
lfrancke, line 334: Due to the nature of finalizers I believe we need a timeout here.

src/crd.rs
```diff
@@ -130,11 +130,34 @@ pub async fn wait_until_crds_present(
...
+                if let Some(timeout_value) = &timeout {
+                    if timeout.is_some() && start.elapsed() >= *timeout_value {
+                        info!(
+                            "Timeout of [{}] seconds reached, returning.",
+                            timeout_value.as_secs()
+                        );
+                        return Err(RequiredCrdsMissing { names });
+                    }
+                }
+            }
```
lfrancke: Isn't this a duplicate now?
You handle this case below already, am I missing something?

lfrancke (changes requested):

src/client.rs
soenkeliebau: I considered that, but we don't really offer timeouts in many other places and instead let users wrap the call in a timeout themselves if they need it.
I figured this was more in line with other fn calls - but am happy to add it too.
No real preference..

src/crd.rs
soenkeliebau: Nope, you are completely right. That was an earlier attempt at addressing your comment which I forgot to clean up.

src/client.rs
lfrancke: Fair enough

src/client.rs
```diff
@@ -345,7 +348,26 @@ impl Client {
...
+                        backoff.as_secs()
+                    );
+                    tokio::time::sleep(backoff).await;
+                }
+                None => {
+                    // We offer no way of specifying a timeout, so this shouldn't happen,
+                    // if it does we'll log an error for now and continue iterating and wait for
+                    // the last interval we saw
+                    error!(
+                        "Waiting for deletion timed out, but no timeout was specified, this is an error and should not happen!"
```
lfrancke, line 366: Looking at ExponentialBackoff::default it defaults to "expire" after 15min: https://github.com/ihrwein/backoff/blob/master/src/default.rs#L13

I'm not sure if you want to change the docs or the behavior or whether having this backoff crate is overkill.
I'm fine with either of those solutions (or another one entirely)

soenkeliebau: Ups, sorry, I missed that. I'll change the default behavior no never expire.

I considered whether we should add that crate, yes. But I pretty much reimplemented the exact functionality when I did it myself, just with less functionality. And I already found stuff that I wanted to add which would have duplicated even more of that crate.

I looked at the crate and it seemed more or less active, docs looked good and the author says where he took inspiration, which is a fairly established backoff algo by Google.

And with an eye on https://github.com/stackabletech/operator-rs/issues/147 I figured it might be worth pulling this in.

lfrancke (approved):

PR #154, opened 2021-06-23
Added product config helper methods to transform role and group config…

maltesander:
… structure into hashmaps for the validation.

Fixed clippy warnings due to rust 1.53 update.

### DIFF
--- a/src/product_config_utils.rs
+++ b/src/product_config_utils.rs
@@ -29,7 +29,7 @@ pub trait Configuration {
         role_name: &str,
     ) -> Result<BTreeMap<String, String>, ConfigError>;
 
-    fn compute_properties(
+    fn compute_files(
         &self,
         resource: &Self::Configurable,
         role_name: &str,
@@ -37,7 +37,7 @@ pub trait Configuration {
     ) -> Result<BTreeMap<String, String>, ConfigError>;
 }
 
-// This deep map causes problems with clippy and rustfmt.
+// This deep map causes problems with clippy and fmt.
 pub type RoleConfigByPropertyKind =
     HashMap<String, HashMap<String, HashMap<PropertyNameKind, BTreeMap<String, String>>>>;
 ///
@@ -305,7 +305,7 @@ where
     }) = config
     {
         final_properties = inner_config
-            .compute_properties(resource, role_name, file)
+            .compute_files(resource, role_name, file)
             .unwrap();
     }
 
@@ -395,7 +395,7 @@ mod tests {
             Ok(result)
         }
 
-        fn compute_properties(
+        fn compute_files(
             &self,
             _resource: &Self::Configurable,
             _role_name: &str,
--- a/src/product_config_utils.rs
+++ b/src/product_config_utils.rs
@@ -9,10 +9,14 @@ use tracing::{debug, error, warn};
 pub enum ConfigError {
     #[error("Invalid configuration found")]
     InvalidConfiguration,
+
+    #[error("Collected product config validation errors: {collected_errors:?}")]
+    ProductConfigErrors {
+        collected_errors: Vec<product_config::error::Error>,
+    },
 }
 
-///
-/// This needs to be implemented by every [Config] struct that may appear in [Role<Config>]
+/// This needs to be implemented by every generic (T) struct that may appear in [Role<T>]
 /// or the top level in order to determine where config properties are configured.
 ///
 /// The options are:
@@ -23,12 +27,10 @@ pub enum ConfigError {
 /// Returned empty Maps will be ignored. Check out `to_hash_map(...)` in
 /// https://github.com/stackabletech/product-config/blob/main/src/ser.rs
 /// if you do not need to differentiate between the options and want to serialize all properties
-/// in a certain config [Config] to one option.
-///
+/// in a certain config T to one option.
 pub trait Configuration {
     type Configurable;
 
-    // TODO: Not sure if I need the role_name here
     // TODO: We need to pass in the existing config from parents to run validation checks and we should probably also pass in a "final" parameter or have another "finalize" method callback
     //  one for each role group, one for each role and one for all of it...
     fn compute_env(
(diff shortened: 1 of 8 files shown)


razvan (dismissed):
LGTM

lfrancke:
I have already started the review, please wait with merging.

src/product_config_utils.rs
```diff
@@ -0,0 +1,1034 @@
...
+/// In addition, the `*_overrides` settings are also merged in the resulting configuration
+/// with the highest priority.
+/// The merge priority chain looks like this:
+///
+/// group overrides -> group config -> role overrides -> role config
+///
+/// where '->' means "overwrites if existing or adds".
+///
+/// The output is a map with one entry, keyed by `role_name` and the value is a map where all
+/// configuration properties defined in the `role` are partitioned by `PropertyNameKind`.
```
lfrancke, line 149:
```suggestion
/// configuration properties defined in the `role` and `role groups` are partitioned by `PropertyNameKind`.
```

src/product_config_utils.rs
```diff
@@ -0,0 +1,1034 @@
...
+/// # Arguments
+/// - `resource`       - Not used directly. It's passed on to the `Configuration::compute_*` calls.
+/// - `role_name`      - Used as key in the output and to partition the configuration properties.
```
lfrancke, line 152:
```suggestion
/// - `role_name`      - The name of the role.
```

src/product_config_utils.rs
```diff
@@ -0,0 +1,1034 @@
...
+/// The output is map where the configuration properties are keyed by `PropertyNameKind`.
+///
+/// # Arguments
+/// - `resource`       - Not used directly. It's passed on to the `Configuration::compute_*` calls.
+/// - `name`           - Not used directly but passed on to the `Configuration::compute_*` calls. It usually
+///                      contains a role name or a role group name.
+/// - `config`         - The configuration properties to partition.
+/// - `property_kinds` - The "buckets" used to partition the configuration properties.
+///
+fn partition_properties_by_kind<T>(
```
lfrancke, line 207: Why do you call this `partition` when the others are all called `transform`?
(but I'm fine with it)

src/product_config_utils.rs
```diff
@@ -0,0 +1,1034 @@
...
+
+        let role_properties =
+            transform_role_to_config(resource, &role_name, &role, property_name_kinds);
+        result.insert(role_name, role_properties);
+    }
+
+    result
+}
+
+// TODO: boolean flags suck, move ignore_warn to be a flag
```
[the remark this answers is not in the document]
djc: I don't think boolean arguments are outright banned, but if are commonly passing around booleans with the same semantics and/or if you are passing multiple booleans in the same function, using enums will definitely help. (I also will sometimes name the value in the caller, for example have `let ignore_warn = true; process_validation_results(_, ignore_warn, _)`). Here it also looks like `ignore_warn` and `ignore_err` potentially encode some redundant state that could be invalid (would it be valid for a caller to call this with `false, true`?), in which case it could definitely make sense to encode this differently (maybe in a tri-state enum or in a custom enum wrapped in `Option`).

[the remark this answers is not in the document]
maltesander: Yeah i agree, for now the (false, true) state seems a litte odd. I think we can keep it for now (as we just set it globally in the operators. But we should move to a more granular method to distinguish ignore warn / errors for each property like lars proposed:
`pub enum Property {
    Simple(String),
    Complex {
        ignore_warning: bool,
        ignore_error: bool,
        value: String,
    },
}`

src/product_config_utils.rs
```diff
@@ -9,12 +9,28 @@ use tracing::{debug, error, warn};
...
     InvalidConfiguration,
+
+    #[error("Collected product config validation errors: {collected_errors:?}")]
+    ProductConfigErrors {
+        collected_errors: Vec<product_config::error::Error>,
+    },
 }
 
+/// This needs to be implemented by every generic (T) struct that may appear in [Role<T>]
+/// or the top level in order to determine where config properties are configured.
```
lfrancke, line 20:
```suggestion
/// This trait is used to compute configuration properties for products.
```

This is just a suggestion but the current version is also not really correct

src/product_config_utils.rs
```diff
@@ -9,12 +9,28 @@ use tracing::{debug, error, warn};
...
+///
+/// The options are:
+/// - Environmental variables (env)
+/// - Command line arguments (cli)
+/// - Configuration files (files)
+///
+/// Returned empty Maps will be ignored. Check out `to_hash_map(...)` in
+/// https://github.com/stackabletech/product-config/blob/main/src/ser.rs
+/// if you do not need to differentiate between the options and want to serialize all properties
+/// in a certain config T to one option.
```
lfrancke, line 30:
```suggestion
/// Returned empty Maps will be ignored.
/// Check out [`crate::ser::to_hash_map`] if you do not need to convert a struct to a HashMap in an easy way..
```

src/product_config_utils.rs
```diff
@@ -181,16 +201,15 @@ where
...
 ///
 /// # Arguments
 /// - `resource`       - Not used directly. It's passed on to the `Configuration::compute_*` calls.
 /// - `name`           - Not used directly but passed on to the `Configuration::compute_*` calls. It usually
 ///                      contains a role name or a role group name.
 /// - `config`         - The configuration properties to partition.
 /// - `property_kinds` - The "buckets" used to partition the configuration properties.
-///
-fn partition_properties_by_kind<T>(
+fn transform_properties_to_kind<T>(
```
lfrancke, line 212: I'm sorry about not being more specific last time. We don't transform properties to kind here so this new name is also wrong :) 

Either keep the old name or chose something that matches what it actually does like `parse_role_config` or something along those lines or `transform_role_to_config` I don't know

lfrancke (changes requested):

src/product_config_utils.rs
maltesander: I think we lose info here. And i would use to_hash_map if i want an **easy** way to convert and put everything e.g. in env variables.

src/product_config_utils.rs
maltesander: I would merge the existing comment with that one.

src/product_config_utils.rs
lfrancke: You definitely want to fix the rust doc link. And yes, my comment was wrong. It should be `if you do need`.

That said: I don't understand the original sentence: `if you do not need to differentiate between the options and want to serialize all properties in a certain config T to one option.`

@soenkeliebau  as someone who hasn't been part of this. Could you say if this comment works for you as is?

lfrancke (changes requested):

src/product_config_utils.rs
soenkeliebau: I don't really understand the initial comment I'm afraid.

lfrancke (approved):

PR #156, opened 2021-06-25
Add Builder types for various Kubernetes objects

### LINKED ISSUE #88: Support creating events

### LINKED ISSUE #94: Add functionality to set required labels to build_metadata function
We have agreed to use a set of default labels, as per what [Kubernetes best practice expects](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). 

> app.kubernetes.io/name	The name of the application	
> app.kubernetes.io/instance	A unique name identifying the instance of an application	
> app.kubernetes.io/version	The current version of the application (e.g., a semantic version, revision hash, etc.)
> app.kubernetes.io/component	The component within the architecture
> app.kubernetes.io/part-of	The name of a higher level application this one is part of
> app.kubernetes.io/managed-by	The tool being used to manage the operation of an application

It would make sense to have these supported by build_metadata in some form.

### LINKED ISSUE #99: Set common labels when creating objects
We have a set of common labels [1] that we want to use for objects in Kubernetes. Some of these will have to be set by the operator as the necessary information is only known there.
But for some labels this could be moved to the operator framework as well.

For example the 'managed-by' label could be filled by adding a function to the operator / crd that returns a string to be put here, which would then be called by the framework.

[1]https://github.com/stackabletech/operator-rs/blob/main/src/labels.rs

### LINKED ISSUE #137: Add tests and documentation for the new builder types
## The problem

TODO

## Proposed solution

Implemented in the branch: https://github.com/stackabletech/operator-rs/tree/pod_builder

/fixes #94 
/fixes #99 
/fixes #88

lfrancke:
/fixes #94
/fixes #99
/fixes #88
/fixes #137


src/builder.rs
```diff
@@ -0,0 +1,831 @@
...
+            },
+        })
+    }
+}
+
+/// A builder to build [`Pod`] objects.
+///
+#[derive(Clone, Default)]
+pub struct PodBuilder {
+    metadata: Option<ObjectMeta>,
```
razvan, line 595: There are four functions to set metadata and a user can easily overwrite it by mistake.

I think the code would be much cleaner and less error prone if this were an `ObjectMetaBuilder` instead of a "materialized" `ObjectMeta`. The only time this would be materialized would be at `build()` time.

src/builder.rs
```diff
@@ -0,0 +1,831 @@
...
+    status: Option<PodStatus>,
+    containers: Vec<Container>,
+}
+
+impl PodBuilder {
+    pub fn new() -> PodBuilder {
+        PodBuilder::default()
+    }
+
+    pub fn metadata_default(&mut self) -> &mut Self {
```
razvan, line 607: This would not be necessary if `metadata` would actually be a builder.

src/builder.rs
```diff
@@ -0,0 +1,831 @@
...
+        self
+    }
+
+    pub fn metadata_builder<F>(&mut self, f: F) -> OperatorResult<&mut Self>
+    where
+        F: Fn(&mut ObjectMetaBuilder) -> &mut ObjectMetaBuilder,
+    {
+        let mut builder = ObjectMetaBuilder::new();
+        let builder = f(&mut builder);
+        self.metadata = Some(builder.build()?);
```
razvan, line 618: Here you could just reuse the inner `ObjectMetaBuilder`  without calling `build()` on it and possibly overwrite previous metadata.

razvan (approved):
Everything works and most comments are just questions that are not "show stoppers". But  the comments related to `PodBuilder::metadata` are real issues.

src/builder.rs
```diff
@@ -0,0 +1,831 @@
...
+use kube::{Resource, ResourceExt};
+use std::collections::{BTreeMap, HashMap, HashSet};
+
+/// A builder to build [`ConfigMap`] objects.
+#[derive(Clone, Default)]
+pub struct ConfigMapBuilder {
+    metadata: Option<ObjectMeta>,
+    binary_data: BTreeMap<String, ByteString>,
+    data: BTreeMap<String, String>,
+    immutable: Option<bool>,
```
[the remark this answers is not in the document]
lfrancke: Good find. I'm surprised that clippy and IntelliJ don't complain about these...

src/builder.rs
```diff
@@ -0,0 +1,831 @@
...
+
+    /// Consumes the Builder and returns a constructed Node
+    pub fn build(&self) -> Node {
+        // We're cloning here because we can't take just `self` in this method because then
+        // we couldn't chain the method with the others easily (because they return &mut self and not self)
+        self.node.clone()
+    }
+}
+
+impl Default for NodeBuilder {
```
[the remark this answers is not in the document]
lfrancke: This was a suggestion from Clippy but you're correct.

src/builder.rs
```diff
@@ -0,0 +1,831 @@
...
+    }
+
+    /// This will replace all existing labels
+    pub fn labels(&mut self, labels: BTreeMap<String, String>) -> &mut Self {
+        self.labels = labels;
+        self
+    }
+
+    /// This sets the common recommended labels (in the `app.kubernetes.io` namespace).
+    /// It is recommended to always call this method and is mostly not required to make testing easier.
```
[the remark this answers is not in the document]
lfrancke: Yeah, the whole sentence is garbage. I'm not sure what I wanted to say to be honest...

src/builder.rs
lfrancke: I understand your concern I don't think I understand your proposed solution.

Can you elaborate?

razvan: I wanted to prepare a patch, but I discovered that the patch would be quite big, so I' trying to sketch the idea here:

```
#[derive(Clone, Default)]
pub struct PodBuilder {
    metadata_builder: ObjectMetaBuilder,
// other fields skipped
}

impl PodBuilder {
    pub fn new() -> PodBuilder {
        PodBuilder {
            metadata: ObjectMetaBuilder::new(),
            ..PodBuilder::default()
        }
    }

   // the only method you need to set meta data fields
   // keep on reusing the inner object meta builder
    pub fn metadata<F>(&mut self, f: F) -> &mut Self
    where
        F: Fn(&mut ObjectMetaBuilder) -> &mut ObjectMetaBuilder,
    {
        f(&mut self.metadata_builder);
        self
    }

// other methods skipped

    /// Consumes the Builder and returns a constructed Pod
    pub fn build(&self) -> OperatorResult<Pod> {

// skip some code here ... and at the result of build() is:
        Ok(Pod {
            metadata: match self.metadata.build() {
                Err => return Err(Error::MissingObjectKey { key: "metadata" }),
                Ok(ref metadata) => metadata.clone(),
            },
            spec: Some(PodSpec {
                containers: self.containers.clone(),
                tolerations: self.tolerations.clone(),
                volumes,
                node_name: self.node_name.clone(),
                ..PodSpec::default()
            }),
            status: self.status.clone(),
        })
   }
}
```

lfrancke: I added the `metadata_default` so it can be used in tests mor easily.
I think `metadata` makes sense for when you really do want to override it all
`metadata_opt` can go for now I agree

I like your idea of having a reusable builder in theory. In practice I'm not sure how often that'd be used really.

razvan (approved):
Fine.

PR #170, opened 2021-07-20
Added GSL crate which is GPL licensed to test if the cargo deny actio…

soenkeliebau:
…n catches this.


soenkeliebau:
It didn't fail for the reason I was hoping it would fail, but I assume this is because the crate I used for testing is not adhering to something. It failed, that is the most important part.

PR #177, opened 2021-07-28
RoleGroup replicas now are Option<u16>

maltesander:
None acts as wildcard to schedule one pod on each node.

### DIFF
--- a/src/k8s_utils.rs
+++ b/src/k8s_utils.rs
@@ -41,7 +41,7 @@ pub fn find_excess_pods<'a>(
 /// Finds all pods matching the required labels.
 /// NOTE: This list can contain duplicates!
 fn find_expected_pods<'a>(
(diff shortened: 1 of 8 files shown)


src/k8s_utils.rs
```diff
@@ -33,16 +33,21 @@ pub fn find_excess_pods<'a>(
...
+            None => used_pods.append(&mut found_pods),
+            Some(replicas) => {
+                used_pods.append(
+                    &mut found_pods
+                        .choose_multiple(&mut rand::thread_rng(), usize::from(*replicas))
+                        .cloned()
+                        .collect(),
+                );
+            }
+        }
```
razvan, line 50:
```suggestion
        used_pods = replicas
            .map(|c| {
                found_pods
                    .choose_multiple(&mut rand::thread_rng(), usize::from(*c))
                    .cloned()
                    .collect()
            })
            .or(Option::from(found_pods))
            .unwrap();
```

razvan (approved):
ok

Issue #144, opened 2021-06-15
Add supporting functionality for restarts: Rolling and non-rolling


lfrancke:
It should be possible for the operator framework to handle restarts (either rolling or non-rolling).

One idea is for the operator to provide the framework with a list of pods that need to be restarted and a RestartPolicy as well as a RequeuePolicy.

PR #184, opened 2021-08-09
Moving command handling to the framework and implement default handling for start, stop and simple restart

soenkeliebau:
Adds traits and methods to the operator framework which offer convenience methods around command handling.

Finding the currently running command or next command that should be run as well as tracking this in the status is handled by the operator framework.

Additionally Traits have been added which operators can implement for their command objects, this allows some generic handling to be done in the framework, while keeping the flexibility for operators to add their own information to command objects if needed for more complex operations.

Default implementations can be used for start, stop and restart by the operators. 

Restart is currently limited in how much of a rolling functionality it offers, as it will still delete all pods before restarting one - but it will delete the pods in a rolling fashion. See https://github.com/stackabletech/operator-rs/issues/144 for a follow up issue.


src/reconcile.rs
```diff
@@ -285,6 +317,226 @@ where
...
+
+    pub async fn default_restart<C, P>(
+        &self,
+        command: &C,
+        pod_provider: &P,
+    ) -> ReconcileResult<Error>
+    where
+        T: HasApplication + HasRoleRestartOrder,
+        C: Command + CanBeRolling + HasRoles,
+        // TODO: Not sure if we can skip 'HasRoles' here and conditinally run code below if it is implemented
```
maltesander, line 383: I dont think that works. You can implement it "conditionally" via impl blocks with different where clauses. But that wont safe much code. 
But i think we have to differentiate. You cannot call this method now if no HasRoles is implemented?

src/reconcile.rs
```diff
@@ -285,6 +317,226 @@ where
...
+                    "Skipping role [{}] during restart, no pods left to restart.",
+                    role
+                );
+                continue;
+            }
+
+            // Track if anything was changed during this run
+            restart_occurred = true;
+            // Restart pods depending on strategy
+            match command.is_rolling() {
```
maltesander, line 434: Here as well. This wont compile if we just leave it out. I think we need more impl blocks.

src/reconcile.rs
```diff
@@ -285,6 +317,226 @@ where
...
+                    let node = current_pod
+                        .spec
+                        .clone()
+                        .unwrap_or_default()
+                        .node_name
+                        .unwrap();
+
+                    self.client.ensure_deleted(current_pod.clone()).await?;
+
+                    let (pods, context) = pod_provider
```
maltesander, line 449: Maybe add a comment that this will be obsolete once https://github.com/stackabletech/operator- rs/issues/207 is implemented?

src/command_controller.rs
```diff
@@ -69,28 +69,35 @@
...
 /// main controller custom resource.
 /// The referenced resource has to be in the same namespace as the command itself.
-pub trait Command {
+pub trait Command: Resource {
     /// Retrieve the potential "Owner" name of this custom resource
     fn get_owner_name(&self) -> String;
+    fn start(&mut self);
+    fn done(&mut self);
+    fn start_time(&self) -> Option<DateTime<FixedOffset>>;
+    fn get_start_patch(&self) -> Value;
```
maltesander, line 100: We could use and `end_time` here too?

maltesander (changes requested):
I didnt try the logic yet, i will test once spark is adapted as well. 
Looks good to me already, just some comments about code / naming etc.

src/reconcile.rs
```diff
@@ -285,6 +298,336 @@ where
...
+            + CustomResourceExt
+            + HasCommands
+            + HasStatus
+            + Send
+            + Sync
+            + 'static,
+        <T as Resource>::DynamicType: Default,
+        <T as HasStatus>::Status: HasCurrentCommand + Debug + Default + Serialize,
+    {
+        let current_command_ref = crate::command::current_command(
```
soenkeliebau, line 612: I think we could make this more testable if we retrieved the status and list of commandrefs here and passed those into `current_command` instead of a client and a resource..

That would mean that we needlessly retrieve all commands even if a command is set in the status on every reconciliation though.

src/crd.rs
```diff
@@ -223,3 +223,11 @@ where
         .await;
     Ok(())
 }
+
+pub trait HasInstance {
```
soenkeliebau, line 227: These should get some comments. Maybe add a todo if the comments become longer..

src/status.rs
```diff
@@ -20,12 +23,34 @@ pub trait Conditions {
     fn conditions_mut(&mut self) -> &mut Vec<Condition>;
 }
 
+/// Provides access to the custom resource status to read, write or delete the current_command.
+pub trait HasCurrentCommand {
+    fn current_command(&self) -> Option<CommandRef>;
+    // TODO: setters are non-rusty, is there a better way? Dirkjan?
+    fn set_current_command(&mut self, command: CommandRef);
+    fn clear_current_command(&mut self);
```
soenkeliebau, line 31: I think this is never used - we use clear_current_command from command.rs instead..we could move this over to the same `get_patch` pattern we've used for other stuff maybe?

soenkeliebau (commented):
LGTM overall

We should go trough and add some more comments and refactor stuff a bit to make it testabler, as well as add tests, but in the interest of getting this in before the hackathon tomorrow I think we can merge as is for now.

Maybe add a couple of todos before merging..

src/status.rs
maltesander: It is used in the same method. Not sure what to do here?

src/crd.rs
maltesander: added small comment.

maltesander (approved):

src/reconcile.rs
maltesander: We decided on doing that after the hackathon.

PR #187, opened 2021-08-11
Adding methods to deal with CRDs in CLI interfaces

### LINKED ISSUE #103: Support for common CLI options in operators
The Stackable agent has a `--version` option to print out the version of the installed binary. This is used for example by T2 to collect the versions of all installed Stackable components in a cluster.
The operators should have an option to print out the version as well.

In addition to that we want a `help` option as well as support to print and save CRDs.

lfrancke:
/closes #103


razvan (commented):
I think a short module level documentation is needed that explains how the actual command line arguments will look like if one uses the functions in this module.

src/cli.rs
```diff
@@ -64,6 +69,121 @@ fn check_path(
...
+/// This will handle a subcommand generated by the [`self::generate_crd_subcommand`] method.
+///
+/// The CRD and the name of the subcommand will be identified by the `kind` of the generic parameter `T` being passed in.
+///
+/// # Arguments
+///
+/// * `matches`: The [`ArgMatches`] object which _might_ contain a match for our current CRD.
+///
+/// returns: A boolean wrapped in a result indicating whether the this method did handle the argument.
+///          If it returns `Ok(true)` the program should abort.
```
razvan, line 139: Can we  enforce that the program should abort and do `std::process::exit(0)` instead of returning `Ok(true)`?

src/cli.rs
```diff
@@ -64,6 +69,121 @@ fn check_path(
...
+///
+/// let command = stackable_operator::cli::generate_crd_subcommand::<FooCluster>();
+/// ```
+pub fn generate_crd_subcommand<'a, 'b, T>() -> App<'a, 'b>
+where
+    T: CustomResourceExt,
+{
+    let kind = T::api_resource().kind;
+
+    SubCommand::with_name(&kind.to_lowercase())
```
soenkeliebau, line 112: If I'm not misreading this a subcommand per custom resource would be generated? 

So from a user perspective, I'd need to run this once per crd (cluster, start, stop, restart, rebalance, ...) to print or save the crd?

```
spark-operator restart print
spark-operator stop save stop.yaml
```

I assume this is in order to nest these under a `crd` command from the operator code? 
Are we planning to add a `--all` parameter as well, that would probably need to be done by the operator itself?

src/cli.rs
soenkeliebau: We could probably exit here, but I'm not sure if that is a call that a library should make. 
Only the application using this method can know what the exact repercussions are if this fails and if it needs to abort or can maybe somehow recover..

soenkeliebau:
> I think a short module level documentation is needed that explains how the actual command line arguments will look like if one uses the functions in this module.

I agree. Could be pretty much copied straight from the Spark operator I think.

One more thing I noticed, there is no mention of a command to auto create the CRDs in Kubernetes. Has that just been left out for this PR, or been decided against entirely?

src/cli.rs
lfrancke: We could make a builder out of this and basically have a CrdSubcommandBuilder to which we could add new Subcommands which it could then put into a list and automatically add the `all` function.... I'll think about it for a bit.

lfrancke: We can do that but we can't generalize the handling I believe

soenkeliebau: > We can do that but we can't generalize the handling I believe

This is in reference to the `--all` flag I assume? 
If so, then I think I agree. I think we could introduce a new struct to represent CustomResources along these lines:

```
struct StackableCrd {
  definition: String,
  resource: ApiResource,
}
```
Then that could be exposed via a trait on the cluster object and retrieved in a `Vec<StackableCrd>` by the framework to handle the all parameter, or passed into the function doing the handling.

I hesitate to suggest this for just the cli handling though, but this might actually turn out to be usefull down the line as well...

I have something like this in the command handling code:

```
pub trait HasCommands {
    fn get_command_types() -> Vec<ApiResource>;
}
```
Which is half of what I suggested above already ...

lfrancke: I'm not sure I understand.
`definition` is the schema?

The problem with your proposal (if I understand it correctly) is that we don't actually have an instance of an object. All we have is the type.

When generating the command we could generate and store the schema and then later don't use the types anymore but do everything based on strings (this sentence probably doesn't make much sense to you...)

I'm not sure if it's worth it. I'm in favor of merging as is. We can improve later.

soenkeliebau: I totally agree, lets merge this as is!

Regarding my idea, I was thinking something along these lines as an associated function on the cluster level structs:

```rust
impl HasCommands for SparkCluster {
    fn get_command_types() -> Vec<StackableCrd> {
        vec![
            StackableCrd{ definition: Start::get_schema(), resource: Start::api_resource() },
            StackableCrd{ definition: Stop::get_schema(), resource: Stop::api_resource() },
            StackableCrd{ definition: Restart::get_schema(), resource: Restart::api_resource() },
        ]
    }
}
```

That would give us something to iterate over in the framework and should contain everything we need to create the crds.

soenkeliebau (approved):
LGTM

PR #190, opened 2021-08-12
Fix command controller

lfrancke:
I made a mistake when introducing the builder patterns and created a ObjectMeta instead of an OwnerReference by accident.


razvan (approved):
ok

PR #192, opened 2021-08-13
Adds a simple method to print startup diagnostic information

lfrancke:
This is related to https://github.com/stackabletech/issues/issues/19

I don't think we can include the full `bult` in `operator-rs` because it auto-generates a file at build time (using a `build.rs` script) and if we include this in operator-rs we'd get the information of `operator-rs` and not of the using library.

See https://github.com/stackabletech/spark-operator/pull/118 for an example on how to use it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,6 +21,7 @@ pub mod product_config_utils;
 pub mod reconcile;
 pub mod role_utils;
 pub mod status;
+pub mod utils;
 pub mod validation;
 
 pub use crate::crd::CustomResourceExt;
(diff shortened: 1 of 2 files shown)


soenkeliebau:
> I don't think we can include the full `bult` in `operator-rs` because it auto-generates a file at build time (using a `build.rs` script) and if we include this in operator-rs we'd get the information of `operator-rs` and not of the using library.

Not sure if this is how it works, but couldn't the framework theoretically have been built at a different time and cargo just used an "old" version when compiling the actual operator (theroretically even with a different version of rustc?)?

soenkeliebau (approved):
LGTM

PR #198, opened 2021-08-20
Resource name generation

maltesander:


src/name_utils.rs
```diff
@@ -0,0 +1,253 @@
...
+fn strip(kind: SubName, sub_name: &str) -> OperatorResult<String> {
+    let alphanumeric: String = sub_name.chars().filter(|c| c.is_alphanumeric()).collect();
+
+    let min = kind.min();
+    if alphanumeric.len() < usize::from(min) {
+        return Err(Error::SubNameTooShort {
+            kind: kind.to_string(),
+            name: sub_name.to_string(),
+            min,
+        });
```
razvan, line 151: How about padding the sub_name with nulls (0) for example instead of erroring because it's too small ? It's what the caller would probably have to do anyway.

src/name_utils.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+pub fn build_resource_name(
+    short_name: &str,
+    cluster_name: &str,
+    role_name: &str,
+    group_name: &str,
+    node_name: Option<&str>,
+    misc_name: Option<&str>,
+) -> OperatorResult<String> {
+    let mut full_name = String::new();
```
razvan, line 106: If `full_name` were a vec you could just append to it and call `join("-")` at the end.

src/name_utils.rs
```diff
@@ -0,0 +1,253 @@
...
+            kind: kind.to_string(),
+            name: sub_name.to_string(),
+            min,
+        });
+    }
+
+    Ok(alphanumeric
+        .chars()
+        .into_iter()
+        .take(usize::from(kind.max()))
```
razvan, line 157: If characters are dropped, a  warning should probably be logged out.

maltesander:
I changed the process a little. Fixed length (min, max) are removed. Now each sub name receives an equal share of the possible characters (63 - reserved_for_hash - len(sub_names) [which are the dashes we add later]). If a sub_name is shorter than its share the characters may be used for another sub name.

src/name_utils.rs
```diff
@@ -0,0 +1,399 @@
...
+/// returns its size.
+///
+/// # Arguments
+///
+/// * `sub_names` - A vector of available sub names.
+/// * `max_chars` - The maximum available characters:
+///                 `RESOURCE_NAME_MAX_LEN` - `KUBERNETES_HASH_MIN_LENGTH` - sub_names.len()
+///
+fn selectable_chars(sub_names: &[String], max_chars: usize) -> usize {
+    (1f32 / sub_names.len() as f32 * max_chars as f32).floor() as usize
```
razvan, line 143: This is losing remainder characters. I think this is better:

```suggestion
        max_chars / sub_names.len() + max_chars % sub_names.len()
```

After this change, two tests need to be updated.

src/name_utils.rs
```diff
@@ -0,0 +1,399 @@
...
+            .collect();
+
+        // if the sub name was extended via carryover, we need to adapt how many
+        // carryover characters we have left.
+        if selected.len() > selectable_chars {
+            carryover -= selected.len() - selectable_chars;
+        }
+
+        full_name.push_str(&selected);
+        full_name.push('-');
```
razvan, line 198: This always adds a '-' at the end. Not  good. Use a `Vec<String>` for `full_name` and do `full_name.join("-")` at the end.

src/name_utils.rs
```diff
@@ -0,0 +1,399 @@
...
+use crate::error::OperatorResult;
+
+/// Adjustable number of characters reserved for a generated Kubernetes hash.
+const KUBERNETES_HASH_MIN_LENGTH: usize = 8;
+/// This is the overall fixed length for resource names.
+/// WARNING: Do not change this number unless any specifications change.
+const RESOURCE_NAME_MAX_LEN: usize = 63;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
razvan, line 26: Just wondering why these are not in error.rs.

src/name_utils.rs
```diff
@@ -0,0 +1,399 @@
...
+        sub_names.push(to_alphanumeric_lowercase_not_empty(SubName::Node, node)?);
+    };
+
+    if let Some(misc) = misc_name {
+        sub_names.push(to_alphanumeric_lowercase_not_empty(SubName::Misc, misc)?);
+    };
+
+    // Calculate the maximum number of available characters
+    // The sub names vector length equals the size of additional required dashes
+    let max_chars = RESOURCE_NAME_MAX_LEN - KUBERNETES_HASH_MIN_LENGTH - sub_names.len();
```
razvan, line 119: You can add one (+1) to this if you remove the trailing '-' that is always added now.

src/name_utils.rs
maltesander: Yes thats what we want. It is intended to work with generatedName from kubernetes, which will append a hash at the end. Wont look good without that.

src/name_utils.rs
maltesander: It will stay (check above).

src/name_utils.rs
maltesander: I like to have fewer operator errors and derive from smaller errors. Its just a preference, no real difference. Since its not an external library it would make sense in the main errors though.

src/name_utils.rs
maltesander: Done.

razvan (approved):
I didn't realize that we are not building the final resource name. Now I understand. It looks good.

PR #199, opened 2021-08-23
Added PodSecurityContextBuilder

razvan:
Required to be able to run pods and containers with users other than root.

Part of https://github.com/stackabletech/issues/issues/59


maltesander (approved):
LGTM.

PR #210, opened 2021-09-14
Implement sticky scheduling

### LINKED ISSUE #207: Implement sticky scheduling
## Description
Currently deleted and restarted pods are randomly scheduled on nodes that fit a label selector.

We want to assign pods of a certain role and role group (even unique required ids like `myid` in ZooKeeper) to fixed nodes, in order to reuse data and settings if pods are restarted or temporarily unavailable. 

Therefore we need to keep track of assigned nodes and id information in the custom resource status.

Info: [Hackathon Kubernetes](https://docs.google.com/document/d/13sCfs-DxCfzdDQUQClONSjZpQIMtyNSQktNpxmoXmJE/edit)

## What it does
* Maps pods to nodes
* requirements: Operators generate pod identities which are the basis of scheduling.
* Given a Pod id, map this pod to one and the same node (stable mapping)
* Role group anti-affinity : pods within a role group should not be mapped to the same node.

## What it doesn't do

* Doesn't crate or delete pods.

razvan:
## Description

Implement a pod scheduler that reuses nodes for the given pod ids.

Fixes #207 

## Review Checklist


CHANGELOG.md
```diff
@@ -2,9 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
-## 0.2.0 - unreleased
+## [Unreleased]
 
-## 0.1.0 - 2021-09-01
+### Added
+- Pod placement scheduler with role+group anti-affinity and K8S stored history.
```
maltesander, line 8: Can you check e.g. [agent](https://github.com/stackabletech/agent/blob/main/CHANGELOG.md) changelog to be consistent? Basically linking the PR the change / add / delete occurs?

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+        unscheduled_pods: Vec<PodIdentity>,
+    },
+
+    #[error("PodIdentity could not be parsed: {pod_id:?}. This should not be happening. Please open a ticket.")]
+    PodIdentityNotParseable { pod_id: String },
+}
+
+/// Returns a Vec of pod identities according to the replica per role+group pair from `eligible_nodes`.
+/// # Arguments
+/// * `app_name` : Application name
```
maltesander, line 61: The rusty way would be ```/// * `app_name` - Application name``` (the dash)

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+pub struct RoleGroupEligibleNodes {
+    node_set: BTreeMap<String, BTreeMap<String, Vec<NodeIdentity>>>,
+}
+
+/// Represents the successful result of a `schedule()`
+/// It contains the current scheduled pods and the remaining pods to be scheduled (`new_mapping`)
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct SchedulerState {
+    current_mapping: PodToNodeMapping,
+    new_mapping: PodToNodeMapping,
```
maltesander, line 144: I actually like the name in the comment (new_mapping -> remaining_mapping or unscheduled_mapping)?
I think new mapping is a bit confusion, since the mapping isnt new at all?

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+}
+
+pub type SchedulerResult<T> = std::result::Result<T, Error>;
+
+/// Schedule pods to nodes. Implementations might use different strategies to select nodes based on
+/// the current mapping of pods to nodes or ignore this completely.
+pub trait Scheduler {
+    /// Returns the state of the schedule which describes both the existing mapping as well as
```
maltesander, line 152: ```Returns the state of the scheduler which describes both the existing mapping as well as```
And if you adapt new_mapping do not for get to change here as well.

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+    /// the newly mapped pods.
+    ///
+    /// Implementations may return an error if not all pods can be mapped to nodes.
+    ///
+    /// # Arguments:
+    /// * `pods` : The list of desired pods. Should contain both already mapped as well as new pods.
```
maltesander, line 158: I would use the dash instead of colon here as well (wont comment that anymore)

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+        nodes: &RoleGroupEligibleNodes,
+        mapped_pods: &PodToNodeMapping,
+    ) -> SchedulerResult<SchedulerState>;
+}
+
+pub trait PodPlacementStrategy {
+    /// Returns the nodes where each pod should be placed or `None` if the placement for the pod
+    /// is not possible.
+    /// Assigns `pods` to `NodeIdentities`. For each pod to be placed, if the corresponding
+    /// node in `preferred_nodes` is `Some()` that try to choose this node.
```
maltesander, line 173: ```node in `preferred_nodes` is `Some()` then try to choose this node.```

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+    eligible_nodes: &'a RoleGroupEligibleNodes,
+    hasher: RefCell<DefaultHasher>,
+}
+
+pub enum ScheduleStrategy {
+    /// A scheduling strategy that will refuse to schedule two pods within one role+group on the same
+    /// node. If no enough pods are available, the pod will not be scheduled on any node.
+    /// This useful for when pods are deployed on a bare metal K8S environment with Stackable agents as nodes.
+    GroupAntiAffinity,
+    /// A scheduling strategy that will simply hash the pod onto one of the existing pods without
```
maltesander, line 207: ```A scheduling strategy that will simply hash the pod onto one of the existing nodes without```
Not sure but i think you mean nodes here?

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+            {
+                Ok(res) => {
+                    self.modified = false;
+                    Ok(res)
+                }
+                err => err,
+            };
+        }
+
+        Ok(resource.clone())
```
maltesander, line 254: I have a similar problem with the versioning. I dont like the clone everytime when basically nothing happens. 

Ill update here once i figured it out, or maybe @siegfriedweber has an idea.

We cannot return a reference here, i think we should just mut& T the resource?

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+
+    pub fn new_mapping(&self) -> PodToNodeMapping {
+        self.new_mapping.clone()
+    }
+}
+
+impl TryFrom<String> for PodIdentity {
+    type Error = Error;
+    fn try_from(s: String) -> Result<Self, Error> {
+        let split = s.split(';').collect::<Vec<&str>>();
```
maltesander, line 278: We should investigate if ";" really is a good seperator and nothing can interfere.

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+                NodeIdentity {
+                    name: pod
+                        .spec
+                        .as_ref()
+                        .unwrap()
+                        .node_name
+                        .as_ref()
+                        .unwrap()
+                        .to_string(),
+                },
```
maltesander, line 328: I know that should be safe, but can we still remove the pure `unwrap()` calls and throw an error?

src/scheduler.rs
```diff
@@ -0,0 +1,1013 @@
...
+                eligible_nodes.clone(),
+                current_mapping,
+            )),
+        }
+    }
+
+    /// Returns the new pod mapping or an error if not all desired pods could be mapped.
+    /// As a side effect, it updates the scheduler history.
+    /// # Arguments
+    /// * [`pods`] : pods that are not scheduled yet
```
maltesander, line 522: You do not need the brackets here.

maltesander (changes requested):
I checked mostly comments etc. 
It is alot of code, lots of structs and alot of traits. Can we split that up somehow to distinct clearer?
E.g. The general scheduler stuff in one file, and each implementation or strategy in a separate file? Probably even worth of its own module especially if it grows.
Just a question. Lets see how @siegfriedweber will dig through that ;)

src/scheduler.rs
siegfriedweber: You could return a `std::borrow::Cow`. But I am not sure if the extra complexity is worth it.

siegfriedweber (changes requested):
I am overwhelmed by this pull request. I added some notes regarding the programming style for now. But I will have a look at the functionality later.

src/scheduler.rs
razvan: We decided to leave it as is for now.

src/scheduler.rs
razvan: We decided to leave it as is. A warning is logged if any of the pod identity fields contains a semicolon. This restriction might be removed in the future if a different implementation of the pod placement history will come up.

src/scheduler.rs
```diff
@@ -0,0 +1,1041 @@
...
+        }
+        if role.contains(SEMICOLON) {
+            warn!("Found semicolon in role name: {}", role);
+        }
+        if group.contains(SEMICOLON) {
+            warn!("Found semicolon in group name: {}", group);
+        }
+        if id.contains(SEMICOLON) {
+            warn!("Found semicolon in pod id: {}", id);
+        }
```
maltesander, line 736: Maybe use `warn!("Found ['{}'] in application name: {}", SEMICOLON, app);` in case it changes?

maltesander (changes requested):
Just one small change.

maltesander (approved):
LGTM.

src/scheduler.rs
```diff
@@ -0,0 +1,1056 @@
...
+use std::cell::RefCell;
+use std::collections::btree_map::Iter;
+use std::collections::hash_map::DefaultHasher;
+use std::convert::TryFrom;
+use std::hash::{Hash, Hasher};
+use std::ops::DerefMut;
+use tracing::{error, warn};
+
+const DEFAULT_NODE_NAME: &str = "<no-nodename-set>";
+const SEMICOLON: &str = ";";
```
siegfriedweber, line 45: Could be named according to its purpose, like `POD_IDENTITY_SEPARATOR`.

siegfriedweber (approved):
LGTM (with one optional proposal)

src/scheduler.rs
razvan: Agreed that it will be done in the next round of refactoring.

PR #211, opened 2021-09-14
Extracted the versioning from the operators, added required traits for generic access

maltesander:
## Description

First tests (with zookeeper) work.

## Review Checklist

### DIFF
--- a/src/versioning.rs
+++ b/src/versioning.rs
@@ -3,21 +3,18 @@
 //! The [`crate::status::Conditions`] and [`crate::status::Versioned`] must be implemented
 //! for the custom resource status to ensure generic access.
 //!
-//! The status field names are fixed (for patching updates) and should be defined in the operators
-//! as follows:
+//! The status field names ("conditions" and "version") are fixed (for patching updates) and should
+//! be defined in the operators as follows:
 //! ```
 //! use k8s_openapi::apimachinery::pkg::apis::meta::v1::Condition;
-//! use schemars::JsonSchema;
-//! use serde::{Deserialize, Serialize};
 //! use stackable_operator::versioning::ProductVersion;
-//! #[derive(Clone, Debug, Default, Deserialize, JsonSchema, Serialize)]
-//! #[serde(rename_all = "camelCase")]
+//!
+//! pub enum SomeVersion { SomeVersion }
+//!
 //! pub struct SomeClusterStatus {
-//! #[serde(default, skip_serializing_if = "Vec::is_empty")]
-//! #[schemars(schema_with = "stackable_operator::conditions::schema")]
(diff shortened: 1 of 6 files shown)


src/versioning.rs
```diff
@@ -0,0 +1,364 @@
...
+/// * `resource` - The cluster custom resource.
+/// * `current_version` - The current version set in the status `ProductVersion`.
+/// * `target_version` - The target version set in the status `ProductVersion`.
+/// * `spec_version` - The version currently specified in the custom resource.
+/// * `conditions` - The conditions from the custom resource status.
+///
+fn build_version_and_condition<T, V>(
+    resource: &T,
+    current_version: &Option<V>,
+    target_version: &Option<V>,
```
razvan, line 187: Should the be a single ProductVersion<V> Parameter?

maltesander: Done

src/versioning.rs
```diff
@@ -0,0 +1,540 @@
...
+}
+
+/// Builds a `ProductVersion` to be written into the custom resource status.
+///
+/// # Arguments
+///
+/// * `current_version` - The optional current version of the cluster.
+/// * `target_version` - The optional target version for upgrading / downgrading the cluster.
+///
+fn build_version<V>(current_version: Option<V>, target_version: Option<V>) -> ProductVersion<V> {
```
razvan, line 350: I think `ProductVersion::new()` is more conventional.

src/versioning.rs
```diff
@@ -0,0 +1,540 @@
...
+    V: Clone + Debug + Display + PartialEq + Serialize + Versioning,
+{
+    if let Some(status) = resource.status() {
+        if let Some(version) = status.version() {
+            if let Some(target_version) = &version.target {
+                let condition = build_versioning_condition(
+                    resource,
+                    status.conditions(),
+                    &format!(
+                        "No upgrade required [{}] is still the current_version",
```
razvan, line 148: We shouldn't assume an "upgrade". Maybe a more neutral message like: "Target version reached..."

src/versioning.rs
```diff
@@ -0,0 +1,540 @@
+//! This module handles up and downgrades for the products handled by the operators.
```
razvan, line 1: I think it would be more accurate to say that:
```suggestion
//! It updates the `UpOrDowngrading` condition by comparing the current and the target cluster versions.

```

razvan (commented):
Some clarifications required.

src/versioning.rs
maltesander: yeah that is definitly right, need to adapt :D

razvan (approved):
ok

PR #225, opened 2021-10-05
Require a context for fieldManager when using Server-Side Apply

teozkr:
## Description
This is required so that different applies don't revert each other. For example, `set_condition` would trample on other conditions set by the same operator (actually, it still does that for another reason, #226).

This is technically breaking (adding a param to the public API), but no Stackable repos currently use the affected methods.

## Review Checklist

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -58,7 +58,7 @@ impl Client {
     /// since it will revert changes that are owned by the `field_manager` but not part of the Apply request.
     fn apply_patch_params(&self, field_manager_scope: impl Display) -> PatchParams {
         let mut params = self.patch_params.clone();
-        // TODO: According to https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller we should always force conflicts in controllers.
+        // According to https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller we should always force conflicts in controllers.
         params.force = true;
         if let Some(manager) = &mut params.field_manager {
             *manager = format!("{}/{}", manager, field_manager_scope);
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -40,27 +40,32 @@ impl Client {
...
                 ..PatchParams::default()
             },
             delete_params: DeleteParams::default(),
             default_namespace,
         }
     }
 
+    /// Server-side apply requires a `field_manager` that uniquely identifies a single usage site,
+    /// since it will revert changes that are owned by the `field_manager` but not part of the Apply request.
+    fn apply_patch_params(&self, context: impl Display) -> PatchParams {
```
teozkr, line 59: Not really a fan of the name `context`, since it conflicts with `Context`. `scope` has similar problems (`kube` already uses `scope` to talk about namespaced vs cluster-global). `field_manager_scope`, maybe?

src/client.rs
```diff
@@ -40,27 +40,32 @@ impl Client {
...
+        let mut params = self.patch_params.clone();
+        // TODO: According to https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller we should always force conflicts in controllers.
```
soenkeliebau, line 61: Do we still need this todo, seems like the next line forces conflicts, which this comment is about unless I misread it?

src/client.rs
soenkeliebau: `field_manager_scope` would work I guess, maybe something with `reference` was the only other thing I came up with ..

soenkeliebau (approved):
LGTM

src/client.rs
teozkr: Oh right, that shouldn't be a TODO anymore. Think there's still value in keeping it there as a reference though.

soenkeliebau (approved):
LGTM

PR #226, opened 2021-10-05
Annotate .status.conditions as map-list

teozkr:
This allows server-side apply to update conditions individually. Sadly, it must
be applied manually for each CRD to take effect.

## Review Checklist

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -397,6 +397,9 @@ impl Client {
 
     /// Sets a condition on a status.
     /// This will only work if there is a `status` subresource **and** it has a `conditions` array.
+    ///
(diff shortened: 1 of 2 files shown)


lfrancke:
Only on my phone: We used to have this and removed it when k8s-openapi introduced the schemas. Does this mean this should really be upstream?

teozkr:
I'll look into it, but sadly I don't /think/ schemars allows this kind of customization on the type level without opting out of the derive macro completely.

teozkr:
Yep.. https://github.com/GREsau/schemars/issues/50

lfrancke:
Okay, good point. Thanks.
Watching that issue now.
We had a "manual" schema for one other thing that we removed....can't remember which... LabelSelector maybe?

PR #256, opened 2021-11-05
Release 0.4.0

razvan:
Release 0.4.0. DO NOT SQUASH MERGE!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ description = "Stackable Operator Framework"
 edition = "2018"
 license = "Apache-2.0"
 name = "stackable-operator"
-version = "0.4.0-nightly"
+version = "0.4.0"
 repository = "https://github.com/stackabletech/operator-rs"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -4,7 +4,7 @@ description = "Stackable Operator Framework"
 edition = "2018"
 license = "Apache-2.0"
 name = "stackable-operator"
-version = "0.4.0-nightly"
+version = "0.5.0-nightly"
```
fhennig, line 7: Why did you bump the version here?
If that's required, it should also be in the changelog

razvan: No, this is the _next development version_. In the CHANGELOG the `[Unreleased]` line will be replaced with 0.5.0 the next time we make a release.

fhennig: Alright :+1:

fhennig (approved):
good :+1:

PR #261, opened 2021-11-15
Add a helper for obtaining the selector labels for a certain role group

teozkr:
## Description

`get_recommended_labels` is too specific (includes stuff like the version number), while `build_common_labels_for_all_managed_resources` is too generic (doesn't include the role (group)).

Also redefines all of the helpers, so that they refine each others' output rather than inserting `APP_NAME_LABEL` in three different places.

## Review Checklist


maltesander (approved):
LGTM.

PR #262, opened 2021-11-16
Don't mandate boxing role configs

teozkr:
## Description

For products where all roles take the same `Configuration` this means that you no longer need the `Role::into`. However, products that do have roles with different `Configuration` types you now need to use `Role::erase` instead.

## Review Checklist
- [X] Code contains useful comments
- [X] (Integration-)Test cases added (or not applicable)
- [X] Documentation added (or not applicable)
- [X] Changelog updated (or not applicable)


maltesander (approved):
LGTM.

teozkr:
Example PR for explicitly boxing: https://github.com/stackabletech/spark-operator/pull/192

PR #263, opened 2021-11-16
Change all instances of &Option<T> to Option<&T>

teozkr:
## Description

`Option<&T>` carries the same expressive power, is slightly more efficient, and
makes it easier to express some code with fewer clones.

In general code that provides provides these parameters will need to change from
`&foo` to `foo.as_ref()`, and from `&Some(foo)` to `Some(&foo)`. Rustc has a
specific lint for `as_ref`.

## Review Checklist


maltesander (approved):
LGTM.

PR #279, opened 2021-12-13
Refine CRD deprecations

teozkr:
## Review Checklist

### DIFF
--- a/src/crd.rs
+++ b/src/crd.rs
@@ -69,6 +69,11 @@ impl<T> CustomResourceExt for T where T: kube::CustomResourceExt {}
 /// - `client`: Client to connect to Kubernetes API and create the CRD with
 /// - `timeout`: If specified, retries creating the CRD for given `Duration`. If not specified,
 ///     retries indefinitely.
+#[deprecated(
+    since = "0.5.0",
+    note = "CRDs should be installed statically rather than by the operator itself"
+)]
+#[allow(deprecated)]
 pub async fn ensure_crd_created<T>(client: &Client) -> OperatorResult<()>
 where
     T: CustomResourceExt,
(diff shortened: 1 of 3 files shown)


src/crd.rs
```diff
@@ -69,6 +69,11 @@ impl<T> CustomResourceExt for T where T: kube::CustomResourceExt {}
 /// - `client`: Client to connect to Kubernetes API and create the CRD with
 /// - `timeout`: If specified, retries creating the CRD for given `Duration`. If not specified,
 ///     retries indefinitely.
+#[deprecated(
```
soenkeliebau, line 72: I think we actually have a ticket somewhere https://github.com/stackabletech/issues/issues/73  about this and wanted to make this configurable in the future..

soenkeliebau (commented):
Just one question..

src/crd.rs
teozkr: Wasn't that pre-k8s? That made some sense when we wanted to say "ok, install this Debian package to install and we'll install the k8s resources we need", less so to have to say "install these k8s resources and we'll create this final one for you".

soenkeliebau: that is true I guess.. but wouldn't it still be an option to just start an operator with `kubectl run ...` if you dislike or don't have Helm?

teozkr: You'd still need to do a bunch of setup in the background, just to ensure that the correct RBAC permissions are applied. The default ServiceAccount certainly won't have permission to create CRD objects in most clusters.

The "correct" solution there would be to `kubectl apply` the kustomize manifests (https://github.com/stackabletech/issues/issues/46).

soenkeliebau: ok, I'll shut up :)

soenkeliebau (approved):

PR #282, opened 2021-12-13
Default Role/RoleGroup configuration when deserializing

### LINKED ISSUE #281: Configuration::compute_* is not invoked if RoleGroup object doesn't define `config` field
This means that the following object works:

```yaml
apiVersion: imaginary.stackable.tech/v1alpha1
kind: ImaginaryCluster
metadata:
  name: imaginary-cluster
spec:
  role:
    roleGroup:
      config: {}
```

But the following might be missing crucial default values:

```yaml
apiVersion: imaginary.stackable.tech/v1alpha1
kind: ImaginaryCluster
metadata:
  name: imaginary-cluster
spec:
  role:
    roleGroup: {}
```

teozkr:
## Description
Fixes #281

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- BREAKING: `Configuration::compute_*` are now invoked even when `config` field is not provided on `Role`/`RoleGroup` ([#282]).
+  - `CommonConfiguration::config` is no longer `Option`al
+  - `Role::config` is no longer `Option`al
+  - `RoleGroup::config` is no longer `Option`al
+
+[#282]: https://github.com/stackabletech/operator-rs/pull/282
+
 ## [0.6.0] - 2021-12-13
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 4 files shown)


src/product_config_utils.rs
```diff
@@ -418,31 +413,17 @@ where
...
-
     // Properties from the role have the lowest priority, so they are computed and added first...
-    if let Some(CommonConfiguration {
-        config: Some(ref config),
-        ..
-    }) = config
-    {
-        final_properties = config.compute_cli(resource, role_name).unwrap();
-    }
+    let mut final_properties = config.config.compute_cli(resource, role_name).unwrap();
```
maltesander, line 422: It was in before, but i think we should not just unwrap here (and in parse_env_properties, parse_file_properties).
Though we have it under control via framework and operators, come to think of it maybe we want to throw an error if some config properties are wrong, bad combination etc. So i think we should handle this here or at least use unwrap_or_default()?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Changed
+- BREAKING: Introduced proper error handling for `parse_cli_properties`, `parse_env_properties` and `parse_file_properties` ([#282]).
```
teozkr, line 8: Isn't `transform_all_roles_to_config` the public part of the API that changed here?

maltesander: True, fixed.

teozkr:
LGTM

Time to give bors the go-ahead?

maltesander (approved):
LGTM.

PR #338, opened 2022-03-04
Role override regression

razvan:
## Description

This just demonstrates a bug in the handling of configuration overrides but doesn't fix it since the bug is not critical.

The bug is described in more detail here: https://github.com/stackabletech/operator-rs/issues/339

The fix might require some redesign of the way config overrides are handled and might not be worth doing in isolation.

Remove the `#[ignore]` attribute on the test to verify the problem.

## Review Checklist


soenkeliebau:
Closed for now since we expect to rework the product config bit in the near future.

PR #341, opened 2022-03-07
[Merged by Bors] - Apply operator templating where applicable

### LINKED ISSUE #323: Github Actions setup: Bring in line with operators
The operator-rs project is not using the operator-templating but there are a bunch of things in the operator templates that would make sense for operator-rs as well.

This is done when the two are in line (where it makes sense, e.g. docker build is not relevant here)

siegfriedweber:
## Description

* Apply [operator templating](https://github.com/stackabletech/operator-templating) where applicable
* Fix links in Rust documentation

Not applied:
* `.dockerignore`
* `.flake8`
* `.github/ISSUE_TEMPLATE/new_version.md`
* `.github/pull_request_template.md`: Helm chart check
* `.github/workflows/build.yml`: Regeneration of CRDs, Helm charts, and Kustomize manifests; Building and publishing of the Docker image
* `.github/workflows/reviewdog.yaml`: flake8, hadolint, shellcheck
* `.pylintrc`
* `deploy/`
* `docker/`
* `Makefile`
* `python/`
* `renovate.json`: lock file maintenance
* `scripts/`

Closes #323

## Review Checklist


deny.toml
```diff
@@ -16,10 +16,6 @@ notice = "warn"
 [bans]
 multiple-versions = "allow"
 
-[[bans.deny]]
```
teozkr, line 19: This should still be banned, and maybe upstreamed to operator-templating.

teozkr (approved):

deny.toml
siegfriedweber: The `time` crate is not used. Does it make sense to ban unused crates? If yes, why exactly `time`?

teozkr: operator-rs depends on `k8s-openapi`, which depends on `chrono`, which depends on `time` 0.1 unless everyone who depends on it is religious about using `default-features = false` (and nobody explicitly enables the feature `oldtime`). This has happened before, and we'd like to avoid it happening again in the future.

`time` 0.1 is unmaintained, and has a bunch of outstanding vulnerability advisories (such as https://rustsec.org/advisories/RUSTSEC-2020-0071.html), which cause false positive errors elsewhere even if we never use them ourselves.

> Does it make sense to ban unused crates?

Yes, bans exist specifically to avoid specific crates getting pulled in as transitive dependencies.

siegfriedweber:
bors merge

PR #342, opened 2022-03-07
[Merged by Bors] - Add support for secret-operator volumes to pod volume builder

teozkr:
## Review Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -6,9 +6,9 @@ use crate::error::{Error, OperatorResult};
 use crate::labels;
 use chrono::Utc;
 use k8s_openapi::api::core::v1::{
-    Affinity, Capabilities, ConfigMap, ConfigMapVolumeSource, Container, ContainerPort,
-    DownwardAPIVolumeSource, EmptyDirVolumeSource, EnvVar, EnvVarSource, Event, EventSource,
-    HostPathVolumeSource, Node, NodeAffinity, NodeSelector, NodeSelectorRequirement,
+    Affinity, CSIVolumeSource, Capabilities, ConfigMap, ConfigMapVolumeSource, Container,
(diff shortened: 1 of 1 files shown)


sbernauer:
Thanks a lot!
In the AuthenticationClass we get a user-provided scope from the user (see below). My understanding is, that the attribute `spec.protocol.ldap.bindCredentials.scope` is a string and is just passed to the Pod volume, is that correct?
In that case we would need a way to just pass in the `secrets.stackable.tech/scope` string and skip creating Rust vectors.

```
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: myldap
spec:
  protocol:
    ldap:
      hostname: ldap.server
      port: 389
      domain: domain.local
      bindCredentials:
        secretClass: superset-with-ldap-ldap-bind
        scope: Service
```

teozkr:
> My understanding is, that the attribute spec.protocol.ldap.bindCredentials.scope is a string and is just passed to the Pod volume, is that correct?

Sort of.. the question there would be how much we want to validate in the operator vs just passing it through. We might also want to force certain scopes when we know that a service will require them.. not sure there.

sbernauer:
Alright, i will think of a solution

sbernauer:
I will also add some unit test as I'm working on new features for `SecretOperatorVolumeSourceBuilder`.
So I'm going to check that in the Review Checklist

sbernauer (approved):

teozkr:
Kinda getting mixed signals here... :P

sbernauer:
I'm happy to merge this as is. I'm working on the "passing a scope string to SecretOperatorVolumeSourceBuilder" but will open a separate PR for that.

teozkr:
Alright.

bors r+

PR #343, opened 2022-03-08
Allow constructing SecretOperatorVolumeScope from user-provided string

sbernauer:
## Description
Follow up of https://github.com/stackabletech/operator-rs/pull/342
Now superset-operator can constuct SecretOperatorVolumeScope from user-provided string from AuthenticationClass CRD

## Review Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -22,6 +22,7 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::{
 use kube::{Resource, ResourceExt};
 use std::collections::BTreeMap;
 use std::fmt;
+use std::str::FromStr;
 use tracing::warn;
 
 /// A builder to build [`ConfigMap`] objects.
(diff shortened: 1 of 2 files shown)


teozkr:
Hm... I'm a bit inclined to think that I'd rather either like to express this in a more YAML/serde-native way immediately, something like:

```yaml
bindCredential:
  secretClass: foo
  scope:
  - pod: {}
  - service: bar
```

Or just upstream the scope decoding logic from the secret operator. The comma-separated list is "just" a consequence of the CSI stuff only allowing a string-to-string map for options, not because that's actually a good API when given some more freedom.

sbernauer:
Ok, so we then modify the AuthenticationClass CRD to something like this?
```
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: myldap
spec:
  protocol:
    ldap:
      hostname: ldap.server
      port: 389
      domain: domain.local
      bindCredentials:
        secretClass: superset-with-ldap-ldap-bind
        scope:
          node: {}
          pod: {}
          services:
            - a
            - b

Issue #356, opened 2022-03-16
Feature: OPA config fragment


fhennig:
We've decided (https://github.com/stackabletech/issues/issues/194) on a common configuration fragment for OPA integration for all operators, the fragment should look like:

    opa:
      configMapName: my-opa-cluster
      package: druid # optional; by default this will be the cluster name

there should also be functionality associated with this, i.e. assembling the final URL

PR #357, opened 2022-03-16
[Merged by Bors] - Added OpaConfig struct

maltesander:
## Description

closes https://github.com/stackabletech/operator-rs/issues/356

## Review Checklist

### DIFF
--- a/src/opa.rs
+++ b/src/opa.rs
@@ -8,7 +8,7 @@
 //! ```rust
 //! use serde::{Deserialize, Serialize};
 //! use stackable_operator::kube::CustomResource;
-//! use stackable_operator::opa::OpaConfig;
+//! use stackable_operator::opa::{OpaApiVersion, OpaConfig};
 //! use stackable_operator::schemars::{self, JsonSchema};
 //!
 //! #[derive(Clone, CustomResource, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
@@ -40,10 +40,8 @@
 //!
 //! let opa_config: &OpaConfig = cluster.spec.opa.as_ref().unwrap();
 //!
-//! assert_eq!(opa_config.package_url(&cluster), "v1/data/test".to_string());
-//! assert_eq!(opa_config.full_package_url(&cluster, "http://localhost:8081"), "http://localhost:8081/v1/data/test".to_string());
-//! assert_eq!(opa_config.rule_url(&cluster, Some("myrule")), "v1/data/test/myrule".to_string());
-//! assert_eq!(opa_config.full_rule_url(&cluster, "http://localhost:8081", Some("myrule")), "http://localhost:8081/v1/data/test/myrule".to_string());
+//! assert_eq!(opa_config.document_url(&cluster, Some("allow"), OpaApiVersion::V1), "v1/data/test/allow".to_string());
+//! assert_eq!(opa_config.full_document_url(&cluster, "http://localhost:8081", None, OpaApiVersion::V1), "http://localhost:8081/v1/data/test".to_string());
 //! ```
 use crate::client::Client;
 use crate::error;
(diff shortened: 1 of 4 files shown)


src/opa.rs
```diff
@@ -0,0 +1,245 @@
...
+    ///
+    /// This may be used if the OPA base url is contained in an ENV variable.
+    ///
+    /// # Example
+    ///
+    /// v1/data/<package>
+    ///
+    /// # Arguments
+    /// * `resource`     - The cluster resource.
+    pub fn package_url<T>(&self, resource: &T) -> String
```
razvan, line 75: Personally I'm not a big fan of this: having a (potentially) complex resource as a dependency (parameter) just to read one field from it is usually an anti pattern. Maybe we can introduce a `ClusterName` type or something like that.

maltesander: Its just something of type `ResourceExt` for the name() and namespace() methods. It defines the behavior as in the doc that we default to the resource name. That would not be true if we have a parameter `cluster_name`.

fhennig:
looks good to me; tests work, docs are nice etc.

But I'll let razvan get back to this

fhennig:
Also we need to wait for github actions to work again :smile:

src/opa.rs
```diff
@@ -0,0 +1,369 @@
...
+//! ```
+use crate::client::Client;
+use crate::error;
+use crate::error::OperatorResult;
+use k8s_openapi::api::core::v1::ConfigMap;
+use kube::ResourceExt;
+use schemars::{self, JsonSchema};
+use serde::{Deserialize, Serialize};
+
+const OPA_API: &str = "v1/data";
```
soenkeliebau, line 56: I apologize in advance for this comment, I know that I said in the call we are fine to hard code this, but if we do it here I believe we create an artificial requirement for all our authorizers to always use the same version of the data api.
I am perfectly fine to have this hard coded here as a default value (maybe with a different name like 'V1_DATA_API" or something similar that allows for easier addition of other apis later) the individual authorizers have to be able to override this. And by that I do not mean that the user should be able to configure this, but that the authorizer code should be able to do so, still hardcoded, just somewhere else :)

src/opa.rs
```diff
@@ -0,0 +1,369 @@
...
+    ///
+    /// This may be used if the OPA base url is contained in an ENV variable.
+    ///
+    /// # Example
+    ///
+    /// v1/data/<package>
+    ///
+    /// # Arguments
+    /// * `resource`     - The cluster resource.
+    pub fn package_url<T>(&self, resource: &T) -> String
```
soenkeliebau, line 77: nit: The name of this function does not refer to the api (data) or version of the api (v1) that is called at all. Not a severe issue, but if at some point in time we also want to support the query api or v2 of the data api we might end up with inconsistent naming. 
Comment goes for all other fns as well I guess..

src/opa.rs
maltesander: introduced `OpaApiVersion` enum to differentiate the api version and respective pathes

src/opa.rs
maltesander: renamed to document_url (which reflects the data api) and added the version enum.

src/opa.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// This may be used if the OPA base url is contained in an ENV variable.
+    ///
+    /// # Example
+    ///
+    /// * `v1/data/<package>`
+    /// * `v1/data/<package>/<rule>`
+    ///
+    /// # Arguments
+    /// * `resource`     - The cluster resource.
+    /// * `rule`         - The rule name. Defaults to `allow`.
```
soenkeliebau, line 89: This is not correct any more I think?

src/opa.rs
```diff
@@ -0,0 +1,310 @@
...
+#[derive(Clone, Debug, Default, Deserialize, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OpaConfig {
+    pub config_map_name: String,
+    pub package: Option<String>,
+}
+
+impl OpaConfig {
+    /// Returns the OPA data API url. If [`OpaConfig`] has no `package` set,
+    /// will default to the cluster `resource` name.
```
soenkeliebau, line 78: Should we also mention what happens if no rule is set?

soenkeliebau (changes requested):

soenkeliebau (approved):

maltesander:
bors merge

PR #365, opened 2022-04-06
Adds structs to allow configuring resource limits and storage parameters

soenkeliebau:
Signed-off-by: Sönke Liebau <soenke.liebau@stackable.de>

## Description

These structs are intended for use by operators to offer a consistent interface across the entire platform.

## Review Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,7 @@ pub mod namespace;
 pub mod opa;
 pub mod pod_utils;
 pub mod product_config_utils;
+mod resources;
 pub mod role_utils;
 pub mod utils;
 pub mod validation;
--- a/src/resources.rs
+++ b/src/resources.rs
@@ -1,0 +1,108 @@
+//! This module defines structs that can be referenced from operator CRDs to provide a common way of
+//! defining resource limits across the Stackable platform.
+//!
+//! The structs define shared limits for memory and cpu resources, with the option of specifying
+//! adding operator specific memory settings as well via a generic part of the struct.
+//! For this generic part of the struct an empty default has been defined, if no extra options are
+//! necessary.
+//!
+//! In addition to the mentioned limits the Resources struct is also generic over the storage
+//! configuration, which allows operators to offer custom configuration here as needed for the
+//! specific product.
+//! For persistent storage it is recommended to base the storage config on the [`PvcConfig`] struct.
+//!
+//! We expect to define common additional runtime limits in this module over time, at the moment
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -10,6 +10,7 @@ pub mod namespace;
 pub mod opa;
 pub mod pod_utils;
 pub mod product_config_utils;
+mod resources;
```
maltesander, line 13: should be pub

src/resources.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Struct that exposes the values for a PVC which the user should be able to influence
+#[derive(strum::Display, Clone, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PvcConfig {
+    capacity: String,
+    storageClass: Option<String>,
+    selectors: LabelSelector,
+}
```
maltesander, line 108: What about optional accessmodes (enum) and a default?

maltesander (commented):
Would it make sense to implement some basic helper methods like create pvc (from PvcConfig) etc.?

src/resources.rs
soenkeliebau: I think the accessmode is nothing the user should be able to influence too much. This should be within the purview of the operator whether it needs write access, read access from many, ...

soenkeliebau:
> Would it make sense to implement some basic helper methods like create pvc (from PvcConfig) etc.?

I have added a fn to do this. I considered the options a bit, ideally I'd like to `impl Into<VolumeBuilder> for PvcConfig`, but  name is required on the constructor of the builder so that doesn't work.

I think just having a fn that returns a builder might make more sense than returning a fully built volumeclaim?

@teozkr any thougths?

teozkr:
To be honest, the FooBuilder-Foo distinction is somewhat arbitrary anyway. I'd be fine with either, but I think returning FooBuilder makes sense.

soenkeliebau:
I have decided to leave it as is :)

The VolumeBuilder is at the volume level and expects a ready-made pvc, so lets return a pvc from here that can then be stuck into a volumebuilder if needed.

maltesander (approved):
LGTM.

sbernauer:
I have on small request: We introduced a new module `commons` where we put all the common structs (currently available on the main branch).
Could you move you're `resources.rs` there?

soenkeliebau:
bors cancel

sbernauer (approved):
LGTM % failing clippy check

Issue #362, opened 2022-04-06
Product-Config 2.0


nightkr:
The current product-config architecture has a few issues that we keep bumping into:

- It's confusing to understand what values come from where
- Validation error messages are produced in the context of the product's configuration format, which is confusing to users who configure it in term of our `ProductCluster` abstraction
- PC only works on unstructured key/value maps that correspond directly to the *product*'s configuration structure, making it difficult to feed config back from PC to the operator itself (for example: port values need to match between the configuration's listen option and the generated discovery configmaps)
  - This often leads us to bypass P-C for these options
  - We have no solid way to get back typed Rust values when using the `Configuration` mechanism to allow fields to be overridden across the `ProductCluster`/`Role`/`RoleGroup` hierarchy (without implementing the same precedence mechanism in the operator itself)
- There is a lot of repetitive boilerplate when linking Rust structs up to the PC machinery

To work this out I propose reorient PC so that the Rust structs are the source of truth for the structure, and so that precedence rules and validations are applied while still working with typed Rust values. We would still need to serialize to the product's native configuration format (and apply overrides to that) eventually, but this would be delayed until it is required.

From the definition side, the API is currently expected to look something like this:

```rust
#[derive(ProductConfig)]
struct ZookeeperConfig {
    #[pc(default = 1000)]
    #[pc(file("hdfs-site.xml", "tick.limit.ms"))]
    tick_limit_ms: i32,
}
```

This would expand to something like the following:

```rust
struct ZookeeperConfigFragment {
    tick_limit_ms: Option<i32>,
}

enum ZookeeperConfigValidationError {
    NoTickLimitMs,
}

impl ProductConfigFragment for ZookeeperConfigFragment {
    type Validated = ZookeeperConfig;
    type ValidationError = ZookeeperConfigValidationError;

    fn merge(self, other: &Self) -> Self {
        Self {
            tick_limit_ms: self.tick_limit_ms.or_else(other.tick_limit_ms),
            ...
        }
    }

    fn default() -> Self {
        Self {
            tick_limit_ms: Some(1000),
        }
    }

    fn validate(self) -> Result<ZookeeperConfig, ZookeeperConfigValidationError> {
        match self {
            ZookeeperConfigFragment { tick_limit_ms: Some(tick_limit_ms) } => Ok(ZookeeperConfig {tick_limit_ms}),
[...]

PR #368, opened 2022-04-08
[Merged by Bors] - Initial work on config merging

teozkr:
## Description

This is a starting point for #362, but is intended to be usable outside of that context (in this case, the primary motivation for getting this done now is #365).

## Work Checklist until R4R

- [X] Merging "primitive" types like `u8` or `String`
- [X] Merging arbitrarily nested structs
- [x] Merging enums
- [x] Merging boxed stuff
- [ ] Merging lists
  - This will probably depend on the lists in question, might want to defer this for a later PR
- [x] Docs

## Review Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     env:
       RUSTC_BOOTSTRAP: 1
     steps:
-      - uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # tag=v3.0.0
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
       - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # renovate: tag=v1.0.7
         with:
           profile: minimal
@@ -59,8 +59,8 @@ jobs:
     continue-on-error: ${{ matrix.checks == 'advisories' }}
 
     steps:
-      - uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # tag=v3.0.0
-      - uses: EmbarkStudios/cargo-deny-action@8acbae97b5d01b0481ae14cee8fcd8f5aa9e374d # tag=v1.2.12
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
+      - uses: EmbarkStudios/cargo-deny-action@ccfac4e084d1dedad50125be4550f87a21ba181e # tag=v1.2.15
         with:
           command: check ${{ matrix.checks }}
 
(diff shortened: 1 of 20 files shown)


teozkr:
We might want to reorganize this into a "virtual root crate + members" layout, but I'd rather split that into a separate PR to avoid cluttering the diff view too much.

maltesander (commented):
Im still on the code, just 2 findings to not get lost.
`stackable-operator-derive` needs to be added to exceptions in the deny.toml.

teozkr:
Hopefully cargo deny should be happy now...

maltesander (approved):
LGTM. Works fine for the usecases i could come up with.
As discussed with @teozkr, im fine with merging for now and see what we stumble across later on.

teozkr:
I've been experimenting a bit further with this in [feature/merge-experiments](https://github.com/stackabletech/operator-rs/tree/feature/merge-experiments), but never really got to a point that I was happy with there. In particular:

- cbb7f0b123e97bee39cc64f4363e88966d616553 changes our uses of `Option` into a new `Inheritable` enum to try to clarify the semantics we want a bit better. This leaves `Option` available for semantically optional fields. However, this has a bunch of downsides:
  - Serde doesn't like differentiating between missing and `null` fields, the closest I could get required sprinkling `#[serde(default)]` on every field, which is.. far from ideal
  - This is pretty much impossible to represent idiomatically in Go, which I suspect will cause problems for getting K8s to accept the schema definitions down the line (I didn't even try this yet at that point)
- 19bdb835eadb25631520cc5eff3e4b32fdef6de4 and 0bebbe7346c393cc6ffc5fa75e33997829ae20d9 further split up `Inheritable` into the `Atomic` and `Enum` enums, allowing any type (that conforms to the regular `Deserialize+Serialize+JsonSchema` requirements) to be used as an atomic, but arguably this is an even worse situation to be in because it further increases complexity without really fixing the root issues
- 7dc134c3b4a5a84bde7369a2052116b3a802e9a1 (the current state of the `feature/merge` branch that this PR concerns) overloads `Option` between "this is optional" and "this should be inherited", leaving the differentiation between those two cases for a future validation step.. this also has a few disadvantages
  - It's impossible to override a field to `null` once a value has been set
  - It's unclear from the struct itself whether a field should actually be optional

I'm starting to believe that "`Option` everywhere" is still the most viable approach though, and that we'll have to handle cases where we want to allow override-to-null separately. I think the "unclear optionality" problem should get better once we have the validation layer in place, at which point the mergeable fragments would likely be generated by the validation layer anyway.

soenkeliebau:
I have tested the code in here with my draft PR in NiFi and everything works as expected. I'd be happy to merge this.

However, I f****d up when trying to merge main into this PR earlier today and may have broken other code, not sure..

PR #396, opened 2022-05-06
[Merged by Bors] - Use ephemeral secret-op volumes rather than csi

teozkr:
## Description

Requires https://github.com/stackabletech/secret-operator/pull/125, see that PR for motivation

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- `SecretClassVolume` and `SecretOperatorVolumeSourceBuilder` now support secret-aware pod scheduling ([#396], [secret-#125]).
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


maltesander (approved):
LGTM. 
Edit: Is `Ephemeral(Box<EphemeralVolumeSource>)` in the enum really required? I mean we do not store 1000s of elements of that in a list. Whats against `#[allow(clippy::enum_large_variants)]` (not sure thats exactly how the lint is called)?

teozkr:
Yeah the main problem with `enum_large_variants` is that it inflates the size of that type regardless of which variant we use. If we don't particularly are about that then we can just use use it inline.

maltesander:
> Yeah the main problem with `enum_large_variants` is that it inflates the size of that type regardless of which variant we use, which might also make things like moves slower depending on the ABI that the compiler ends up deciding on. If we don't particularly are about that then we can just use use it inline.

Yeah but we do not really move that around or store a lot of elements in a list or sth. I think we can ignore the memory overhead for one item here?

sbernauer (approved):

PR #397, opened 2022-05-06
[Merged by Bors] - Added config::merge::chainable_merge()

razvan:


src/config/merge.rs
```diff
@@ -391,4 +418,24 @@ mod tests {
             BTreeMap::from([("a", Acc(4)), ("b", Acc(2)), ("c", Acc(5))])
         );
     }
+
+    #[test]
+    fn test_chainable_merge() {
```
teozkr, line 423: No need to keep the same test case as both a regular unit test and a doctest

teozkr (commented):
Did you upload your branch where you use this in the operator?

src/config/merge.rs
razvan: Here is a usecase: https://github.com/stackabletech/hdfs-operator/pull/164/files#diff-b10253ea28ef51076309f375e97cd5e2795500b0e07c5326c413055bffcb4a85R253

I will remove the test

maltesander (approved):
LGTM.

razvan:
bors merge

PR #398, opened 2022-05-06
[Merged by Bors] - Set fields to None instead of Some(empty map)

razvan:


teozkr:
What's the rationale here?

razvan:
Consistency I guess. If the operator defaults for these fields are None, this method will turn them into Some().

soenkeliebau:
ping @teozkr: are you happy with the reasoning here?

teozkr (approved):
Sorry, forgot about this PR.

I don't think it really matters since K8s should just deep-merge the requests objects anyway, but semantically I think this makes sense to say.

razvan:
bors merge

razvan:
bors merge

PR #400, opened 2022-05-10
[Merged by Bors] - Add Memory structs to convert K8S resource limits to Java heap options.

razvan:


src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+    Kibi,
+    Mebi,
+    Gibi,
+    Tebi,
+    Pebi,
+    Exbi,
+}
+
+impl BinaryMultiple {
+    pub fn to_legacy(&self) -> String {
```
teozkr, line 19:
```suggestion
    pub fn to_java_memory_unit(&self) -> String {
```

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+        match self {
+            BinaryMultiple::Kibi => "k".to_string(),
+            BinaryMultiple::Mebi => "m".to_string(),
+            BinaryMultiple::Gibi => "g".to_string(),
+            BinaryMultiple::Tebi => "t".to_string(),
+            BinaryMultiple::Pebi => "p".to_string(),
+            BinaryMultiple::Exbi => "e".to_string(),
```
teozkr, line 26: Java only supports up to Gibibytes.

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    pub fn upscale(&self) -> Self {
```
teozkr, line 30: What does this do? Why does `Kibi` transparently do nothing rather than fail?

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+            "ki" | "kib" => Ok(BinaryMultiple::Kibi),
+            "mi" | "mib" => Ok(BinaryMultiple::Mebi),
+            "gi" | "gib" => Ok(BinaryMultiple::Gibi),
+            "ti" | "tib" => Ok(BinaryMultiple::Tebi),
+            "pi" | "pib" => Ok(BinaryMultiple::Pebi),
+            "ei" | "eib" => Ok(BinaryMultiple::Exbi),
+            _ => Err(Error::InvalidQuantityUnit {
+                value: q.to_string(),
+            }),
+        }
```
teozkr, line 57: Capitalization matters, for example `mi` should be milli (or fail loudly). K8s quantities also generally leave the unit itself (`B` here) out.

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct Memory {
+    value: f32,
+    unit: BinaryMultiple,
+}
+
+impl Memory {
+    /// Scale by the given factor. If the factor is less then one
+    /// the unit granularity is increased one level to ensure eventual
+    /// conversions to Java heap settings don't end up with zero values..
```
teozkr, line 71: This scaling seems like something we'd want to do when rounding/formatting the value for output?

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn scale(&self, factor: f32) -> Self {
```
teozkr, line 72: This could be an `impl Mul<f32> for Memory`

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+    }
+
+    /// The Java heap settings do not support fractional values therefore
+    /// this cannot be implemented without loss of precision.
+    pub fn to_java_heap(&self, factor: f32) -> String {
+        let scaled = self.scale(factor);
+        format!("-Xmx{:.0}{}", scaled.value, scaled.unit.to_legacy())
+    }
```
teozkr, line 91: 1. Should the arbitrary scaling really be embedded into this call?
2. `Memory(2Gi).to_java_heap(0.8)` would be truncated down to 1Gi, which is a pretty far cry from the expected 1.6Gi. We should probably either hard-code it to output Mebibytes, or hard-code a certain number of significant digits rather than just ask whether it'd truncate down to zero.

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn from_str(q: &str) -> OperatorResult<Self> {
+        let mut v = String::from("");
+        let mut u = String::from("");
+
+        for c in q.chars() {
+            if c.is_numeric() || c == '.' {
+                v.push(c);
+            } else {
+                u.push(c);
+            }
```
teozkr, line 106: 1. This will allow you to continue reading numbers after the unit
2. No need to copy the quantity string here, you can just slice the `&str`.. for example:

```rust
let start_of_unit = q.find(|c| c != '.' && !c.is_numeric()).context(NoUnitSnafu)?;
let (value, unit) = q.split_at(start_of_unit);
```

src/memory.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+impl FromStr for BinaryMultiple {
+    type Err = Error;
+
+    fn from_str(q: &str) -> OperatorResult<BinaryMultiple> {
+        let lq = q.to_lowercase();
+        match lq.as_str() {
+            "ki" | "kib" => Ok(BinaryMultiple::Kibi),
```
soenkeliebau, line 48: I've done a bit of reading and experimenting and as far as I can tell K8s also accepts m | "" | k | M | G | T | P | E as suffix here, which are a different unit, but seem to be treated as equal by k8s at least for PV(C) sizing.
I think we can probably get away with just adding these as synonyms in this list and treating them as equal?

src/memory.rs
teozkr: I'd scale them.

src/memory.rs
teozkr: milli comes from https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/.. looks like the unit for that is `m` rather than `mi`. Still applies though that `1mB` is a very different thing from `1MB`.

src/memory.rs
teozkr: But saturating here doesn't make sense when we still multiply the amount by 1024 on the other end.

The more correct way then would be to have `BinaryAmount::upscale()` fail and the full unit-and-value scaler react to that by saturating and swallowing the error.

src/memory.rs
teozkr: > It's a convenience since it's what we always have to do when computing the max Java heap.

`memory.to_java_heap(0.8)` doesn't save much typing compared to `(memory * 0.8).to_java_heap()`, and is (IMO) much less obvious about what's going on (especially without reading the definition of `to_java_heap` again).

> 2. It gets mapped to -Xmx1638m which assuming K8S does the right thing, is what we want.

Ok, then I'll need to read through this again because I don't understand where this is happening.

teozkr: Ah, you have the threshold where you try to avoid wasting more than 0.2x the base unit. I'd still prefer to always render to megabytes.. but meh.

src/memory.rs
```diff
@@ -6,7 +6,7 @@
 //! - units are case sensitive ("2gi" is invalid)
 //! - exponential notation is not supported.
 //!
-//! For more info in Kubernetes qunatities see: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go
+//! For more info in Kubernetes qunatities see: <https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go>
```
teozkr, line 9: `qunatities`

teozkr (approved):

src/memory.rs
razvan: oh man...

razvan:
bors merge

PR #401, opened 2022-05-10
[Merged by Bors] - Add pathStyleAccess to S3ConnectionSpec

siegfriedweber:
## Description

`pathStyleAccess` added to `commons::s3::S3ConnectionSpec`

## Review Checklist


sbernauer:
I think the main question is: What is the default value? ;)
Pros of default to true:
* It should work anywhere (vs. lots of users having problems and need to figure out their S3 doesn't support host style access)
Pros of default to false:
* Maybe better performance for AWS S3? Or lower costs from AWS S3?

So i guess we should default to true for maximum portability

sbernauer:
And i still can't get around that the bool is optional and not just simply defaulting to true (or false).
But i guess that's the way it is :D

maltesander:
> I think the main question is: What is the default value? ;) Pros of default to true:
> 
> * It should work anywhere (vs. lots of users having problems and need to figure out their S3 doesn't support host style access)
>   Pros of default to false:
> * Maybe better performance for AWS S3? Or lower costs from AWS S3?
> 
> So i guess we should default to true for maximum portability

I think pathstyle access will be deprecated at some point? Not sure for every provider, at least for AWS. So for now i would default to true, but still use it in every example and set it explicitly.

sbernauer:
Sounds good :+1: 
The Option actually is really bad IMHO here as we cannot simply set the default to true but instead every individual operator has to enforce the default by itself. Is there any chance of using a bool and default to true?

maltesander:
> Sounds good +1 The Option actually is really bad IMHO here as we cannot simply set the default to true but instead every individual operator has to enforce the default by itself. Is there any chance of using a bool and default to true?

I think it makes sense that each operator does it on its own. When setting that you can just have an unwrap_or() with the default best suited for the operator.

maltesander (approved):
LGTM.

maltesander:
bors merge

sbernauer (approved):

sbernauer:
bors cancel

maltesander:
bors merge

maltesander:
bors merge

PR #436, opened 2022-07-12
[Merged by Bors] - Orphaned resources handling

siegfriedweber:
## Description

Provides a structure where cluster resources can be added.

When calling the `update` function, all provided resources will be patched and the deployed resources which are not provided will be deleted.

This is work in progress! The implementation is not suitable for all operators and will be changed.

The implementation is already tested in:
* stackabletech/hbase-operator#215
  Integration tests: [Build Status](https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hbase-operator-it-custom/18/)
* https://github.com/stackabletech/airflow-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/druid-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/hdfs-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/hive-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/kafka-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/nifi-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/opa-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/superset-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/trino-operator/tree/orphaned_resources_handling
* https://github.com/stackabletech/zookeeper-operator/tree/orphaned_resources_handling

@adwk67 will have a look at spark-k8s-operator if there is anything to do.

## Review Checklist

### DIFF
--- a/src/cluster_resources.rs
+++ b/src/cluster_resources.rs
@@ -18,8 +18,9 @@ use crate::{
     kube::{Resource, ResourceExt},
     labels::{APP_INSTANCE_LABEL, APP_MANAGED_BY_LABEL, APP_NAME_LABEL},
 };
+use kube::core::ErrorResponse;
 use serde::{de::DeserializeOwned, Serialize};
-use tracing::info;
+use tracing::{debug, info};
 
 /// A cluster resource handled by `ClusterResources`.
 pub trait ClusterResource:
@@ -269,6 +270,10 @@ impl ClusterResources {
     /// Deletes all deployed resources which are labelled as if they belong to this cluster
     /// instance but are not contained in the given list.
     ///
+    /// If it is forbidden to list the resources of the given kind then it is assumed that the
+    /// caller is not in charge of these resources, the deletion is skipped, and no error is
+    /// returned.
+    ///
     /// # Arguments
     ///
     /// * `client` - The client which is used to access Kubernetes
@@ -276,25 +281,38 @@ impl ClusterResources {
         &self,
         client: &Client,
     ) -> OperatorResult<()> {
-        let deployed_cluster_resources = self.list_deployed_cluster_resources::<T>(client).await?;
+        match self.list_deployed_cluster_resources::<T>(client).await {
+            Ok(deployed_cluster_resources) => {
+                let orphaned_resources = deployed_cluster_resources
+                    .into_iter()
+                    .filter(|r| !self.resources.contains(&r.into()))
+                    .collect::<Vec<_>>();
 
-        let orphaned_resources = deployed_cluster_resources
-            .into_iter()
-            .filter(|r| !self.resources.contains(&r.into()))
-            .collect::<Vec<_>>();
+                if !orphaned_resources.is_empty() {
+                    info!(
+                        "Deleting orphaned {}: {}",
+                        T::plural(&()),
+                        ClusterResources::print_resources(&orphaned_resources),
+                    );
+                    for resource in orphaned_resources.iter() {
+                        client.delete(resource).await?;
+                    }
+                }
 
-        if !orphaned_resources.is_empty() {
-            info!(
-                "Deleting orphaned {}: {}",
-                T::plural(&()),
-                ClusterResources::print_resources(&orphaned_resources),
-            );
-            for resource in orphaned_resources.iter() {
-                client.delete(resource).await?;
+                Ok(())
             }
+            Err(Error::KubeError {
+                source: kube::Error::Api(ErrorResponse { code: 403, .. }),
+            }) => {
+                debug!(
+                    "Skipping deletion of orphaned {} because the operator is not allowed to list \
+                      them and is therefore probably not in charge of them.",
+                    T::plural(&())
+                );
+                Ok(())
+            }
+            Err(error) => Err(error),
         }
-
-        Ok(())
     }
 
     /// Creates a string containing the names and if present namespaces of the given resources
--- a/src/cluster_resources.rs
+++ b/src/cluster_resources.rs
@@ -21,6 +21,17 @@ use crate::{
 use serde::{de::DeserializeOwned, Serialize};
 use tracing::info;
 
+/// A cluster resource handled by `ClusterResources`.
+pub trait ClusterResource:
+    Clone + Debug + DeserializeOwned + Resource<DynamicType = ()> + Serialize
+{
+}
+
+impl ClusterResource for ConfigMap {}
+impl ClusterResource for DaemonSet {}
+impl ClusterResource for Service {}
+impl ClusterResource for StatefulSet {}
+
 /// A structure containing the cluster resources.
 ///
 /// Cluster resources can be added and on finalizing, orphaned resources are deleted. A cluster
@@ -180,10 +191,11 @@ impl ClusterResources {
     /// * `app.kubernetes.io/instance = <cluster.name>`
     /// * `app.kubernetes.io/managed-by = <app_name>-operator`
     /// * `app.kubernetes.io/name = <app_name>`
-    pub async fn add<T>(&mut self, client: &Client, resource: &T) -> OperatorResult<T>
-    where
-        T: Clone + Debug + DeserializeOwned + Resource<DynamicType = ()> + Serialize,
-    {
+    pub async fn add<T: ClusterResource>(
+        &mut self,
+        client: &Client,
+        resource: &T,
+    ) -> OperatorResult<T> {
         ClusterResources::check_label(resource.labels(), APP_INSTANCE_LABEL, &self.app_instance)?;
         ClusterResources::check_label(resource.labels(), APP_MANAGED_BY_LABEL, &self.manager)?;
         ClusterResources::check_label(resource.labels(), APP_NAME_LABEL, &self.app_name)?;
@@ -236,12 +248,6 @@ impl ClusterResources {
     /// All orphaned resources, i.e. resources which are labelled as if they belong to this cluster
     /// instance but were not added to the cluster resources, are deleted.
     ///
-    /// The following resource types are considered:
-    /// * `ConfigMap`
-    /// * `DaemonSet`
-    /// * `Service`
-    /// * `StatefulSet`
-    ///
     /// The following resource labels are compared:
     /// * `app.kubernetes.io/instance`
     /// * `app.kubernetes.io/managed-by`
@@ -266,10 +272,10 @@ impl ClusterResources {
     /// # Arguments
     ///
     /// * `client` - The client which is used to access Kubernetes
-    async fn delete_orphaned_resources<T>(&self, client: &Client) -> OperatorResult<()>
-    where
-        T: Clone + Debug + DeserializeOwned + Resource<DynamicType = ()>,
-    {
+    async fn delete_orphaned_resources<T: ClusterResource>(
+        &self,
+        client: &Client,
+    ) -> OperatorResult<()> {
         let deployed_cluster_resources = self.list_deployed_cluster_resources::<T>(client).await?;
 
         let orphaned_resources = deployed_cluster_resources
@@ -293,10 +299,7 @@ impl ClusterResources {
 
     /// Creates a string containing the names and if present namespaces of the given resources
     /// sorted by name and separated with commas.
-    fn print_resources<T>(resources: &[T]) -> String
-    where
-        T: Resource<DynamicType = ()>,
-    {
+    fn print_resources<T: ClusterResource>(resources: &[T]) -> String {
         let mut output = resources
             .iter()
             .map(ClusterResources::print_resource)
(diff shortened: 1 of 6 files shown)


src/cluster_resources.rs
```diff
@@ -0,0 +1,393 @@
...
+    ///
+    /// The following resource labels are compared:
+    /// * `app.kubernetes.io/instance`
+    /// * `app.kubernetes.io/managed-by`
+    /// * `app.kubernetes.io/name`
+    ///
+    /// # Arguments
+    ///
+    /// * `client` - The client which is used to access Kubernetes
+    pub async fn finalize(self, client: &Client) -> OperatorResult<()> {
```
soenkeliebau, line 260: Just a comment, not sure if this happened to others as well: Before actually reading the fn I kept thinking this was something to do with finalizers and was surprised that this is not the case. In any other context the name fits I think, but in k8s it migth already be overloaded?

src/cluster_resources.rs
```diff
@@ -0,0 +1,393 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use tracing::{debug, info};
+
+/// A cluster resource handled by `ClusterResources`.
+pub trait ClusterResource:
+    Clone + Debug + DeserializeOwned + Resource<DynamicType = ()> + Serialize
+{
+}
+
+impl ClusterResource for ConfigMap {}
```
soenkeliebau, line 31: Can we maybe add a comment here about this mainly being a marker trait, but that additional implementation is required in the finalize fn if more structs are added here, just to be on the safe side?

src/cluster_resources.rs
```diff
@@ -0,0 +1,393 @@
...
+        self.delete_orphaned_resources::<DaemonSet>(client).await?;
+        self.delete_orphaned_resources::<ConfigMap>(client).await?;
+
+        Ok(())
+    }
+
+    /// Deletes all deployed resources which are labelled as if they belong to this cluster
+    /// instance but are not contained in the given list.
+    ///
+    /// If it is forbidden to list the resources of the given kind then it is assumed that the
```
soenkeliebau, line 273: Would it maybe be worthwhile to change the signature of this fn to return a OperatorResult<bool> or something similar, which then shows if resources were present that nothing could be done about?
Not sure what scenarios this might apply to, but if an operator later needs this functionality it might be easy to add now, but have a larger blast radius later.

src/cluster_resources.rs
```diff
@@ -0,0 +1,393 @@
...
+        }
+    }
+
+    /// Lists the deployed resources with instance, name, and managed-by labels equal to this
+    /// cluster instance.
+    ///
+    /// # Arguments
+    ///
+    /// * `client` - The client which is used to access Kubernetes
+    async fn list_deployed_cluster_resources<T: ClusterResource>(
```
soenkeliebau, line 344: Should this maybe be 'pub'?

src/cluster_resources.rs
```diff
@@ -0,0 +1,393 @@
...
+        let mut output = resources
+            .iter()
+            .map(ClusterResources::print_resource)
+            .collect::<Vec<_>>();
+        output.sort();
+        output.join(", ")
+    }
+
+    /// Creates a string containing the name and if present namespace of the given resource.
+    fn print_resource<T: ClusterResource>(resource: &T) -> String {
```
soenkeliebau, line 330: Should this maybe be 'pub'? Or, perhaps a debug impl. that just calls this fn ?

src/cluster_resources.rs
```diff
@@ -0,0 +1,393 @@
...
+                        client.delete(resource).await?;
+                    }
+                }
+
+                Ok(())
+            }
+            Err(Error::KubeError {
+                source: kube::Error::Api(ErrorResponse { code: 403, .. }),
+            }) => {
+                debug!(
```
soenkeliebau, line 307: Should this maybe be at the warn level? Usually this will mean something is messed up, no?

soenkeliebau (commented):
Looks good to me overall, just left a few questions that came to mind while looking through this.

src/cluster_resources.rs
siegfriedweber: I see three options to reduce the confusion:

1. Use a synonym of `finalize`, like `close` or `complete`, to underline that this is the terminal operation of `ClusterResources`.
2. Use a name describing what the function is actually doing, like `delete_orphaned_resources` or `delete_remaining_cluster_resources`.
3. Convert this function into a `Drop` trait so that it is implicitly called. This could be difficult to achieve with the asynchronous context and it increases the magic behind the scenes.

Which option do you prefer?

src/cluster_resources.rs
siegfriedweber: Comment added. Please have a look!

src/cluster_resources.rs
siegfriedweber: If a resource cannot be deleted then an error is returned. The function terminates after the first error. So if a list of resources, resource names, or errors should be returned then the `Error` enum must be extended accordingly, e.g.:

```rust
pub enum Error {
    #[error("Errors occurred: {errors:?}")]
    Errors { errors: Vec<Error> },
}
```

The `delete_orphaned_resources` function must then collect all failed resources, resource names, or results, e.g.:

```rust
let mut results = Vec::new();
for resource in orphaned_resources {
    let result = client.ensure_deleted(resource).await;
    results.push(result);
}
let errors = results
    .into_iter()
    .filter_map(Result::err)
    .collect::<Vec<_>>();
if errors.is_empty() {
    Ok(())
} else {
    Err(Error::Errors { errors })
}
```

Is this what you want to achieve?

src/cluster_resources.rs
siegfriedweber: This function lists all resources which are labelled if they belong to this cluster instance. This list can contain orphaned resources which were not added to `ClusterResources` before. It would confuse the user of the `ClusterResources` struct if this function would be public:

```rust
let mut cluster_resources = ClusterResources::new(...);

// deploys resource1
cluster_resources.add(client, resource1).await?;

// returns resource1 and resource2
let deployed_cluster_resources = cluster_resources.list_deployed_cluster_resources(client).await?;

// Why do the cluster_resources differ from the deployed_cluster_resources?
```

src/cluster_resources.rs
siegfriedweber: > Should this maybe be 'pub'?

This is just a helper function. If its functionality is of public interest then it should be located in the implementation of a `Resource` or `ResourceExt` struct. I would not make arbitrary functions public because then they cannot be changed again without breaking the clients of this public API. Therefore I would keep the public interface as small as necessary to increase the evolvability.

> Or, perhaps a debug impl. that just calls this fn ?

Traits can only be implemented on trait objects. `Resource` cannot be made into an object because it contains associated functions without a `self` parameter, like `kind`. See https://doc.rust-lang.org/reference/items/traits.html#object-safety

src/cluster_resources.rs
siegfriedweber: This case occurs for all resources which are not used by the operator. Most operators do not use DaemonSets and then this "error" occurs which can be ignored. The log message is shown on every `reconcile` call, so it should have a log level below the default one which is probably "info".

The error could be avoided if the user would have an option to enter the used resource kinds in `ClusterResources::new` but this is cumbersome and error-prone if the operator is extended.

src/cluster_resources.rs
soenkeliebau: I think I'm in favor of 2.

Creating resources is done upon adding them to the struct, not when calling this fn, correct?

src/cluster_resources.rs
soenkeliebau: Looks good, thanks!

src/cluster_resources.rs
soenkeliebau: I think I misread your comment slightly..

> If it is forbidden to list the resources of the given kind then it is assumed that the caller is not in charge of these resources, the deletion is skipped, and no error is returned.

My thinking was that it simply skips some of the resources that it cannot delete, but it would simply skip the entire step when it is not allowed to list, which makes sense.
I was just wondering if the information that deletion was skipped might be interesting to the calling code at some point in time?

src/cluster_resources.rs
soenkeliebau: ok

src/cluster_resources.rs
soenkeliebau: ok

soenkeliebau: Actually I have spent some more thought on this..

I am currently in need of a fn sort of like this one for the NiFi operator. The scenario is an update of the NiFi version, which requires a full restart of the cluster. So the reconciliation checks fairly early on, whether the version is different and then scales all deployed statefulsets to 0.
For this I'd like to list the deployed statefulsets for this cluster and iterate over these, which could be done one of two ways: 
- list the resources in k8s
- generate the list of statefulsets in the regular reconciliation 

I think there is no right or wrong way here, it is more a question of "how do we want to treat concurrent edits of the NifiCluster resource" - so basically someone changes the version and scales the cluster in the same edit of the NifiCluster object.

It would probably make more sense to do the scaling first to avoid upgrading nodes just to then delete them, but from an implementation point of view I think doing the upgrade first might be easier. And then it would actually make sense to be able to list the owned statefulsets for this cluster.

sbernauer:
It would be great if you could also add Support for Secrets, as the trino-operator creates a Secret used as a shared secret https://trino.io/docs/current/security/internal-communication.html#configure-shared-secret

sbernauer:
Otherwise LGTM

src/cluster_resources.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// The following resource labels are compared:
+    /// * `app.kubernetes.io/instance`
+    /// * `app.kubernetes.io/managed-by`
+    /// * `app.kubernetes.io/name`
+    ///
+    /// # Arguments
+    ///
+    /// * `client` - The client which is used to access Kubernetes
+    pub async fn finalize(self, client: &Client) -> OperatorResult<()> {
+        self.delete_orphaned_resources::<Service>(client).await?;
```
Pscheidl, line 279: There's no need to wait for each deletion sequentially, `tokio::join!` can be used to execute deletion of all the resources concurrently.

src/cluster_resources.rs
```diff
@@ -0,0 +1,404 @@
...
+        self.delete_orphaned_resources::<StatefulSet>(client)
+            .await?;
+        self.delete_orphaned_resources::<DaemonSet>(client).await?;
+        self.delete_orphaned_resources::<ConfigMap>(client).await?;
```
Pscheidl, line 283:
```suggestion
        tokio::try_join!(
            self.delete_orphaned_resources::<Service>(client),
            self.delete_orphaned_resources::<StatefulSet>(client),
            self.delete_orphaned_resources::<DaemonSet>(client),
            self.delete_orphaned_resources::<ConfigMap>(client)
        )?;
```

src/cluster_resources.rs
siegfriedweber: I renamed the function to `delete_orphaned_resources`.

> Creating resources is done upon adding them to the struct, not when calling this fn, correct?

Correct. This is necessary because some operators use the data from the patched resources to create new resources, e.g. when creating discovery maps.

src/cluster_resources.rs
siegfriedweber: Adapted. Thank you!

siegfriedweber:
> It would be great if you could also add Support for Secrets, as the trino-operator creates a Secret used as a shared secret https://trino.io/docs/current/security/internal-communication.html#configure-shared-secret

If I understood the source code of the Trino operator correctly then there is one shared secret per Trino cluster. This secret has an owner reference to the cluster and is removed when the cluster is deleted. I do not see how this secret could become orphaned.

src/cluster_resources.rs
siegfriedweber: > I was just wondering if the information that deletion was skipped might be interesting to the calling code at some point in time?

The `delete_orphaned_resouces` function tries to delete resources of the kinds which implement the `ClusterResource` trait. Currently there are four kinds which implement this trait. If an operator would treat skipped kinds and another kind is added because it is used in a different operator then this could break the operator trying to handle skipped kinds. From the perspective of one operator, the kinds handled by the `ClusterResources`, are arbitrary and therefore the operator should not handle them.

src/cluster_resources.rs
siegfriedweber: I looked at the `feat/update` branch of the NiFi operator and several questions arose. Perhaps we should have a meeting and present only the outcome in this discussion thread?

src/cluster_resources.rs
soenkeliebau: makes sense

soenkeliebau (approved):

sbernauer:
> If I understood the source code of the Trino operator correctly then there is one shared secret per Trino cluster. This secret has an owner reference to the cluster and is removed when the cluster is deleted. I do not see how this secret could become orphaned.

You're absolutely right!

siegfriedweber:
bors merge

PR #448, opened 2022-08-01
Add wrapper for serde_yaml

siegfriedweber:
## Description

* Add wrapper for `serde_yaml` which adheres to the YAML specification. Operators should use the `yaml` module instead of `serde_yaml`.
* Update `serde_yaml` to version `0.9`.

This pull request solves the problem that `serde_yaml` in version `0.9` does not emit leading dashes anymore (see dtolnay/serde-yaml#266) which does not conform to the YAML specification (see https://yaml.org/spec/1.2.2/#22-structures). This becomes especially a problem when concatenating YAMLs which is done in many operators as follows:

```
Command::Crd => println!(
    "{}{}{}",
    serde_yaml::to_string(&SupersetCluster::crd())?,
    serde_yaml::to_string(&SupersetDB::crd())?,
    serde_yaml::to_string(&DruidConnection::crd())?
),
```

However, I did not move the concatenation to operator-rs, so it must still be done in the operator:

```
Command::Crd => {
    let yamls = vec![
        yaml::to_string(&SupersetCluster::crd())?,
        yaml::to_string(&SupersetDB::crd())?,
        yaml::to_string(&DruidConnection::crd())?,
    ];
    println!("{}", yamls.concat());
}
```

It can be discussed if this step should also be moved into operator-rs.

According to the usage in the operators (https://github.com/search?l=&q=org%3Astackabletech+%22serde_yaml%3A%3A%22+language%3ARust&type=code), it should be sufficient to forward only `Error`, `from_str`, and `to_string` from `serde_yaml`.

The implementation of `yaml::to_string` could be improved to use less space but with increased code complexity.

## Review Checklist

### DIFF
--- a/src/yaml.rs
+++ b/src/yaml.rs
@@ -1,0 +1,65 @@
+//! Wrapper around serde_yaml which adheres to the YAML specification
+//!
+//! Operators should use this module instead of serde_yaml.
+use serde::ser;
+
+use crate::error::OperatorResult;
+
+pub use serde_yaml::from_str;
+pub use serde_yaml::Error;
+
+/// Serialize the given data structure as a String of YAML.
+///
+/// Serialization can fail if `T`'s implementation of `Serialize` decides to
+/// return an error.
+pub fn to_string<T>(value: &T) -> OperatorResult<String>
+where
+    T: ?Sized + ser::Serialize,
+{
+    let yaml = serde_yaml::to_string(value)?;
+    Ok(format!("---\n{}", yaml))
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::BTreeMap;
+
+    use super::*;
+
+    #[test]
+    fn yaml_with_leading_dashes_can_be_deserialized() {
+        let yaml = "\
+            ---\n\
+            key: value";
+
+        let actual_value: BTreeMap<_, _> = from_str(yaml).expect("deserializable value");
+
+        let expected_value: BTreeMap<_, _> = [("key", "value")].into();
+
+        assert_eq!(expected_value, actual_value);
+    }
+
+    #[test]
+    fn yaml_without_leading_dashes_can_be_deserialized() {
+        let yaml = "key: value";
+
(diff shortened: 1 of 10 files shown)


teozkr:
Oh, I would have called it `serialize_crds` or something, since that's the actual use-case we're trying to solve here.

Otherwise a more descriptive name might be `documents_to_string` or something? `to_string` doesn't really make it clear to me what the difference is compared to `serde_yaml::to_string`.

src/yaml.rs
```diff
@@ -0,0 +1,65 @@
+//! Wrapper around serde_yaml which adheres to the YAML specification
```
soenkeliebau, line 1: I think strictly speaking the --- are only mandatory if the file starts with directives, to delimit these from the actual content of the file.
So arguably serde_yaml adheres to the spec, even though I have to say that I do not agree with the upstream change.
Besides that, it might be worthwhile just adding one sentence about what exactly the difference is between this and serde_yaml to make it easier to understand.

soenkeliebau (approved):
Code looks good, ran tests, looks good to me overall.

Not sure I agree with the upstreams change, but it is what it is :)

src/yaml.rs
siegfriedweber: In YAML 1.0 and 1.1, a YAML stream can start with an "implicit document" (without dashes) or an "exlicit document" (with dashes), see https://yaml.org/spec/1.0/index.html#id2561559 and https://yaml.org/spec/1.1/current.html#id898785.

It is the same in YAML 1.2 but "implicit documents" are called "bare documents", see https://yaml.org/spec/1.2.2/#913-bare-documents.

So serde_yaml is compliant with the YAML specification. I will change my pull request.

soenkeliebau: Is it worth opening an issue against yamllint for this?

siegfriedweber: There is already a closed issue with this topic: adrienverge/yamllint#302

siegfriedweber:
Pull request closed in favor of #450

PR #450, opened 2022-08-02
[Merged by Bors] - Ensure that explicit YAML documents are generated for CRDs

siegfriedweber:
## Description

* YAML module added with a function to serialize a data structure as an explicit YAML document. The YAML documents generated by the functions in `crd::CustomResourceExt` are now explicit documents and can be safely concatenated to produce a YAML stream.
* `serde_yaml` 0.8.26 -> 0.9.9

CRDs can now safely be concatenated by using `CustomResourceExt::print_yaml_schema` in a row:

```diff
-//! use kube::CustomResourceExt;
+//! use stackable_operator::CustomResourceExt;

 //! let opts = Opts::from_args();
 //!
 //! match opts.command {
-//!     cli::Command::Crd => println!(
-//!         "{}{}",
-//!         serde_yaml::to_string(&FooCluster::crd())?,
-//!         serde_yaml::to_string(&BarCluster::crd())?,
-//!     ),
+//!     cli::Command::Crd => {
+//!         FooCluster::print_yaml_schema()?;
+//!         BarCluster::print_yaml_schema()?;
+//!     },
 //!     cli::Command::Run { .. } => {
 //!         // Run the operator
 //!     }
```

**Hint:** If this was the only usage of `serde_yaml` in an operator then it can be removed from the dependencies.

## Review Checklist


src/commons/s3.rs
```diff
@@ -235,14 +236,13 @@ mod test {
...
-            serde_yaml::to_string(&bucket).unwrap(),
+            yaml::to_explicit_document_string(&bucket).unwrap(),
             "---
 bucketName: test-bucket-name
-connection:
-  inline:
-    host: host
-    port: 8080
-    accessStyle: VirtualHosted
+connection: !inline
```
teozkr, line 242: Hm, does this come from serde_yaml? We might need to go back to the old tagging format somehow…

teozkr: https://github.com/dtolnay/serde-yaml/issues/298

maltesander:
Does it make sense to simply reexport serde_yaml here as well to have consistent versions?
In the OPA PR https://github.com/stackabletech/opa-operator/pull/326 i think we have a problem with different versions (e.g. 0.8.25 in the operator-rs and 0.9 in OPA). The operator-rs has a YamlSerilizationError with source on the serde_yaml error which changed from 0.8 to 0.9.

teozkr:
@maltesander I believe that usage falls away anyway, if you use the new helper? So you should be able to remove the `serde_yaml` dependency entirely.

maltesander:
> @maltesander I believe that usage falls away anyway, if you use the new helper? So you should be able to remove the `serde_yaml` dependency entirely.

Ok (did not check the PR), thats fine with me then.

src/yaml.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Serialization can fail if `T`'s implementation of `Serialize` decides to return an error.
+///
+/// An [`Error::UnsupportedYamlDocumentError`] is returned if the used `serde_yaml` version
+/// generates a directive document.
+pub fn to_explicit_document_string<T>(value: &T) -> OperatorResult<String>
+where
+    T: ?Sized + ser::Serialize,
+{
+    // The returned document type depends on the serde_yaml version.
+    let document = serde_yaml::to_string(value)?;
```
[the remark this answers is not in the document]
siegfriedweber: So, you would abort if the returned YAML document type is not the expected one according to the `serde_yaml` version? For instance:

```
assert!(!document.starts_with("---") && !document.starts_with('%'));
```

I did it this way because in my opinion the "contract" with `serde_yaml` is not defined by its version and changelog but by its documentation. The `to_string` function returns "a String of YAML" according to its documentation and that has not changed over the versions. So I can expect a valid YAML document of any document type and therefore I have to check the type and do the necessary conversion. Unfortunately the directive document case is too complex that I am willing to implement it.

But I see that the reduced complexity also has its value. So I would change it if my comment did not convince you.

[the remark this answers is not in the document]
teozkr: I'd put that check into a `#[test]` so that we detect it when upgrading.

The change is listed as a breaking in `serde_yaml`'s [changelog](https://github.com/dtolnay/serde-yaml/releases/tag/0.9.0), implying that any future change would also have to be semver-major (and thus has to be explicitly approved and tested for operator-rs).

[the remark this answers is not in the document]
siegfriedweber: Document type check removed. If `serde_yaml::to_string` does not return the expected type then the already existing unit test fails (tested with `serde_yaml:0.8`):

```
thread 'yaml::tests::value_can_be_serialized_to_an_explicit_document_string' panicked at 'assertion failed: `(left == right)`
  left: `"---\nkey: value\n"`,
 right: `"---\n---\nkey: value\n"`', src/yaml.rs:37:9
```

src/commons/s3.rs
siegfriedweber: The proposed solution in `serde_yaml:0.9.4` is to annotate every enum field with `#[serde(with = "serde_yaml::with::singleton_map")]`.

But this solution does not work together with `JsonSchema`, see GREsau/schemars#89.

One workaround is to use the serde parameters `serialize_with` and `deserialize_with`:

```
#[serde(
    default,
    skip_serializing_if = "Option::is_none",
    serialize_with = "serde_yaml::with::singleton_map::serialize",
    deserialize_with = "serde_yaml::with::singleton_map::deserialize"
)]
pub connection: Option<S3ConnectionDef>,
```

This becomes quite verbose.

In which cases do the YAML tags actually pose a problem?

siegfriedweber:
We assume that the changes in serde_yaml, especially the generated YAML tags, have a high impact on a lot of projects and that there will adaptations in one of the next serde_yaml releases. So we decided to postpone this pull request until the next release and then reevaluate the situation.

src/commons/s3.rs
siegfriedweber: `serde_yaml:0.9.9` introduced [singleton_map_recursive](https://docs.rs/serde_yaml/0.9.9/serde_yaml/with/singleton_map_recursive/index.html) which allows to serialize enums as YAML maps containing one entry in which the key identifies the variant name.

This is now used in `stackable_operator::yaml::serialize_to_explicit_document`. Further annotations are not necessary in the operators.

teozkr (approved):
Nice!

siegfriedweber:
bors merge

Issue #470, opened 2022-09-20
Add product image selection struct


vsupalov:
Implementation of design described in [ADR023](https://docs.stackable.tech/home/stable/contributor/adr/ADR023-product-image-selection.html). Create reusable code, which will be integrated into all operators to enable new product/operator version specification.

This task is related to two epics:

* https://github.com/stackabletech/issues/issues/256
* https://github.com/stackabletech/issues/issues/257

The first epic is the primary focus at the moment, as it is a major blocker to get to multi-arch.

The feature will be implemented to be optional for now.

## Acceptance criteria

* [ ] Capture requirements described in [ADR](https://docs.stackable.tech/home/stable/contributor/adr/ADR023-product-image-selection.html) as test cases for the feature inside of operator-rs
  * All use cases and config examples described in "Option 6"
  * All examples described in "Decision Drivers"
* [ ] Operator using this in the future, will tell the function what product it operates on (to build the image name)
* [ ] Provide helper functions in operator-rs to replace existing helper functions like image_version and product_version in operator codebases
* [ ] Implement `pullPolicy` as well
* [ ] Implement `stackableVersion` case handling, as described in ADR Option 6
* [ ] Implement `custom` case handling, as described in ADR Option 6
  * limit functionality to public repositories
  * OPTIONAL: Implement `pullSecrets` aspects of ADR, and using private repositories
* [ ] Implement erroring when user tries to use `stackable` format for now
* [ ] Use changes in https://github.com/stackabletech/kafka-operator/issues/477 to verify that the solution is viable and usable

## Related out-of-scope followup tasks

To be performed in https://github.com/stackabletech/issues/issues/257

* If not implemented above: `pullSecrets` aspects of ADR, and using private repositories

# Status
* Needs work in kube-rs => https://github.com/kube-rs/kube/pull/1028
* Needs bump of kube-rs to 0.75.0 (and higher later on) in operator-rs to use the new feature => https://github.com/stackabletech/operator-rs/issues/471
* PR => https://github.com/stackabletech/operator-rs/pull/476
* PR for reference implementation in kafka-operator => https://github.com/stackabletech/kafka-operator/pull/482

lfrancke:
Is this an epic or a task?

vsupalov:
[...]

PR #476, opened 2022-09-27
[Merged by Bors] - Add product image selection struct

sbernauer:
## Description
For https://github.com/stackabletech/operator-rs/issues/470
Reference implementation in kafka: https://github.com/stackabletech/kafka-operator/pull/482
ADR is [here](https://docs.stackable.tech/home/stable/contributor/adr/ADR018-product_image_versioning.html)

The idea is to add all the currently known enum variants to see that the concept works. Before merging the `ProductImageSelection::Stackable` enum variant will be removed/commented out.

## Review Checklist

### DIFF
--- a/src/builder/pod/mod.rs
+++ b/src/builder/pod/mod.rs
@@ -3,6 +3,7 @@ pub mod security;
 pub mod volume;
 
 use crate::builder::meta::ObjectMetaBuilder;
+use crate::commons::product_image_selection::ProductImage;
 use crate::error::{Error, OperatorResult};
 
 use k8s_openapi::{
@@ -153,6 +154,18 @@ impl PodBuilder {
         self.image_pull_secrets
             .get_or_insert_with(Vec::new)
             .extend(secrets.map(|s| LocalObjectReference { name: Some(s) }));
+        self
+    }
+
+    pub fn image_pull_secrets_from_product_image(
+        &mut self,
+        product_image: &ProductImage,
+    ) -> &mut Self {
+        if let Some(pull_secrets) = &product_image.pull_secrets() {
+            self.image_pull_secrets
+                .get_or_insert_with(Vec::new)
+                .extend_from_slice(pull_secrets);
+        }
         self
     }
 
--- a/src/builder/pod/mod.rs
+++ b/src/builder/pod/mod.rs
@@ -157,8 +157,7 @@ impl PodBuilder {
         self
     }
 
-    /// Adds the following pod attributes from a [ResolvedProductImage]:
-    /// * image_pull_secrets
+    /// Extend the pod's image_pull_secrets field with the pull secrets from a given [ResolvedProductImage]
     pub fn image_pull_secrets_from_product_image(
         &mut self,
         product_image: &ResolvedProductImage,
--- a/src/builder/pod/mod.rs
+++ b/src/builder/pod/mod.rs
@@ -157,6 +157,8 @@ impl PodBuilder {
         self
     }
 
+    /// Adds the following container attributes from a [ResolvedProductImage]:
+    /// * image_pull_secrets
     pub fn image_pull_secrets_from_product_image(
         &mut self,
         product_image: &ResolvedProductImage,
--- a/src/builder/pod/mod.rs
+++ b/src/builder/pod/mod.rs
@@ -3,7 +3,7 @@ pub mod security;
 pub mod volume;
 
 use crate::builder::meta::ObjectMetaBuilder;
-use crate::commons::product_image_selection::ProductImage;
+use crate::commons::product_image_selection::ResolvedProductImage;
 use crate::error::{Error, OperatorResult};
 
 use k8s_openapi::{
@@ -159,9 +159,9 @@ impl PodBuilder {
 
     pub fn image_pull_secrets_from_product_image(
         &mut self,
-        product_image: &ProductImage,
+        product_image: &ResolvedProductImage,
     ) -> &mut Self {
-        if let Some(pull_secrets) = &product_image.pull_secrets() {
+        if let Some(pull_secrets) = &product_image.pull_secrets {
             self.image_pull_secrets
                 .get_or_insert_with(Vec::new)
                 .extend_from_slice(pull_secrets);
--- a/src/builder/pod/mod.rs
+++ b/src/builder/pod/mod.rs
@@ -157,7 +157,7 @@ impl PodBuilder {
         self
     }
 
-    /// Adds the following container attributes from a [ResolvedProductImage]:
+    /// Adds the following pod attributes from a [ResolvedProductImage]:
     /// * image_pull_secrets
     pub fn image_pull_secrets_from_product_image(
         &mut self,
--- a/src/commons/product_image_selection.rs
+++ b/src/commons/product_image_selection.rs
@@ -47,11 +47,11 @@ pub struct ResolvedProductImage {
 }
 
 impl ProductImageSelection {
-    pub fn resolve(self, image_base_name: &str) -> ResolvedProductImage {
+    pub fn resolve(&self, image_base_name: &str) -> ResolvedProductImage {
         match self {
             ProductImageSelection::Custom(custom) => ResolvedProductImage {
-                image: custom.custom,
-                product_version: custom.product_version,
+                image: custom.custom.to_string(),
+                product_version: custom.product_version.to_string(),
             },
             ProductImageSelection::StackableVersion(stackable_version) => {
                 let repo = stackable_version
@@ -65,7 +65,7 @@ impl ProductImageSelection {
                 );
                 ResolvedProductImage {
                     image,
-                    product_version: stackable_version.product_version,
+                    product_version: stackable_version.product_version.to_string(),
                 }
             }
             ProductImageSelection::Stackable(stackable) => {
@@ -78,7 +78,7 @@ impl ProductImageSelection {
                 );
                 ResolvedProductImage {
                     image,
-                    product_version: stackable.product_version,
+                    product_version: stackable.product_version.to_string(),
                 }
             }
         }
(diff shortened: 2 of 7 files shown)


src/builder/pod/mod.rs
```diff
@@ -156,6 +157,19 @@ impl PodBuilder {
         self
     }
 
+    /// Extend the pod's image_pull_secrets field with the pull secrets from a given [ResolvedProductImage]
+    pub fn image_pull_secrets_from_product_image(
```
teozkr, line 161: Honestly I don't think I have a better way to do this.. but it feels awkward that you need to call both `image_from` and `image_pull_secrets_from` to get the correct behaviour, with no warning that you've done it wrong.

src/commons/product_image_selection.rs
```diff
@@ -0,0 +1,347 @@
...
+pub enum ProductImageSelection {
+    // Order matters!
+    // The variants will be tried from top to bottom
+    Custom(ProductImageCustom),
+    StackableVersion(ProductImageStackableVersion),
+    // The following enum variant is commented out for now, as the operators currently don't know which stackableVersions they are compatible with.
+    // In the future they should be able to automatically pick a fitting stackableVersion for them.
+    // The concept of the untagged enum was tested with all known upcoming variants to make sure we don't run into strange problems in the future.
+    // They code snippets are left for illustration and can be commented in as soon as the operators can pick the stackableVersions automatically.
+    Stackable(ProductImageStackable),
```
teozkr, line 34: Still not sure why these need to be separate rather than just making `stackable_version` optional at some future time.

src/commons/product_image_selection.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+impl Default for ProductImageSelection {
+    fn default() -> Self {
+        Self::Stackable(ProductImageStackable {
+            product_version: "auto".to_string(),
```
teozkr, line 40: If `auto` is special then it should probably be an enum...

src/commons/product_image_selection.rs
```diff
@@ -0,0 +1,347 @@
...
+    stackable_version: String,
+    /// Name of the docker repo, e.g. `docker.stackable.tech/stackable`
+    repo: Option<String>,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ProductImageStackable {
+    /// Version of the product, e.g. `1.4.1`.
+    /// Set it to `auto` to let the operator automatically pick the recommended product version he is compatible with.
```
teozkr, line 71: Is `auto` intended to only work for PIS and not PISV?

src/commons/product_image_selection.rs
```diff
@@ -0,0 +1,347 @@
...
+    Custom(ProductImageCustom),
+    StackableVersion(ProductImageStackableVersion),
+    // The following enum variant is commented out for now, as the operators currently don't know which stackableVersions they are compatible with.
+    // In the future they should be able to automatically pick a fitting stackableVersion for them.
+    // The concept of the untagged enum was tested with all known upcoming variants to make sure we don't run into strange problems in the future.
+    // They code snippets are left for illustration and can be commented in as soon as the operators can pick the stackableVersions automatically.
+    Stackable(ProductImageStackable),
+}
+
+impl Default for ProductImageSelection {
```
teozkr, line 37: Doesn't look like this `Default` impl is used anywhere?

src/commons/product_image_selection.rs
```diff
@@ -0,0 +1,347 @@
+use k8s_openapi::api::core::v1::LocalObjectReference;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use strum::AsRefStr;
+
+pub const STACKABLE_DOCKER_REPO: &str = "docker.stackable.tech/stackable";
+
+#[derive(Clone, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ProductImage {
```
teozkr, line 10: Should this be fragmented/optional?

src/commons/product_image_selection.rs
sbernauer: Yes. In PISV case the users needs to specify the stackable version, hence the exact product version is also needed

src/commons/product_image_selection.rs
sbernauer: I think previously we needed it this way, but now i agree with you.
We will comment out this variant for now anyway

src/builder/pod/mod.rs
sbernauer: Agreed, but i also don't know a better way of doing this

src/commons/product_image_selection.rs
sbernauer: Honestly, i have never used fragments yet. i would leave the decision to you. I personally am quite happy with the current solution

src/commons/product_image_selection.rs
sbernauer: Yep, that's correct. Was a leftover, removed it

src/commons/product_image_selection.rs
sbernauer: Not entirely sure if this gives the best user experience. Original it was planned to make it Option<String> but that's not possible.
Anyway we will comment out this code, as we don't support the recommended Product version yet. I would suggest reviewing that when we actually plan to add that functionality

sbernauer:
@teozkr would be great if you could find the time to have another look so that we can complete things from the board

src/commons/product_image_selection.rs
teozkr: It's probably not required for now...

src/commons/product_image_selection.rs
teozkr: Did you get around to looking at this?

src/commons/product_image_selection.rs
teozkr: Looks like this is still around?

sbernauer: I will remove the `auto` feature from this PR once you are happy with the overall solution

src/commons/product_image_selection.rs
sbernauer: I haven't started to be honest. I would think about the actual design when we start implementing the `Stackable``feature (use this product version and automatically find the stackable version)

teozkr: Disabling that mode (for now) would be ok for me.

src/commons/product_image_selection.rs
teozkr: The overall structure is ok to me.

sbernauer:
@teozkr removed the `Stackable` enum variant

teozkr (approved):

sbernauer:
bors cancel

sbernauer:
I will also add some helpers for the recommended labels

sbernauer:
@teozkr added the label helpers, please have a look

src/commons/product_image_selection.rs
```diff
@@ -0,0 +1,313 @@
...
+    stackable_version: String,
+    /// Name of the docker repo, e.g. `docker.stackable.tech/stackable`
+    repo: Option<String>,
+}
+
+#[derive(Clone, Debug, PartialEq, JsonSchema)]
+pub struct ResolvedProductImage {
+    /// Version of the product, e.g. `1.4.1`.
+    pub product_version: String,
+    /// App version as it should be used into [`get_recommended_labels`]
```
teozkr, line 60:
```suggestion
    /// App version as formatted for [`get_recommended_labels`]
```

src/labels.rs
```diff
@@ -31,10 +33,16 @@ pub struct ObjectLabels<'a, T> {
     pub owner: &'a T,
     /// The name of the app being managed, such as `zookeeper`
     pub app_name: &'a str,
-    /// The version of the app being managed (not of the operator)
+    /// The version of the app being managed (not of the operator).
+    ///
+    /// If setting this label on a Stackable product please use the provided `app_version_label` attribute of the
+    /// [`ResolvedProductImage`] struct
```
teozkr, line 39:
```suggestion
    /// If setting this label on a Stackable product then please use [`ResolvedProductImage::app_version_label`]
```

src/labels.rs
```diff
@@ -31,10 +33,16 @@ pub struct ObjectLabels<'a, T> {
...
     ///
-    /// This version should include the Stackable version, such as `0.1.0-stackable0.1.0`. In the case of custom product images,
-    /// the tag should be appended, such as `<productversion>-<tag>`. However, this is pure documentation and should not be parsed.
+    /// This version should include the Stackable version, such as `3.0.0-stackable0.1.0`.
+    /// If the Stackable version is not known, the product version should be used together with a suffix (if possible).
```
teozkr, line 42:
```suggestion
    /// If the Stackable version is not known, then the product version should be used together with a suffix (if possible).
```

src/labels.rs
```diff
@@ -31,10 +33,16 @@ pub struct ObjectLabels<'a, T> {
...
+    /// In the case of custom product images provided by the user, only the product version is known,
```
teozkr, line 43:
```suggestion
    /// If a custom product image is provided by the user (in which case only the product version is known),
```

src/labels.rs
```diff
@@ -31,10 +33,16 @@ pub struct ObjectLabels<'a, T> {
...
+    /// `3.0.0-<tag-of-custom-image>` should be used.
```
teozkr, line 44:
```suggestion
    /// then the format `3.0.0-<tag-of-custom-image>` should be used.
```

src/labels.rs
```diff
@@ -31,10 +33,16 @@ pub struct ObjectLabels<'a, T> {
...
+    /// However, this is pure documentation and should not be parsed.
```
teozkr, line 45:
```suggestion
    ///
    /// However, this is pure documentation and should not be parsed.
```

teozkr (approved):

Issue #491, opened 2022-10-17
Maybe get_recommended_labels should take a struct of named arguments


teozkr:
The current list of string arguments is confusing to follow (especially if you don't use rust-analyzer with parameter inlays), and is confusing and error-prone to update in operators when new arguments are added (often in the middle of the list).

This PR adds another parameter to the list: https://github.com/stackabletech/operator-rs/pull/492/files#diff-f3a8a77aa3434fb4aef8eb9c3d011832fc27301972b9975aa303832c3ed33675 Clippy also complains about the many parameters.

This is the function in its current state: https://github.com/stackabletech/operator-rs/blob/main/src/labels.rs#L22

Acceptance criteria:
- [ ] get_recommended_labels does not take so many arguments
- [ ] with_recommended_labels also

teozkr:
I actually kind of think even just doing the easy "struct of `&str`s would be a huge improvement, but I believe @siegfriedweber had strong opinions here so I'll wait for him to return...

siegfriedweber:
A struct would be an improvement because the field names help to assign the correct values and the struct can be reused so that repetition is reduced.

I had a more elaborate solution in mind where you just pass the cluster spec and a "controller spec" where the recommended labels are derived from. The `ClusterResources` already contain some of the required information. The application version could be made accessible via a Trait. The labels for the role and role-group could be extracted into another builder function because not all resources are at a role-group level and they currently use "global" as a placeholder. If the recommended labels do not match exactly what the caller needs then she or he can overwrite them with other functions in the `ObjectMetaBuilder`.

But after a closer look, the implementation of my solution is not as straight forward as I hoped. So I am also fine with the easy struct solution.

teozkr:
Is there much further refinement planned here then?

fhennig:
We could still use a trait instead of a struct and then in the trait we could implement functions like the `superset_version` and we could hardcode some things in there, instead of having the `OPERATOR_NAME` static vars etc. 

But it's not a strong opinion

lfrancke:
The description does not really talk about the solution but the details are in the comments so I'm fine with it.

PR #492, opened 2022-10-17
[Merged by Bors] - Separate operator and controller names

### LINKED ISSUE #482: ClusterResources should take the operator and controller names separately
Currently, `ClusterResources` takes a single `manager` argument that must be unique per controller. [The documentation suggests identifying both the controller and operator names](https://github.com/stackabletech/operator-rs/blob/406c2fb62144e834981fdd89ed7203c10c0a9e5e/src/cluster_resources.rs#L161-L166) (`foo-operator_foo-controller`), but [in practice many single-controller operators just use the operator name](https://github.com/stackabletech/hbase-operator/pull/215/files/258ad6d44bff169f3566a9ffeccbeed980284662#diff-86a911b13abf77825206401b5fca7a0dece65999676e0ec1bf4b666e559e939aR36) (`foo-operator`).

This is not only inconsistent across the platform, but can lead to orphaned resources during upgrades that add more controllers (necessitating a switch to the expanded format). Additionally, it's generally a convention that un-namedspaced identifiers are reserved for the cluster administrators and users, third-party software should use the `ns/id` format for string keys (such as field managers or labels).

# Acceptance

- [ ] `ClusterResources::new` takes separate arguments for the operator and controller names
- [ ] `labels::get_recommended_labels` takes separate arguments for the operator and controller names
- [ ] `managed-by` labels and field managers follow the `operator/controller` format
- [ ] The documentation is amended to encourage FQDN-style operator names (for example: `zookeeper.stackable.tech` rather than `zookeeper-operator`)

teozkr:
## Description

Fixes #482

## Review Checklist

### DIFF
--- a/src/builder/meta.rs
+++ b/src/builder/meta.rs
@@ -339,7 +339,7 @@ mod tests {
                 "test_app",
                 "1.0",
                 "app.stackable.tech",
-                "appclusters",
+                "appcluster",
                 "component",
                 "role",
             )
--- a/src/builder/meta.rs
+++ b/src/builder/meta.rs
@@ -334,7 +334,15 @@ mod tests {
             .namespace("bar")
             .ownerreference_from_resource(&pod, Some(true), Some(false))
             .unwrap()
-            .with_recommended_labels(&pod, "test_app", "1.0", "app-operator", "component", "role")
+            .with_recommended_labels(
+                &pod,
+                "test_app",
+                "1.0",
+                "app.stackable.tech",
+                "appclusters",
+                "component",
+                "role",
+            )
             .with_annotation("foo", "bar")
             .build();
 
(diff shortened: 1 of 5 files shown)


src/builder/meta.rs
```diff
@@ -153,17 +153,19 @@ impl ObjectMetaBuilder {
         resource: &T,
         app_name: &str,
         app_version: &str,
-        app_managed_by: &str,
-        app_component: &str,
-        role_name: &str,
+        operator_name: &str,
+        controller_name: &str,
+        app_role: &str,
+        app_role_group: &str,
```
teozkr, line 159: This triggers a cargo lint due to #491, not sure if we want to tackle that in this PR...

siegfriedweber: I would combine both tickets into one PR because they create a breaking change at the same place. If we would merge this PR and not immediately the PR for #491 then we would force people to adapt their code twice. But I would also not implement #491 hastily without a thourough refinement.

teozkr: Updated to account for #491 being merged.

sbernauer (approved):

PR #501, opened 2022-11-10
[Merged by Bors] - Structify arguments to `get_recommended_labels`

### LINKED ISSUE #491: Maybe get_recommended_labels should take a struct of named arguments

teozkr:
## Description

Hopefully this should help clarify how the different names hold together, and help make it clearer how to adjust for later changes going forward.

This doesn't cover `role_group_selector_labels` or `role_selector_labels` since they have a more obvious inherent structure to them, and I don't see them changing as much (especially since they are effectively part of the public API).

This is kind of coupled to #492, do we want to merge that into this PR or do that one on top of this afterwards?

Fixes #491.

## Review Checklist

### DIFF
--- a/src/labels.rs
+++ b/src/labels.rs
@@ -32,8 +32,8 @@ pub struct ObjectLabels<'a, T> {
     pub app_name: &'a str,
     /// The version of the app being managed (not of the operator)
     ///
-    /// This version should include the Stackable version, such as `0.1.0-stackable0.1.0`. However, this is pure
-    /// documentation and should not be parsed.
+    /// This version should include the Stackable version, such as `0.1.0-stackable0.1.0`. In the case of custom product images,
+    /// the tag should be appended, such as `<productversion>-<tag>`. However, this is pure documentation and should not be parsed.
     pub app_version: &'a str,
     /// The name of the operator and controller managing the object
     pub managed_by: &'a str,
--- a/src/labels.rs
+++ b/src/labels.rs
@@ -2,6 +2,9 @@ use const_format::concatcp;
 use kube::api::{Resource, ResourceExt};
 use std::collections::BTreeMap;
 
+#[cfg(doc)]
+use crate::builder::ObjectMetaBuilder;
+
 const APP_KUBERNETES_LABEL_BASE: &str = "app.kubernetes.io/";
 
 /// The name of the application e.g. "mysql"
--- a/src/labels.rs
+++ b/src/labels.rs
@@ -18,24 +18,45 @@ pub const APP_PART_OF_LABEL: &str = concatcp!(APP_KUBERNETES_LABEL_BASE, "part-o
 pub const APP_MANAGED_BY_LABEL: &str = concatcp!(APP_KUBERNETES_LABEL_BASE, "managed-by");
 pub const APP_ROLE_GROUP_LABEL: &str = concatcp!(APP_KUBERNETES_LABEL_BASE, "role-group");
 
+/// Recommended labels to set on objects created by Stackable operators
+///
+/// See [`get_recommended_labels`] and [`ObjectMetaBuilder::with_recommended_labels`].
+#[derive(Debug, Clone, Copy)]
+pub struct ObjectLabels<'a, T> {
+    /// The name of the object that this object is being created on behalf of, such as a `ZookeeperCluster`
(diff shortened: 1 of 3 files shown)


src/labels.rs
```diff
@@ -18,61 +21,82 @@ pub const APP_PART_OF_LABEL: &str = concatcp!(APP_KUBERNETES_LABEL_BASE, "part-o
...
+///
+/// See [`get_recommended_labels`] and [`ObjectMetaBuilder::with_recommended_labels`].
+#[derive(Debug, Clone, Copy)]
+pub struct ObjectLabels<'a, T> {
+    /// The name of the object that this object is being created on behalf of, such as a `ZookeeperCluster`
+    pub owner: &'a T,
+    /// The name of the app being managed, such as `zookeeper`
+    pub app_name: &'a str,
+    /// The version of the app being managed (not of the operator)
+    pub app_version: &'a str,
```
sbernauer, line 34: Could we please document what exactly is meant to be put in here? is it
* 3.3.1
* 3.3.1-stackable0.1.0
* One of the above
* It doesn't really matter, its for documentation purpose

AFAIK we currently put in `3.3.1-stackable0.1.0` but will need to switch to `3.3.1` when implementing product image select (as we don't know the stackable version in all cases)

src/labels.rs
```diff
@@ -18,61 +21,82 @@ pub const APP_PART_OF_LABEL: &str = concatcp!(APP_KUBERNETES_LABEL_BASE, "part-o
...
-    app_role: &str,
-    app_role_group: &str,
+    ObjectLabels {
+        owner,
+        app_name,
+        app_version,
+        managed_by,
+        role,
+        role_group,
+    }: ObjectLabels<T>,
```
sbernauer, line 52: Nice, didn't know that's possible :D

sbernauer (commented):
LGTM overall

src/labels.rs
teozkr: I'd say a combination of 1 and 4. But sure.

teozkr: Actually, on second thought some properties depend on the stackable release too so maybe it just makes sense to put in both...

teozkr: caf2ac6

sbernauer: I would prefer `3.3.1-stackable0.1.0`, but we don't have the information of the stackable version `0.1.0` if the users provides a custom image.
Is it ok in that case to put in `3.3.1`? If so, would be great to have this documented to avoid further discussion.
E.g.

This version should include the Stackable version, such as `3.0.0-stackable0.1.0`. If the Stackable version is not known, only the product version can be used, such as `3.0.0`. However, this is pure documentation and should not be parsed.

teozkr: > but we don't have the information of the stackable version 0.1.0 if the users provides a custom image.

In that case I'd just put whatever tag the user specified. Or `<version>-<tag>` or some such.

teozkr: (Hopefully whatever tagging system the user chooses to use actually makes sense to them, at least...)

sbernauer: Ah that's a good idea. So new suggestion

This version should include the Stackable version, such as 3.0.0-stackable0.1.0. If the Stackable version is not known, the product version should be used together with a suffix (if possible). In the case of custom product images provided by the user, only the product version is known, `3.0.0-<tag-of-custom-image>` should be used. However, this is pure documentation and should not be parsed.

teozkr: 165c1943511220a2bf81836a278c41199cdf7f65

sbernauer: thx!

sbernauer (approved):

PR #536, opened 2023-01-16
[Merged by Bors] - Added empty dir helper to PodBuilder

fhennig:
## Description

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Extended the `LdapAuthenticationProvider` with functionality to build add Volumes and Mounts to PodBuilder and ContainerBuilder ([#535]).
-- extended the `PodBuilder` with `add_volume_with_empty_dir` utility function ([#536]).
+- Extended the `PodBuilder` with `add_volume_with_empty_dir` utility function ([#536]).
 
 [#535]: https://github.com/stackabletech/operator-rs/pull/535
 [#536]: https://github.com/stackabletech/operator-rs/pull/536
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Extended the `LdapAuthenticationProvider` with functionality to build add Volumes and Mounts to PodBuilder and ContainerBuilder ([#535])
+- Extended the `LdapAuthenticationProvider` with functionality to build add Volumes and Mounts to PodBuilder and ContainerBuilder ([#535]).
+- extended the `PodBuilder` with `add_volume_with_empty_dir` utility function ([#536]).
 
 [#535]: https://github.com/stackabletech/operator-rs/pull/535
+[#536]: https://github.com/stackabletech/operator-rs/pull/536
 
 ## [0.30.2] - 2022-12-20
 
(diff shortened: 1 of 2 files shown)


src/builder/pod/mod.rs
```diff
@@ -163,6 +163,16 @@ impl PodBuilder {
         self
     }
 
+    /// Utility function for the common case of adding an empty dir Volume
```
sbernauer, line 166:
```suggestion
    /// Utility function for the common case of adding an emptyDir Volume
```

src/builder/pod/mod.rs
```diff
@@ -163,6 +163,16 @@ impl PodBuilder {
...
+    /// with the given name and no medium and no quantity.
+    pub fn add_volume_with_empty_dir(&mut self, name: impl Into<String>) -> &mut Self {
```
sbernauer, line 168:
```suggestion
    pub fn add_empty_dir_volume(&mut self, name: impl Into<String>) -> &mut Self {
```

sbernauer (approved):
lgtm otherwise

fhennig:
bors cancel

CHANGELOG.md
```diff
@@ -6,9 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Extended the `LdapAuthenticationProvider` with functionality to build add Volumes and Mounts to PodBuilder and ContainerBuilder ([#535])
+- Extended the `LdapAuthenticationProvider` with functionality to build add Volumes and Mounts to PodBuilder and ContainerBuilder ([#535]).
+- extended the `PodBuilder` with `add_volume_with_empty_dir` utility function ([#536]).
```
maltesander, line 10:
```suggestion
- Extended the `PodBuilder` with `add_volume_with_empty_dir` utility function ([#536]).
```

src/builder/pod/mod.rs
```diff
@@ -163,6 +163,16 @@ impl PodBuilder {
...
 
+    /// Utility function for the common case of adding an emptyDir Volume
+    /// with the given name and no medium and no quantity.
+    pub fn add_empty_dir_volume(&mut self, name: impl Into<String>) -> &mut Self {
+        self.add_volume(
+            VolumeBuilder::new(name)
+                .with_empty_dir(None::<String>, None)
+                .build(),
+        )
+    }
```
maltesander, line 174: We should at least be able to set a size? Meaning pass through the `quantity: Option<Quantity>`

maltesander (changes requested):

src/builder/pod/mod.rs
fhennig: I think the majority of cases doesn't use a size, and I'd like to keep it tidy. why not have two functions in that case? We can have as many as we want!

maltesander: I dont see the point for that. `with_empty_dir` already takes a quantity, why not pass it through instead of setting it to None  and/or create another method?

fhennig: Otherwise if I'm reading the code I gotta think about this `None` parameter. If there is nothing its easier to read/think about

maltesander: I mean you have the prototype of that function, you can but do not have to specify a quantity (which will default to whatever). I dont think this is hard to read :D

maltesander (approved):
LGTM.

fhennig:
bors merge

PR #544, opened 2023-01-25
[Merged by Bors] - Feat: Implement various Ops for MemoryQuantity; new CpuQuantity

fhennig:
## Description

I have used these functions in my Druid PR already: https://github.com/stackabletech/druid-operator/pull/380

## Review Checklist

### DIFF
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -1,5 +1,7 @@
 use std::str::FromStr;
 
+use k8s_openapi::apimachinery::pkg::api::resource::Quantity;
+
 use crate::error::{Error, OperatorResult};
 
 /// A wrapper around a Quantity, to make working with CPU quantities easier.
@@ -50,6 +52,22 @@ impl FromStr for CpuQuantity {
                 millis: millis_float as usize,
             });
         }
+    }
+}
+
+impl TryFrom<&Quantity> for CpuQuantity {
+    type Error = Error;
+
+    fn try_from(q: &Quantity) -> Result<Self, Self::Error> {
+        Self::from_str(&q.0)
+    }
+}
+
+impl TryFrom<Quantity> for CpuQuantity {
+    type Error = Error;
+
+    fn try_from(q: Quantity) -> Result<Self, Self::Error> {
+        Self::try_from(&q)
     }
 }
 
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -8,7 +8,7 @@ use crate::error::{Error, OperatorResult};
 /// Supports conversion from quantities.
 ///
 /// A CPU Quantity cannot have a precision finer than 'm' (millis), so we use that as
-/// our internal representation (see: 
+/// our internal representation (see:
 /// `<https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu>`).
 pub struct CpuQuantity {
     millis: usize,
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -49,7 +49,7 @@ impl FromStr for CpuQuantity {
             let cpu_millis: usize = value.parse().map_err(|_| Error::InvalidCpuQuantity {
                 value: q.to_owned(),
             })?;
-            Ok(CpuQuantity { millis: cpu_millis })
+            Ok(Self::from_millis(cpu_millis))
         } else {
             let cpus = q.parse::<f32>().map_err(|_| Error::InvalidCpuQuantity {
                 value: q.to_owned(),
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -85,4 +85,13 @@ mod test {
         let result = CpuQuantity::from_str(s).unwrap();
         assert_eq!(millis, result.as_milli_cpus())
     }
+
+    #[rstest]
+    #[case("1.11111")]
+    #[case("1000.1m")]
+    #[case("500k")]
+    fn test_from_str_err(#[case] s: &str) {
+        let result = CpuQuantity::from_str(s);
+        assert!(result.is_err());
+    }
 }
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -31,6 +31,12 @@ impl CpuQuantity {
 impl FromStr for CpuQuantity {
     type Err = Error;
 
+    /// Only two formats can be parsed
+    /// - <usize>m
+    /// - <f32>
+    /// For the float, only milli-precision is supported.
+    /// Using more precise values will trigger an error, and using any other
+    /// unit than 'm' or None will also trigger an error.
     fn from_str(q: &str) -> OperatorResult<Self> {
         let start_of_unit = q.find(|c: char| c != '.' && !c.is_numeric());
         if let Some(start_of_unit) = start_of_unit {
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -91,6 +91,9 @@ mod test {
     #[case("1000m", 1000)]
     #[case("500m", 500)]
     #[case("2.5", 2500)]
+    #[case("0.2", 200)]
+    #[case("0.02", 20)]
+    #[case("0.002", 2)]
     fn test_from_str(#[case] s: &str, #[case] millis: usize) {
         let result = CpuQuantity::from_str(s).unwrap();
         assert_eq!(millis, result.as_milli_cpus())
@@ -100,6 +103,7 @@ mod test {
     #[case("1.11111")]
     #[case("1000.1m")]
     #[case("500k")]
+    #[case("0.0002")]
     fn test_from_str_err(#[case] s: &str) {
         let result = CpuQuantity::from_str(s);
         assert!(result.is_err());
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -37,7 +37,7 @@ impl FromStr for CpuQuantity {
             let cpu_millis: usize = value.parse().map_err(|_| Error::InvalidCpuQuantity {
                 value: q.to_owned(),
             })?;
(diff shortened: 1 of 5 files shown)


src/memory.rs
```diff
@@ -127,7 +127,102 @@ pub fn to_java_heap_value(
...
+            unit: BinaryMultiple::Gibi,
+        }
+    }
+
+    pub fn from_mebi(mebi: f32) -> Self {
+        Self {
+            value: mebi,
+            unit: BinaryMultiple::Mebi,
+        }
+    }
```
maltesander, line 150: I do not really see the point of this. I would combine this into a generic Self::new(value: f32, unit: ...) or use MemoryQuantity {} directly.

src/memory.rs
```diff
@@ -127,7 +127,102 @@ pub fn to_java_heap_value(
...
+    /// If the MemoryQuantity value is smaller than 1 (starts with a zero), convert it to a smaller
+    /// unit until the non fractional part of the value is not zero anymore.
+    /// This can fail if the quantity is smaller than 1kB.
+    fn ensure_no_zero(&self) -> OperatorResult<Self> {
+        if self.value < 1. {
+            self.scale_down_unit()?.ensure_no_zero()
+        } else {
+            Ok(*self)
+        }
+    }
```
maltesander, line 194: The value could be 0 here? If yes this will loop forever. I would add a check for this.

src/memory.rs
```diff
@@ -127,7 +127,102 @@ pub fn to_java_heap_value(
...
+    /// The original quantity may be rounded down to achive a compact, natural number representation.
+    /// This rounding may cause the quantity to shrink by up to 20MB.
+    /// Useful to set memory quantities as JVM paramters.
+    pub fn format_for_java(&self) -> OperatorResult<String> {
+        let m = self
+            .scale_to_at_most_gb() // Java Heap only supports specifying kb, mb or gb
+            .ensure_no_zero()? // We don't want 0.9 or 0.2
+            .ensure_integer(MemoryQuantity::from_mebi(20.))?; // Java only accepts integers not floats
+        Ok(format!("{}{}", m.value, m.unit.to_java_memory_unit()))
+    }
```
maltesander, line 224: Why 20MB?

src/memory.rs
```diff
@@ -184,6 +279,88 @@ impl Mul<f32> for MemoryQuantity {
...
+                value: self.scale_to(rhs.unit).value + rhs.value,
+                unit: rhs.unit,
+            }
+        } else {
+            MemoryQuantity {
+                value: self.value + rhs.scale_to(self.unit).value,
+                unit: self.unit,
+            }
+        }
+    }
```
maltesander, line 333: I think its counter intuitive to eventually have the `self` unit changed. I think this can be shortend to (same for substract):
```suggestion
    fn add(self, rhs: MemoryQuantity) -> Self::Output {
        MemoryQuantity {
            value: self.value + rhs.scale_to(self.unit).value,
            unit: self.unit,
        }
    }
```

src/memory.rs
```diff
@@ -184,6 +279,88 @@ impl Mul<f32> for MemoryQuantity {
...
+        let this_val = self.scale_to(BinaryMultiple::get_smallest()).value;
+        let other_val = other.scale_to(BinaryMultiple::get_smallest()).value;
+        this_val.partial_cmp(&other_val)
+    }
+}
+
+impl Ord for MemoryQuantity {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        let this_val = self.scale_to(BinaryMultiple::get_smallest()).value;
+        let other_val = other.scale_to(BinaryMultiple::get_smallest()).value;
```
maltesander, line 347: The scaling is not required, already covered in `partial_cmp`

src/memory.rs
```diff
@@ -184,6 +279,88 @@ impl Mul<f32> for MemoryQuantity {
...
+        // Note: We just assume that our values are always not NaN, so we are actually Ord.
+        // A MemoryQuantity with NaN is not permissible.
```
maltesander, line 349: Im not a fan of having unwraps. This is 100% sure :D ?

src/memory.rs
```diff
@@ -127,7 +127,102 @@ pub fn to_java_heap_value(
...
+            value: self.value.fract(),
+            unit: self.unit,
+        };
+        if fraction_memory < tolerated_rounding_loss {
+            Ok(self.floor())
+        } else {
+            self.scale_down_unit()?
+                .ensure_integer(tolerated_rounding_loss)
+        }
+    }
```
maltesander, line 211: Again, be careful with the recursion and zero values.

maltesander (changes requested):

src/cpu.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    #[rstest]
+    #[case("1.11111")]
+    #[case("1000.1m")]
+    #[case("500k")]
+    fn test_from_str_err(#[case] s: &str) {
+        let result = CpuQuantity::from_str(s);
+        assert!(result.is_err());
+    }
+}
```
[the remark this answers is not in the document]
fhennig: Good suggestion, thanks!

src/memory.rs
fhennig: I don't want to expose the `BinaryMultiple` if I can avoid it. It also reads nicely in code. If you don't oppose strongly, I'd like to keep it.

src/memory.rs
fhennig: Well, the other java formatting function uses 0.2 arbitrarily. thats why it's with a "2". But other than that, I don't know

src/memory.rs
fhennig: Of course you could create a MemoryQuantity with a `NaN` value, and this will break. We could try to guard against this at all the places where the MemoryQuantity is created, and try to not have a NaN inside like that to make this unwrap safer, but we'd still have to have the unwrap here.

The problem is that floats are actually not ordered, because of NaN. so in the `None` case, what should we return? There is only "Less", "Equal", or "Greater". We can't return an error, because this is a trait function and we _have_ to return an Ordering. What do you think is the best way forward here?

src/memory.rs
maltesander: But the 0.2 is the leftover we would not use for heap allocation from the overall resources? Thats different. Or do i get something wrong here?

src/error.rs
```diff
@@ -85,12 +85,21 @@ pub enum Error {
...
+    InvalidCpuQuantity { value: String },
+
     #[error("Invalid quantity unit [{value}]")]
     InvalidQuantityUnit { value: String },
 
     #[error("No quantity unit provided for [{value}]")]
     NoQuantityUnit { value: String },
 
+    #[error("Unsupported Precision [{value}]")]
+    UnsupportedQuantityPrecision { value: String },
```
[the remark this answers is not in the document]
fhennig: done

src/memory.rs
maltesander: BinaryMultiple is pub and exposed? What if we need kibi etc.? Not sure what the gain is here.

src/memory.rs
fhennig: Alright, I see now that an unwrap is not good. This is unfortunate, but I've opted to simply remove the `Ord` implementation.

maltesander: fine with me :)

src/cpu.rs
```diff
@@ -0,0 +1,109 @@
+use std::str::FromStr;
+
+use k8s_openapi::apimachinery::pkg::api::resource::Quantity;
+
+use crate::error::{Error, OperatorResult};
+
+/// A wrapper around a [`Quantity`], to make working with CPU quantities easier.
```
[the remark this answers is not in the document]
fhennig: updated

src/error.rs
```diff
@@ -85,12 +85,21 @@ pub enum Error {
...
+    #[error("Invalid quantity [{value}]")]
+    InvalidCpuQuantity { value: String },
+
     #[error("Invalid quantity unit [{value}]")]
     InvalidQuantityUnit { value: String },
 
     #[error("No quantity unit provided for [{value}]")]
     NoQuantityUnit { value: String },
 
+    #[error("Unsupported Precision [{value}]. CPU quantities can only be given without a unit or with 'm")]
```
maltesander, line 97: Not sure if we can copy this but thats whats written in the kubernetes docs.
```suggestion
    #[error("Unsupported Precision [{value}]. Kubernetes doesn't allow you to specify CPU resources with a precision finer than 1m. Because of this, it's useful to specify CPU units less than 1.0 or 1000m using the milliCPU form; for example, 5m rather than 0.005")]
```

src/memory.rs
```diff
@@ -82,6 +82,9 @@ pub struct MemoryQuantity {
 ///   supports up to Gibibytes while K8S quantities can be expressed in Exbibytes.
 /// - the heap size has a non-zero value.
 /// Fails if it can't enforce the above restrictions.
+#[deprecated(
+    note = "use \"-Xmx\" + MemoryQuantity::try_from(quantity).scale_to(unit).format_for_java()"
+)]
```
maltesander, line 87: I would add the version (if the new tag will be 0.33.0). Did not check clippy for this. Same for `to_java_heap_value`
```suggestion
#[deprecated(
    since= "0.33.0",
    note = "use \"-Xmx\" + MemoryQuantity::try_from(quantity).scale_to(unit).format_for_java()"
)]
```

fhennig: done

maltesander (commented):
Changelog missing (please mention the deprecation). Despite the 20MB LGTM!

fhennig:
bors merge

Issue #596, opened 2023-05-31
Add support for requests, limits and ResourceQuotas


Techassi:
See ["Configure Memory and CPU Quotas for a Namespace"](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/) in the Kubernetes docs.

Each container (and pod) needs to specify how much CPU and memory can be consumed at max. The implementation should use default values if no explicit values are set. The framework also needs to allow setting custom limits on containers and pods.

PR #598, opened 2023-06-01
Add ContainerType and default resource quotas

Techassi:
Implements #596

### DIFF
--- a/src/builder/pod/container.rs
+++ b/src/builder/pod/container.rs
@@ -5,8 +5,8 @@ use k8s_openapi::api::core::v1::{
 use std::fmt;
 
 use crate::{
-    commons::product_image_selection::ResolvedProductImage, error::Error,
-    validation::is_rfc_1123_label,
+    commons::product_image_selection::ResolvedProductImage, container_type::ContainerType,
+    error::Error, validation::is_rfc_1123_label,
 };
 
 /// A builder to build [`Container`] objects.
@@ -27,6 +27,10 @@ pub struct ContainerBuilder {
     liveness_probe: Option<Probe>,
     startup_probe: Option<Probe>,
     security_context: Option<SecurityContext>,
+
+    /// Type of container. This indicates wether the container is a main, init
+    /// or sidecar container. Defaults to [`ContainerType::Main`].
+    container_type: ContainerType,
 }
 
 impl ContainerBuilder {
@@ -223,14 +227,25 @@ impl ContainerBuilder {
         self
     }
 
+    /// Sets the container type.
+    pub fn container_type(&mut self, container_type: ContainerType) -> &mut Self {
+        self.container_type = container_type;
+        self
+    }
+
     pub fn build(&self) -> Container {
+        let resources = self
+            .resources
+            .clone()
+            .unwrap_or(ResourceRequirements::from(&self.container_type));
+
         Container {
             args: self.args.clone(),
             command: self.command.clone(),
             env: self.env.clone(),
             image: self.image.clone(),
             image_pull_policy: self.image_pull_policy.clone(),
-            resources: self.resources.clone(),
+            resources: Some(resources),
             name: self.name.clone(),
             ports: self.container_ports.clone(),
             volume_mounts: self.volume_mounts.clone(),
--- a/src/container_type.rs
+++ b/src/container_type.rs
@@ -1,0 +1,84 @@
+use k8s_openapi::{
+    api::core::v1::ResourceRequirements, apimachinery::pkg::api::resource::Quantity,
+};
+
+/// Describes the type of container. The types are:
+///
+/// - `Main`: This type of container runs the main appliction, like Superset or
+///    NiFi
+/// - `Init`: This type of container runs only once at the start of the product
+///    cluster and executes initial tasks like database migrations or insertion
+///    of data
+/// - `Sidecar`: Sidecar containers run alongside the main containers for
+///    additional tasks
+#[derive(Clone, Debug, Default)]
+pub enum ContainerType {
+    #[default]
+    Main,
+    Init,
+    Sidecar,
+}
+
+impl From<ContainerType> for ResourceRequirements {
+    fn from(container_type: ContainerType) -> Self {
+        Self::from(&container_type)
+    }
+}
+
+impl From<&ContainerType> for ResourceRequirements {
+    fn from(container_type: &ContainerType) -> Self {
+        match container_type {
+            ContainerType::Main => ResourceRequirements {
+                limits: Some(
+                    [
+                        ("cpu".into(), Quantity("4".into())),
+                        ("memory".into(), Quantity("4Gi".into())),
+                    ]
+                    .into(),
+                ),
+                requests: Some(
+                    [
+                        ("cpu".into(), Quantity("2".into())),
+                        ("memory".into(), Quantity("2Gi".into())),
+                    ]
+                    .into(),
+                ),
+                ..Default::default()
+            },
+            ContainerType::Init => ResourceRequirements {
+                limits: Some(
+                    [
+                        ("cpu".into(), Quantity("10m".into())),
+                        ("memory".into(), Quantity("128Mi".into())),
+                    ]
+                    .into(),
+                ),
+                requests: Some(
+                    [
+                        ("cpu".into(), Quantity("10m".into())),
+                        ("memory".into(), Quantity("128Mi".into())),
+                    ]
+                    .into(),
+                ),
+                ..Default::default()
+            },
+            ContainerType::Sidecar => ResourceRequirements {
+                limits: Some(
+                    [
+                        ("cpu".into(), Quantity("2".into())),
+                        ("memory".into(), Quantity("2Gi".into())),
+                    ]
+                    .into(),
+                ),
+                requests: Some(
+                    [
+                        ("cpu".into(), Quantity("1".into())),
+                        ("memory".into(), Quantity("1Gi".into())),
+                    ]
+                    .into(),
+                ),
+                ..Default::default()
+            },
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ pub mod client;
 pub mod cluster_resources;
 pub mod commons;
 pub mod config;
+pub mod container_type;
 pub mod cpu;
 pub mod crd;
 pub mod error;


Techassi:
I incorporated the code from @sbernauer solution. Additionally the presence of CPU and memory limits is validated when building the Pod spec.

Techassi:
This PR also adds `ContainerType` which might be somewhat out of scope, but this topic should be discussed as it might prove useful in the future. Having container types attached to each container would allow us to more easily specify custom code based on the type of container, like resource quota limits or something else entirely.

Depending on the discussion, the affected code could be removed from this PR and moved to a separate PR.

src/container_type.rs
```diff
@@ -0,0 +1,84 @@
...
+/// Describes the type of container. The types are:
+///
+/// - `Main`: This type of container runs the main appliction, like Superset or
+///    NiFi
+/// - `Init`: This type of container runs only once at the start of the product
+///    cluster and executes initial tasks like database migrations or insertion
+///    of data
+/// - `Sidecar`: Sidecar containers run alongside the main containers for
+///    additional tasks
+#[derive(Clone, Debug, Default)]
```
sbernauer, line 14: If we go with the ContainerType I would make it more explicit, e.g. have
* Product (main)
* copy files (init)
* text replacement (init)
* create key and truststore (init)
* logging (sidecar)
* metrics (sidecar)
* healthcheck (sidecar)
* ...

src/container_type.rs
```diff
@@ -0,0 +1,84 @@
...
+    Sidecar,
+}
+
+impl From<ContainerType> for ResourceRequirements {
+    fn from(container_type: ContainerType) -> Self {
+        Self::from(&container_type)
+    }
+}
+
+impl From<&ContainerType> for ResourceRequirements {
```
sbernauer, line 28: We really only want to set default resources on init containers

sbernauer (changes requested):
In general I would prefer to not include the ContainerType bits in this PR

src/builder/pod/mod.rs
```diff
@@ -367,8 +394,57 @@ impl PodBuilder {
...
+        Ok(PodTemplateSpec {
+            metadata: self.metadata.clone(),
+            spec: Some(self.build_spec()?),
+        })
+    }
+
+    /// Returns if any of the containers is missing the resource quota limit
+    /// for `key`. If so, [`Some(container_name)`] is returned, otherwise
+    /// [`None`].
+    fn is_missing_limit_for(&self, key: &str) -> Option<String> {
```
sbernauer, line 425:
```suggestion
    fn check_resource_limit_set(&self, key: &str) -> OperatorResult<()> {
```

src/builder/pod/mod.rs
```diff
@@ -367,8 +394,57 @@ impl PodBuilder {
...
+            tracing::warn!("Tried building a PodTemplate for a PodBuilder with a status, the status will be ignored...");
+        }
+
+        Ok(PodTemplateSpec {
+            metadata: self.metadata.clone(),
+            spec: Some(self.build_spec()?),
+        })
+    }
+
+    /// Returns if any of the containers is missing the resource quota limit
```
sbernauer, line 422:
```suggestion
    /// Returns if any of the containers is missing the given resource limit
```

src/builder/pod/container.rs
```diff
@@ -223,14 +227,25 @@ impl ContainerBuilder {
         self
     }
 
+    /// Sets the container type.
+    pub fn container_type(&mut self, container_type: ContainerType) -> &mut Self {
```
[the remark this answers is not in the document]
Techassi: Agreed. This will be a mandatory parameter in the `ContainerBuilder::new()` function.

src/builder/pod/mod.rs
Techassi: Changed in 5621b6f

src/builder/pod/mod.rs
Techassi: Changed in 5621b6f

src/container_type.rs
Techassi: The exact variants should be discussed in the upcoming separate PR. And yes we can be more explicit about the operations the container executes.

src/container_type.rs
Techassi: Yeah makes sense!

rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
```
sbernauer, line 1: LGTM. CC @razvan because you know the templating stuff pretty good. Maybe we want to template this here.

src/builder/pod/mod.rs
```diff
@@ -161,7 +162,33 @@ impl PodBuilder {
...
 
-    pub fn add_init_container(&mut self, container: Container) -> &mut Self {
+    /// Add the given init container.
+    /// If no resources are set, we set a default limit of 10m CPU and 128Mi
+    /// memory. Request values are set to the same values as the limits.
+    pub fn add_init_container(&mut self, mut container: Container) -> &mut Self {
+        // https://github.com/stackabletech/issues/issues/368:
+        // We only set default limits on *init* containers, as they normally
+        // simply copy stuff around, do some text replacement or, at a maximum,
+        //  create a tls truststore.These operations should normally complete
```
sbernauer, line 172:
```suggestion
        // create a tls truststore.These operations should normally complete
```

src/builder/pod/mod.rs
```diff
@@ -281,12 +327,27 @@ impl PodBuilder {
...
+    ///     apimachinery::pkg::api::resource::Quantity
+    /// };
+    ///
     /// let pod = PodBuilder::new()
     ///     .metadata_default()
     ///     .add_container(
     ///         ContainerBuilder::new("container")
     ///             .unwrap()
     ///             .add_volume_mount("listener", "/path/to/volume")
+    ///             .resources(ResourceRequirements {
```
sbernauer, line 341: Nit: Maybe we can have a fn on ContainerBuilder similar to `fn set_resources(cpu_limit: Quantity, memory_limit: Quantity, cpu_request:: Option<Quantity>, memory_request:: Option<Quantity>)`

src/commons/resources.rs
```diff
@@ -318,6 +319,107 @@ impl<T, K> Into<ResourceRequirements> for Resources<T, K> {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("missing {resource_key} resource {resource_policy} for container {container_name}")]
+pub struct ResourceRequirementsPolicyError {
```
sbernauer, line 324: I struggle a bit with the word `Policy`, as it does not seem to be a k8s term (correct me if I'm wrong).
Finding an alternative is pretty hard. WDYT of simply calling it `ResourceRequirements` instead of `ResourceRequirementsPolicy` (through the PR)?

src/commons/resources.rs
```diff
@@ -318,6 +319,107 @@ impl<T, K> Into<ResourceRequirements> for Resources<T, K> {
...
+
+                if !requests.contains_key(resource) {
+                    return Err(ResourceRequirementsPolicyError {
+                        container_name: self.name.clone(),
+                        resource_key: resource.into(),
+                        resource_policy: policy,
+                    });
+                }
+            }
+            ResourceRequirementsPolicy::Claims => todo!(),
```
sbernauer, line 404: Marker to not forget.

src/commons/resources.rs
```diff
@@ -318,6 +319,107 @@ impl<T, K> Into<ResourceRequirements> for Resources<T, K> {
...
+                        resource_key: resource.into(),
+                        resource_policy: policy,
+                    });
+                }
+            }
+            ResourceRequirementsPolicy::Requests => {
+                let requests = rr
+                    .requests
+                    .as_ref()
+                    .ok_or(ResourceRequirementsPolicyError {
```
sbernauer, line 390: :+1:

sbernauer (changes requested):
Overall direction seems good.
I had the idea, that it would be good to (also?) check the resources in the cluster resources code. 
I think we should extend the `ClusterResource` trait https://github.com/stackabletech/operator-rs/blob/5478d90da51ee099360f2ccd003119f62b56c806/src/cluster_resources.rs#L46 with a fn to get the *optional* PodSpec and implement it for Job, DaemonSet and Statefulset.
Something like 
```
    fn pod_spec(self) -> Option<PodSpec> {
        None
    }
```
Afterwards than make assertions on the Pod here https://github.com/stackabletech/operator-rs/blob/5478d90da51ee099360f2ccd003119f62b56c806/src/cluster_resources.rs#L406.

src/commons/resources.rs
Techassi: The thing is that the term `ResourceRequirements` already exists (Imported from `k8s_openapi` and used across `operator-rs`. The term describes the complete struct, which contains the fields `requests`, `limits` and `claims`. One alternative name could be `ResourceRequirementsType`.

src/commons/resources.rs
Techassi: How do we handle claims? Do we need to do any validation here?

rust-toolchain.toml
razvan: This is already in operator-templating.

razvan (commented):
Sorry for not commenting sooner. Now that I have some experience with failed tests due to resource constraints I find that failing to create pods /statefulsets/etc. when resource requests are missing is confusing:

* The error messages get lost in the reconciliation log stream
* As a user/admin I watch K8S objects and I am surprised when they are not there.
* I have to watch logs for these errors explicitly and there is no good/easy  tooling for that now.
* In case of such failures, K8S persists the message in the status of these objects. I can use kubectl to inspect them.

TL;DR: I propose to not fail when resource requests are missing. but maybe just warn.

razvan:
One other reason not to fail in this case is because it feels like such a special case and inconsistent.

There so many other reasons why creating K8S objects (pods/statefulsets) is not "semantically correct" yet we still create them.

razvan:
And one last reason why I don't like this approach: the status fields of the clusters is not updated to explain why the cluster is in an "incomplete" state.

src/commons/resources.rs
sbernauer: Nope, I'm not aware of any claim usages. I would suggest to not support it for now

rust-toolchain.toml
sbernauer: Yeah, but looking at https://github.com/stackabletech/operator-templating/blob/main/repositories.yaml only for the product operator, is that correct? We are in operator-rs here.
And: If the answer is "Not worth the effort" that's totally fine by me.

src/commons/resources.rs
Techassi: So should we remove it from the enum completely or should we keep the `todo!()` macro (or print out an error)

src/builder/pod/mod.rs
Techassi: Agreed, will implement this.

src/commons/resources.rs
sbernauer: i would remove the enum variant

src/builder/pod/mod.rs
Techassi: Added in d48d06c

src/commons/resources.rs
Techassi: Renamed in d48d06c

src/commons/resources.rs
Techassi: Removed in d48d06c

Techassi:
Since d48d06c, the `PodBuilder` will only log a warning when building the Pod / PodSpec and resource limits.

Techassi:
> Overall direction seems good. I had the idea, that it would be good to (also?) check the resources in the cluster resources code. I think we should extend the `ClusterResource` trait
> 
> https://github.com/stackabletech/operator-rs/blob/5478d90da51ee099360f2ccd003119f62b56c806/src/cluster_resources.rs#L46
> with a fn to get the _optional_ PodSpec and implement it for Job, DaemonSet and Statefulset.
> Something like
> 
> ```
>     fn pod_spec(self) -> Option<PodSpec> {
>         None
>     }
> ```
> 
> Afterwards than make assertions on the Pod here
> 
> https://github.com/stackabletech/operator-rs/blob/5478d90da51ee099360f2ccd003119f62b56c806/src/cluster_resources.rs#L406
> .

Do we really need another check here? All Pods/PodSpecs built by the `PodBuilder` should contain resource requirements, and if not, a warning will be logged.

rust-toolchain.toml
Techassi: Okay how should we proceed then? I think manually updating this should be easy, but it should probably be documented (or be part of the post/pre-release tasks).

src/builder/pod/container.rs
```diff
@@ -223,6 +223,104 @@ impl ContainerBuilder {
         self
     }
 
+    /// Sets the CPU limit and optionally request quantity for this container.
+    /// This overwrites previously defined quantitities for both limit and
+    /// request. Internally this uses [`self.with_resource`].
+    pub fn with_cpu(&mut self, limit: Quantity, request: Option<Quantity>) -> &mut Self {
```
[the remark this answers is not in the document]
Techassi: Actually, I'm experimenting with such a builder right now. Regarding the misuse from above: Separating the CPU limit from request (with two builder methods) would make this more clear in my opinion.

sbernauer:
> Do we really need another check here? All Pods/PodSpecs built by the PodBuilder should contain resource requirements, and if not, a warning will be logged.

Nope, current state is fine for me :+1: It's only about warnings, nothing more

src/builder/pod/container.rs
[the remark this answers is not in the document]
Techassi: Implemented in b81eeb6

Techassi:
I implemented the `ResourceRequirementsBuilder` which should support all the stuff we need right now. Next step would to add some doc comments. The PR should then be ready to merge.

src/builder/pod/mod.rs
```diff
@@ -367,7 +437,62 @@ impl PodBuilder {
...
+    }
+
     fn build_spec(&self) -> PodSpec {
+        // We don't hard error here, because if we do, the StatefulSet (for
+        // example) doesn't show up at all. Instead users then need to comb
+        // through the logs to find the error. That's why we opted to just
+        // throw a warning which will get displayed in the Kubernetes
+        // status.
+
+        if let Err(err) = self.check_container_resource(ResourceRequirementsType::Limits, "cpu") {
```
sbernauer, line 487: I would rather build the PodSpec, and than use
```
impl ResourceRequirementsExt for PodSpec {
    fn check_resource_requirement(
        &self,
        rr_type: ResourceRequirementsType,
        resource: &str,
    ) -> Result<(), ResourceRequirementsError> {
        for container in &self.containers {
            container.check_resource_requirement(rr_type, resource)?
        }

        Ok(())
    }
}
```

We also need to check this in `ClusterResource::add`, otherwise it was useless to extend the Trait ;)
```
        if let Some(pod_spec) = resource.pod_spec() {
            if let Err(err) =
                pod_spec.check_resource_requirement(ResourceRequirementsType::Limits, "cpu")
            {
                warn!("{}", err)
            }
            if let Err(err) =
                pod_spec.check_resource_requirement(ResourceRequirementsType::Limits, "memory")
            {
                warn!("{}", err)
            }
        }
```

src/builder/pod/mod.rs
```diff
@@ -367,7 +437,62 @@ impl PodBuilder {
         self
     }
 
+    /// Consumes the Builder and returns a constructed [`Pod`]
+    pub fn build(&self) -> OperatorResult<Pod> {
```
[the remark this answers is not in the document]
Techassi: Yes I moved the private method `build_spec` to the end of the `impl` block. I like grouping public and private methods and usually I put private methods at the end.

src/commons/resources.rs
```diff
@@ -318,6 +319,123 @@ impl<T, K> Into<ResourceRequirements> for Resources<T, K> {
...
+        &self,
+        rr_type: ResourceRequirementsType,
+        resource: &str,
+    ) -> Result<(), ResourceRequirementsError>;
+
+    /// Returns wether the implementor has a [`ResourceRequirementsType`] set
+    /// for a `resource`.
+    fn has_resource_requirement(&self, rr_type: ResourceRequirementsType, resource: &str) -> bool {
+        self.check_resource_requirement(rr_type, resource)
+            .map_or(false, |_| true)
```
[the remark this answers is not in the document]
Techassi: Changed in 18c98da

src/builder/resources.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl ResourceRequirementsBuilder<state::Initial> {
+    pub fn new() -> Self {
+        ResourceRequirementsBuilder::default()
+    }
+
+    pub fn with_cpu_limit(
+        self,
+        limit: Quantity,
```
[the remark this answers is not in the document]
Techassi: Yes this is nice, changed in 18c98da

src/builder/resources.rs
```diff
@@ -0,0 +1,154 @@
...
+use k8s_openapi::{
+    api::core::v1::ResourceRequirements, apimachinery::pkg::api::resource::Quantity,
+};
+
+mod state {
+    #[derive(Debug, Default)]
+    pub struct Initial {}
+    pub struct MissingCpuLimit {}
+    pub struct MissingMemLimit {}
+    pub struct Final {}
```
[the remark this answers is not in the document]
Techassi: Changed in 18c98da

src/builder/resources.rs
```diff
@@ -0,0 +1,218 @@
...
+            Some(limits)
+        } else {
+            None
+        };
+
+        let requests = if !requests.is_empty() {
+            Some(requests)
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
Techassi: This pattern is also used [here](https://github.com/stackabletech/operator-rs/blob/main/src/commons/resources.rs#L302-L317). I just adjusted the ordering, otherwise leaving as is.

src/builder/pod/container.rs
```diff
@@ -223,6 +218,11 @@ impl ContainerBuilder {
         self
     }
 
+    pub fn with_resources(&mut self, resources: ResourceRequirements) -> &mut Self {
```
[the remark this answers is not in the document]
Techassi: I feel like builders should prefix all methods with either `add` or `with`. This makes it super obvious that the user is dealing with a builder.

src/builder/pod/mod.rs
Techassi: Added in 597e751

src/cluster_resources.rs
```diff
@@ -58,6 +61,10 @@ pub trait ClusterResource:
     fn maybe_mutate(self, _strategy: &ClusterResourceApplyStrategy) -> Self {
         self
     }
+
+    fn pod_spec(self) -> Option<PodSpec> {
```
sbernauer, line 65: Maybe this helps avoiding unnecessary clones introduced in 597e751e37540dd7d4c3192b7d786043048f0f9c
```suggestion
    fn pod_spec(&self) -> Option<PodSpec> {
```

Techassi: 3fc8f1d takes `&self` and clones inside the method.

src/builder/resources.rs
```diff
@@ -0,0 +1,218 @@
+use std::{collections::BTreeMap, marker::PhantomData};
```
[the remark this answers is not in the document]
Techassi: Moved in 3fc8f1d

src/builder/pod/mod.rs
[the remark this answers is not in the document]
sbernauer: Kind of makes sense ;) On the other hand having build() at the bottom aligns better with the code flow. LGTM ;)

src/cluster_resources.rs
```diff
@@ -58,6 +61,10 @@ pub trait ClusterResource:
     fn maybe_mutate(self, _strategy: &ClusterResourceApplyStrategy) -> Self {
         self
     }
+
+    fn pod_spec(&self) -> Option<PodSpec> {
```
sbernauer, line 65:
```suggestion
    fn pod_spec(&self) -> Option<&PodSpec> {
```

src/cluster_resources.rs
```diff
@@ -155,7 +162,13 @@ impl ClusterResource for Service {}
 impl ClusterResource for ServiceAccount {}
 impl ClusterResource for RoleBinding {}
 impl ClusterResource for Secret {}
-impl ClusterResource for Job {}
+
+impl ClusterResource for Job {
+    fn pod_spec(&self) -> Option<PodSpec> {
+        self.spec.clone().unwrap_or_default().template.spec
+    }
```
sbernauer, line 169:
```suggestion
    fn pod_spec(&self) -> Option<&PodSpec> {
        self.spec
            .as_ref()
            .and_then(|spec| spec.template.spec.as_ref())
    }
```

src/cluster_resources.rs
```diff
@@ -171,7 +184,12 @@ impl ClusterResource for StatefulSet {
             | ClusterResourceApplyStrategy::NoApply => self,
         }
     }
+
+    fn pod_spec(&self) -> Option<PodSpec> {
+        self.spec.clone().unwrap_or_default().template.spec
+    }
```
sbernauer, line 190:
```suggestion
    fn pod_spec(&self) -> Option<&PodSpec> {
        self.spec
            .as_ref()
            .and_then(|spec| spec.template.spec.as_ref())
    }
```

src/cluster_resources.rs
```diff
@@ -206,6 +224,10 @@ impl ClusterResource for DaemonSet {
...
```
sbernauer, line 230:
```suggestion
    fn pod_spec(&self) -> Option<&PodSpec> {
        self.spec
            .as_ref()
            .and_then(|spec| spec.template.spec.as_ref())
    }
```

sbernauer (commented):
This way we don't need to clone the PodSpec (which is quite big) for *every* reconciliation loop

src/cluster_resources.rs
sbernauer: Continued below with suggestion

src/commons/resources.rs
```diff
@@ -318,6 +319,135 @@ impl<T, K> Into<ResourceRequirements> for Resources<T, K> {
...
 
+#[derive(Debug, thiserror::Error)]
+#[error("missing {resource_key} resource {resource_type} for container {container_name}")]
+pub struct ResourceRequirementsError {
+    resource_type: ResourceRequirementsType,
+    container_name: String,
+    resource_key: String,
+}
+
+/// [`ResourceRequirementsType`] dsscribes the available resource requirement
```
sbernauer, line 330:
```suggestion
/// [`ResourceRequirementsType`] describes the available resource requirement
```

src/builder/pod/resources.rs
```diff
@@ -0,0 +1,222 @@
...
+                "setting resource '{}' directly is discouraged - use provided methods instead",
+                resource
+            );
+            return self;
+        }
+
+        let resource = resource.to_string();
+
+        if self.other.contains_key(&resource) {
+            warn!("resource '{}' already set, not overwriting", resource);
```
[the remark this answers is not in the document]
Techassi: Changed to FIXME in 73f0fe0. We should tackle this after our container purpose discussion and maybe in the PR where we introduce the Type State Pattern in the `ContainerBuilder`.

src/commons/resources.rs
Techassi: Fixed in 73f0fe0

src/builder/pod/container.rs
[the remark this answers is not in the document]
Techassi: Reverted in 59049cc

src/builder/pod/mod.rs
```diff
@@ -161,7 +164,43 @@ impl PodBuilder {
...
+        // create a tls truststore.These operations should normally complete
+        // in <= 1s, so worst-case the Pod will take 1-2s longer to start up
+        // when the default is too low. However, things are different with
+        // sidecars, where e.g. a bundle builder, metric collector or a vector
+        // log sidecar can be overloaded and slow down operations or cause
+        // missing data, e.g. metrics or logs. Hence we don't apply any defaults
+        // for sidecars, product operators have to explicitly make a decision
+        // on what the resource limits should be.
+
+        // TODO (Techassi): These defaults should not live here and should
```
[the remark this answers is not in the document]
Techassi: Changed to FIXME in 73f0fe0. We should tackle this after our container purpose discussion and maybe in the PR where we introduce the Type State Pattern in the `ContainerBuilder`.

Techassi:
> Only thing remaining is the `with_resources` discussion and adding a warning when mem request < limit

Shouldn't we add a warning when request > limit. And also for both CPU and memory?

sbernauer:
Only for memory, not for CPU. CPU can be easily throttled when the node does not have enough cpu resources. When the pod uses more memory than it requested it needs to be killed to get memory back.

src/builder/pod/mod.rs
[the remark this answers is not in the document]
sbernauer: As we skip the container purpose for now, I don't see a better place to put this (where we now we deal with init containers only) ;)
Fine for me to leave it in anyway

src/builder/pod/mod.rs
```diff
@@ -390,31 +499,40 @@ impl PodBuilder {
...
             ..PodSpec::default()
+        };
+
+        // We don't hard error here, because if we do, the StatefulSet (for
+        // example) doesn't show up at all. Instead users then need to comb
+        // through the logs to find the error. That's why we opted to just
+        // throw a warning which will get displayed in the Kubernetes
+        // status. Additionally the Statefulset will have events describing the
+        // actual problem.
+
```
sbernauer, line 510: I would prefer something like this, as this is more compact (same in cluster_resources)
```
        pod_spec
            .check_resource_requirement(ResourceRequirementsType::Limits, "cpu")
            .unwrap_or_else(|err| warn!("{err}"));
        pod_spec
            .check_resource_requirement(ResourceRequirementsType::Limits, "memory")
            .unwrap_or_else(|err| warn!("{err}"));
        pod_spec
            .check_limit_to_request_ratio(&ComputeResource::Cpu, LIMIT_REQUEST_RATIO_CPU)
            .unwrap_or_else(|err| warn!("{err}"));
        pod_spec
            .check_limit_to_request_ratio(&ComputeResource::Memory, LIMIT_REQUEST_RATIO_MEMORY)
            .unwrap_or_else(|err| warn!("{err}"));
```

src/commons/resources.rs
```diff
@@ -318,17 +326,210 @@ impl<T, K> Into<ResourceRequirements> for Resources<T, K> {
...
+}
+
+/// [`ResourceRequirementsType`] describes the available resource requirement
+/// types. The user can set limits, requests and claims. This enum makes it
+/// possible to check if containers set one or more of these types.
+#[derive(Copy, Clone, Debug, Display, PartialEq, Eq, PartialOrd, Ord)]
+#[strum(serialize_all = "lowercase")]
+pub enum ResourceRequirementsType {
+    Limits,
+    Requests,
```
sbernauer, line 357:
```suggestion
    Requests,
    Limits,
```

sbernauer (changes requested):
LGTM % final comments

src/commons/resources.rs
Techassi: Changed in f5a7a7c

src/builder/pod/mod.rs
Techassi: Changed in f5a7a7c

sbernauer: Please also for cluster_resources code :)

sbernauer (approved):
LGTM % last comment (also change cluster_resources)
Thanks for driving this!

src/builder/pod/mod.rs
Techassi: Updated `cluster_resources` in e57a117

PR #600, opened 2023-06-02
Add support for container types

Techassi:
Implements container types discussed in #598


Techassi:
We concluded that we currently do not need container purposes and decided to maybe implement this in the future if the need for it arises.

PR #601, opened 2023-06-13
[Merged by Bors] - Add podOverrides to common struct CommonConfiguration

sbernauer:
## Description

Needed for https://github.com/stackabletech/issues/issues/346

## Review Checklist

### DIFF
--- a/src/role_utils.rs
+++ b/src/role_utils.rs
@@ -93,12 +93,14 @@ use crate::{
     product_config_utils::Configuration,
 };
 use derivative::Derivative;
-use k8s_openapi::apimachinery::pkg::apis::meta::v1::LabelSelector;
+use k8s_openapi::{
+    api::core::v1::PodTemplateSpec, apimachinery::pkg::apis::meta::v1::LabelSelector,
+};
 use kube::{runtime::reflector::ObjectRef, Resource};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, JsonSchema, PartialEq, Serialize)]
 #[serde(
     rename_all = "camelCase",
     bound(deserialize = "T: Default + Deserialize<'de>")
@@ -117,6 +119,8 @@ pub struct CommonConfiguration<T> {
     // BTreeMap to keep some order with the cli arguments.
     #[serde(default)]
     pub cli_overrides: BTreeMap<String, String>,
+    #[serde(default)]
+    pub pod_overrides: PodTemplateSpec,
 }
 
 fn config_schema_default() -> serde_json::Value {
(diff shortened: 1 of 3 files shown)


nightkr (approved):
I might add a callout to the changelog that every operator must support this when upgrading.

lfrancke:
I'm not blocking this but to me it's totally non-obvious how to use this and what takes priority.
As in: Can users really "override" stuff that our operators set or can they just set things that weren't possible so far?

And do our operators have a way of _forcing_ something (i.e. overriding the podOverrides)?

src/role_utils.rs
```diff
@@ -117,6 +119,32 @@ pub struct CommonConfiguration<T> {
     // BTreeMap to keep some order with the cli arguments.
     #[serde(default)]
     pub cli_overrides: BTreeMap<String, String>,
+    #[serde(default)]
+    #[schemars(schema_with = "pod_overrides_schema")]
```
chaicesan, line 123: I'm still quite new to rust. What does adding this line here cause to happen? It's not quite clear to me how it's making the containers field not mandatory. Does it depend on what the `gen` parameter in `pod_overrides_schema` ends up being?

sbernauer:
> I might add a callout to the changelog that every operator must support this when upgrading.

Good point, added

sbernauer:
> I'm not blocking this but to me it's totally non-obvious how to use this and what takes priority.

Documented the behavior in https://github.com/stackabletech/documentation/pull/415/commits/05ff37d7164087489a50082a97cfd66be58bfd8e

src/role_utils.rs
sbernauer: That's bit of magic. https://github.com/Arnavion/k8s-openapi/ takes our custom CRD struct (e.g. TrinoCluster) and automatically generates a CRD schema that can be used for k8s. Which is pretty convenient, as we can model our CRDs in rust code and let k8s-openapi do all the hard work of creating a k8s schema.
To make things even more convenient k8s-openapi offers read-to-use schemas for all the k8s core types (such as PodTemplateSpec).
E.g. the schema for PodSpec is coming from here: https://github.com/Arnavion/k8s-openapi/blob/136f8ae2195f6b72567b3fcce51b5d1572cbfe39/src/v1_27/api/core/v1/pod_spec.rs#L708
The problem is that the schema has containers listed as required here https://github.com/Arnavion/k8s-openapi/blob/136f8ae2195f6b72567b3fcce51b5d1572cbfe39/src/v1_27/api/core/v1/pod_spec.rs#L1208
So we basically take the schema of PodSpec, modify it slightly (remove the required fields), and than use it for our CRD.

sbernauer: The `gen`attribute actually doesn't matter too much, I don't even know what purpose it serves ;) We just pass it along to the `PodTemplateSpec::json_schema` function

nightkr: So a few notes here. The CRDs include a [JSON Schema](https://json-schema.org/) representation of the types, which the K8s API server validates the submitted objects against. `kube::CustomResource` uses the [`schemars::JsonSchema`](https://docs.rs/schemars/0.8.12/schemars/trait.JsonSchema.html) trait to generate this portion of the CRD. `#[schemars(schema_with)]` lets us override the schema used for a specific field, in this case to prevent some fields from being mandatory.

nightkr: As for [`gen`](https://docs.rs/schemars/0.8.12/schemars/gen/struct.SchemaGenerator.html), it's machinery used by schemars to carry configuration, and to be able to share certain kinds of global context (like reusing type definitions, though that's not allowed by K8s anyway).

PR #608, opened 2023-06-16
[Merged by Bors] - feat: add Display for ClusterCondition

Techassi:
This PR adds a `Display` impl for `ClusterCondition`. It might be worth a thought that we want to provide different `display_*` methods for various levels of verbosity.

### DIFF
--- a/src/status/condition/mod.rs
+++ b/src/status/condition/mod.rs
@@ -591,4 +591,82 @@ mod test {
         assert_eq!(got.status, expected.status);
         assert_eq!(got.message, expected.message);
     }
+
+    #[test]
+    fn test_display_short() {
+        let condition = ClusterCondition {
+            type_: ClusterConditionType::Available,
+            status: ClusterConditionStatus::False,
+            message: Some("This should not be displayed".into()),
+            ..Default::default()
+        };
+
+        assert!(!condition.is_good());
+        assert_eq!(condition.display_short(), "Unavailable".to_string());
+
+        let condition = ClusterCondition {
+            type_: ClusterConditionType::Available,
+            status: ClusterConditionStatus::True,
+            message: Some("This should not be displayed".into()),
+            ..Default::default()
+        };
+
+        assert!(condition.is_good());
+        assert_eq!(condition.display_short(), "Available".to_string());
+    }
+
+    #[test]
+    fn test_display_long() {
+        let condition = ClusterCondition {
+            type_: ClusterConditionType::Available,
+            status: ClusterConditionStatus::False,
+            message: Some("This should be displayed".into()),
+            ..Default::default()
+        };
+
+        assert!(!condition.is_good());
+        assert_eq!(
+            condition.display_long(),
+            "Unavailable: This should be displayed".to_string()
+        );
+
+        let condition = ClusterCondition {
+            type_: ClusterConditionType::Available,
+            status: ClusterConditionStatus::True,
+            message: Some("This should be displayed".into()),
+            ..Default::default()
+        };
+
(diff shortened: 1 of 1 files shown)


src/status/condition/mod.rs
```diff
@@ -108,6 +108,58 @@ pub struct ClusterCondition {
...
+    /// Returns a short display string. This method wraps the
+    /// [`std::fmt::Display`] implementation of the [`ClusterCondition`].
+    pub fn display_short(&self) -> String {
+        self.to_string()
+    }
+
+    /// Returns a long display string. This method uses the
+    /// [`std::fmt::Display`] implementation of the [`ClusterCondition`] and
+    /// combines it with the optional message to provide more context.
+    pub fn display_long(&self) -> String {
```
sbernauer, line 155: WDYT of a fn `display_long_only_on_err(&self) -> String`, that only appends the message when the condition is bad - which depends on condition?

Techassi: Add the method `display_short_or_long` in bbaa7af. This either returns the long string when there is an error, or the short version when there is no error.

sbernauer: I think we can not simply say "True is good, False is bard". We need to match the type I guess.

Techassi: Yeah true. I feel like `ClusterConditions` are designed rather weird. I will come up with a solution which decides failure case based on the type and status.

Techassi: I added an `is_error` method which determines if the condition is an error based on the type and status.

src/status/condition/mod.rs
```diff
@@ -108,6 +108,98 @@ pub struct ClusterCondition {
...
+            },
+        };
+
+        out.fmt(f)
+    }
+}
+
+impl ClusterCondition {
+    /// Returns if the [`ClusterCondition`] is considered to be an error.
+    pub fn is_error(&self) -> bool {
```
sbernauer, line 147: Like this method! Im just struggling to find a better word than "error" (as I would not consider startup a error). Maybe we want to rename it to is_good() and invert the logic

src/status/condition/mod.rs
```diff
@@ -108,6 +108,98 @@ pub struct ClusterCondition {
...
+    /// [`std::fmt::Display`] implementation of the [`ClusterCondition`] and
+    /// combines it with the optional message to provide more context.
+    pub fn display_long(&self) -> String {
+        match &self.message {
+            Some(message) => format!("{}: {}", self, message),
+            None => self.to_string(),
+        }
+    }
+
+    /// Returns either a short ot long display string, This method additionally
```
sbernauer, line 188: Please also update the docs to match the new solution

Techassi: Updated doc comment in 9d3bfa9

src/status/condition/mod.rs
```diff
@@ -108,6 +108,98 @@ pub struct ClusterCondition {
...
+        };
+
+        out.fmt(f)
+    }
+}
+
+impl ClusterCondition {
+    /// Returns if the [`ClusterCondition`] is considered to be in a good /
+    /// healthy state.
+    pub fn is_good(&self) -> bool {
```
sbernauer, line 148: Just wanted to make sure it's explicitly made public. No need to comment ;)

sbernauer (approved):
Thx!

src/status/condition/mod.rs
Techassi: Yes. This allows outside code (for example `stackablectl`) to query this information. This is i.e. used for the colored output.

PR #609, opened 2023-06-21
[Merged by Bors] - fix: Display for ReconciliationPaused

Techassi:
The display string for the `ReconciliationPaused` was swapped. This fixes the display impl introduced in #608.


sbernauer (approved):
bors r+

PR #610, opened 2023-06-22
[Merged by Bors] - Allow users to request a secret format

nightkr:
## Description

The client side of https://github.com/stackabletech/secret-operator/pull/286

## Review Checklist


src/builder/pod/volume.rs
```diff
@@ -346,6 +360,23 @@ impl SecretOperatorVolumeSourceBuilder {
     }
 }
 
+#[derive(Clone)]
+pub enum SecretFormat {
```
lfrancke, line 364: Maybe some docs would be useful here.
To me it's not clear what those are.
Pkcs12 is the only real "format" here.

What would I get with Tls and kerberos?

nightkr: Yeah that's fair. They correspond to the secret-op formats (https://docs.stackable.tech/home/stable/secret-operator/secretclass.html#format, the secret-op PR adds TLS/PKCS#12 to the list as well).

Basically, `Tls` will give you a triple of PEM files as per K8s convention (`ca.crt`, `tls.crt`, `tls.key`), and `Kerberos` will give you a Kerberos keytab and associated `krb5.conf`.

Selecting an invalid combination (like requesting a Kerberos secretclass as PKCS#12) will fail and give you an error message.

lfrancke: Ah! Understood. Yeah, a reference back to that page or something similar would be useful.

lfrancke: Looks good to me, thank you! I'll resolve this and will let others comment on the rest

lfrancke:
FYI: There is still a conflict in this branch

nightkr:
Yeah I know, I branched off of 0.41.0 to keep the changes in zk-op more focused on what's required for this specifically.

src/builder/pod/volume.rs
```diff
@@ -346,6 +360,29 @@ impl SecretOperatorVolumeSourceBuilder {
...
+
+impl SecretFormat {
+    fn as_str(&self) -> &'static str {
+        match self {
+            SecretFormat::Tls => "tls",
+            SecretFormat::TlsPkcs12 => "tls-pkcs12",
+            SecretFormat::Kerberos => "kerberos",
+        }
+    }
+}
```
Techassi, line 384: It might be worth to impl `Display`. This would also get rid of `.as_str().to_string()` above. It might even be possible to use `strum` as we already use it in some other place IIRC.

sbernauer (approved):
LGTM % @Techassi comment on strum

sbernauer (approved):
LGTM

PR #644, opened 2023-09-05
Add PodListeners

nightkr:
## Description

This is a new metadata object written in https://github.com/stackabletech/listener-operator/pull/100, which gives external entities the same address information as the pod volume provides internally.

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


src/commons/listener.rs
```diff
@@ -103,6 +105,54 @@ pub struct ListenerStatus {
...
+#[serde(rename_all = "PascalCase")]
+pub enum AddressType {
+    Hostname,
+    #[serde(rename = "IP")]
+    Ip,
+}
+
+/// Informs users about [`Listener`] objects that are bound by a given [`Pod`].
+///
+/// This is not expected to be created or modified by users.
```
fhennig, line 124: Which component creates this object

src/commons/listener.rs
```diff
@@ -7,7 +7,9 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 #[cfg(doc)]
-use k8s_openapi::api::core::v1::{Node, Pod, Service, Volume};
+use k8s_openapi::api::core::v1::{
+    Node, PersistentVolume, PersistentVolumeClaim, Pod, Service, Volume,
+};
```
fhennig, line 12: It'd make me really happy if we had a module-level docstring as well for this file (although slightly out of scope for this PR). A general destriction of what the structs are for, which other components interact with them, and how they are related would be great. maybe also a sentence or two about the underlying motivation.

fhennig (commented):
Heyho, sorry to be a PITA about the docs, but I had some ideas for how things could be documented better.

src/commons/listener.rs
nightkr: I can try to add a bit, but most of these are also things that should probably go in the user-facing Antora docs...

nightkr: 6d33001232177474309398984ea811661ae0fa8f

fhennig: perfect, thanks :ok_hand:

src/commons/listener.rs
```diff
@@ -103,6 +133,56 @@ pub struct ListenerStatus {
...
+pub enum AddressType {
+    Hostname,
+    #[serde(rename = "IP")]
+    Ip,
+}
+
+/// Informs users about [`Listener`]s that are bound by a given [`Pod`].
+///
+/// This is not expected to be created or modified by users. It will be created by
+/// the Stackable Listener Operator when mounting the listener volume, and is always.
```
fhennig, line 153:
```suggestion
/// the Stackable Listener Operator when mounting the listener volume, and is always
```

fhennig (changes requested):

src/commons/listener.rs
nightkr: oops.. heh D:

fhennig (dismissed):
:ship:

fhennig:
Uhm technically I guess

> CRD changes approved

the changes would need to be approved?

But maybe we can already merge this as I think it'll be an uncontroversial change. But can you put it on the agenda for the arch meeting?

nightkr:
Good point, I'll add it to the meeting.

src/commons/listener.rs
```diff
@@ -103,6 +133,56 @@ pub struct ListenerStatus {
 pub struct ListenerIngress {
     /// The hostname or IP address to the [`Listener`].
     pub address: String,
+    /// The type of address (`Hostname` or `IP`).
+    pub address_type: AddressType,
```
Techassi, line 137: We could maybe combine those into a complex enum `ListenerAddress`:

```rust
pub enum ListenerAddress {
    Hostname(String),
    Ip(std::net::IpAddress),
}
```

nightkr: Yes, I was tempted to do this (and we already do this internally in secret-op). The downside is that it means that we need to either always `#[serde(flatten)]` it in, or introduce another level of nesting (breaking compatibility).

nightkr: I'm also not sure about a good naming scheme in that case.. then again, now would be the time, if ever, to do that compat break.

Techassi: Yeah I feel like the address should _itself_ contain the type/kind of address it is. Keeping track of this in two different fields seems a little weird.

Techassi: Regarding the naming: I think `ListenerAdress` seems nice, but `ListenerIngressAddress` might technically be more correct.

nightkr: §orry, in terms of naming I mean… `address.type` makes sense, but `address.address` feels very redundant. I'd like to keep it in the same field rather than split it into `address.ip` and `address.hostname`, since many consumers should be fine with either…

Techassi: Sorry, I don't quite follow. The structure would look like this:

```rust
pub struct ListenerIngress {
    pub address: ListenerIngressAddress,
    pub ports: BTreeMap<String, i32>, // I would use u32, as there are no negative ports
}

pub enum ListenerIngressAddress {
    Ip(std::net::IpAddress),
    Hostname(String),
}
```

The address could then be retrieved by `ingress.address`. We then could use match to handle both variants.

siegfriedweber: I think that @Techassi's proposal is to derive the address type from the content of the address field instead of explicitly defining it:

```yaml
# no address type field
address: 10.1.2.3                 # looks like an IPv4 address
address: 2001:db8::8a2e:370:7334  # looks like an IPv6 address
address: example.com              # looks like a hostname
```

In my opinion, we should imitate the Kubernetes API in our resource definitions as close as possible so that the user is not confronted with another Kubernetes lingo. For instance, an EndpointSlice looks as follows:

```yaml
apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
addressType: IPv4
endpoints:
  - addresses:
      - 10.1.2.3
```

The `addressType` (`IPv4`, `IPv6`, `FQDN`) specifies the content of the `addresses` field.

The current implementation of the `ListenerIngress` with `address` and `addressType` is close to it. So I would leave it as is.

src/commons/listener.rs
```diff
@@ -103,6 +133,56 @@ pub struct ListenerStatus {
...
     /// Port mapping table.
     pub ports: BTreeMap<String, i32>,
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, JsonSchema)]
```
siegfriedweber, line 142: The traits `Copy`, `Eq`, and `PartialEq` could be added to the (fieldless) enums and `Eq` and `PartialEq` to the structs. This saves us from making a new operator-rs release if the traits are needed in the future.

siegfriedweber (changes requested):

src/commons/listener.rs
nightkr: I tried to reply to this the other day, but it seems to have been eaten by the comment monster. Basically, my concern is that I specifically _don't_ want people to have to pattern match if they just want the address, since the common case "should" be to just use it in a URL or whatever.

src/commons/listener.rs
nightkr: Sure.

nightkr: df72972cd643851b85fb301f3e0da7641dbdd160

siegfriedweber (dismissed):
LGTM

siegfriedweber (dismissed):

maltesander (dismissed):
LGTM! I like the current naming approach. Lets wait for @siegfriedweber and @Techassi still?

CHANGELOG.md
```diff
@@ -6,12 +6,19 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `PodListeners` CRD ([#644]).
+
+### Changed
+
+- Derive `Eq` and `Copy` where applicable for listener CRDs ([#644]).
+
 - Add support for tls pkcs12 password to secret operator volume builder ([#645]).
```
maltesander, line 15: I think the merged screwed this up (should be under added)

siegfriedweber: fixed in c6cae866ed9121b04183c69512f171fdd5ca4d76

siegfriedweber (approved):

PR #647, opened 2023-09-07
feat: Add support for human-readable duration formats

### LINKED ISSUE #649: Tracking: Support human-readable duration format in CRDs
**I AS A USER** want to not put in random big integer numbers in CRDs such as `terminationgraceperiodseconds: 8763762374`, but instead want to have some human readable duration such as `terminationgraceperiod: 42h`.

**I AS A STACKABLE DEVELOPER** want to not implement this in all operators, so have it in operator-rs

## Requirements

1. Must: Try to stick to the k8s way of specifying durations. It is ok to omit or add more features if this makes sense. In general its always easy to add new features, and hard to remove them again.
2. Must: Having support for `s,m,h,d,w`
3. Could: Having support for `M,y` as the implementation is more complicated but we might need this for ca certificate lifetime.
4. Could: Millisecond support
5. Must NOT: Support long variants `5seconds` or `42minutes`. K8s does not support this as well
6. Should: You should be able to do all the stuff std::time::Duration supports

## Implementation decisions

* Custom struct `Duration` with inner type `std::time::Duration`
* `impl Deref for Duration` with `Target = std::time::Duration`
* operators should use the Duration struct everywhere in their code possible and prefer that over `std::time::Duration` so that, for example, log messages are consistent to the k8s way of formatting durations.

## Implementation checklists

```[tasklist]
### Tasks
- [x] https://github.com/stackabletech/operator-rs/pull/647
- [x] https://github.com/stackabletech/operator-rs/pull/654
- [ ] https://github.com/stackabletech/trino-operator/pull/474
- [ ] https://github.com/stackabletech/operator-rs/pull/671
- [ ] https://github.com/stackabletech/documentation/pull/467
```

```[tasklist]
### Operator Tasks
- [ ] https://github.com/stackabletech/airflow-operator/pull/333
- [ ] https://github.com/stackabletech/druid-operator/pull/478
- [ ] https://github.com/stackabletech/hbase-operator/pull/401
- [ ] https://github.com/stackabletech/hdfs-operator/pull/408
- [ ] https://github.com/stackabletech/hive-operator/pull/378
- [ ] https://github.com/stackabletech/spark-k8s-operator/pull/292
- [ ] https://github.com/stackabletech/superset-operator/pull/412
- [ ] https://github.com/stackabletech/zookeeper-operator/pull/733
- [ ] https://github.com/stackabletech/opa-operator/pull/483
- [ ] https://github.com/stackabletech/secret-operator/pull/306
- [ ] https://github.com/stackabletech/listener-operator/pull/115
```

Techassi:
Fixes #649. This was discussed in an arch meeting roughly two weeks ago (when this PR was opened).

It is part of the effort to unify duration formats used in our CRDs. Operators would need to use the exported `Duration` type to support human-readable duration formats. This PR adds support for human-readable durations, like `2d2h20m42s` or`15d2m2s`. The `Duration` type requirements are tracked in #649 

These changes are breaking, because:

- in some cases the type changes from `number` to `string`
- in cases where `string` is already used, the parsing might be slightly different

```[tasklist]
# Author
- [X] Changes are OpenShift compatible
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


Techassi:
In its current form, the `Duration` is not usable in const environments, thus we are unable to define constant values like here: https://github.com/stackabletech/trino-operator/blob/main/rust/crd/src/lib.rs#L124

There currently is an open [issue](https://github.com/tailhook/humantime/issues/29) upstream in `humantime`, but there seems little to no progress. We could, however, implement our own (simplified) parsing.

Techassi:
Constant value are now possible:

```rust
use stackable_operator::duration::Duration;

const DEFAULT_TIMEOUT: Duration = Duration::from_secs(3600);
```

src/duration/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let input = s.trim();
+
+        // An empty or non-ascii input is invalid
+        if input.is_empty() || !input.is_ascii() {
+            return Err(DurationParseError::InvalidInput);
+        }
+
+        // Let's split up individual parts separated by a space
+        let parts: Vec<&str> = input.split(' ').collect();
```
nightkr, line 54: The Go format is `1d2h3m`, not `1d 2h 3m`.

src/duration/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        // Parse each part as a DurationFragment and extract the final duration
+        // of each fragment in milliseconds
+        let values: Vec<u128> = parts
+            .iter()
+            .map(|p| p.parse::<DurationFragment>())
+            .map(|r| r.map(|f| f.millis()))
+            .collect::<Result<Vec<_>, DurationFragmentParseError>>()?;
```
nightkr, line 62: I think I'd rather parse each parse each fragment as a `std::time::Duration` and then sum them up (I also made this mistake in handling seconds too directly in my parser).

src/duration/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+        Ok(Self(std::time::Duration::from_millis(
+            values.iter().fold(0, |acc, v| acc + (*v as u64)),
+        )))
+    }
+}
+
+impl Display for Duration {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // If the inner Duration is zero, print out '0ms' as milliseconds
+        // is the base unit for our Duration.
```
nightkr, line 76: Seconds are the SI base unit of time, milliseconds are subdivisions of that. (Following the logic of using the smallest known unit, would we want to switch to `0ns` if we added support for that? I'd say no.

src/duration/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+        let err = Duration::from_str(input).unwrap_err();
+        assert_eq!(err, expected_err)
+    }
+
+    #[rstest]
+    #[case("15d 2m 2s")]
+    #[case("1h 20m")]
+    #[case("1m")]
+    #[case("1s")]
+    fn to_string(#[case] duration: &str) {
```
nightkr, line 304: This only tests already-canonical durations, also need to consider whether `70s` should stay that or become `1m10s` (for example).

src/duration/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+    fmt::{Display, Write},
+    num::ParseIntError,
+    ops::{Deref, DerefMut},
+    str::FromStr,
+};
+
+use derivative::Derivative;
+use schemars::JsonSchema;
+use strum::IntoEnumIterator;
+use thiserror::Error;
```
nightkr, line 25: thiserror?

src/duration/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+    ExpectedNumber,
+
+    #[error("expected character, the duration fragments must end with an alphabetic character")]
+    ExpectedCharacter,
+
+    #[error("failed to parse fragment value as integer")]
+    ParseIntError(#[from] ParseIntError),
+
+    #[error("failed to parse fragment unit")]
+    UnitParseError,
```
nightkr, line 186: These error messages could use some context on where exactly the problem is.

nightkr (changes requested):

src/duration/mod.rs
Techassi: Okay I can adjust the parsing to handle the Go format. I would then however not support the space-separated format anymore.

src/duration/mod.rs
Techassi: I can add a `From<DurationFragment> for std::time:.Duration` impl.

src/duration/mod.rs
Techassi: I will add tests for these cases.

src/duration/mod.rs
Techassi: We already use it in other places in `operator-rs`. I opted to use `thiserror`, because we want to move in that direction anyways.

src/duration/mod.rs
Techassi: Agreed, I will add some more context.

src/duration/mod.rs
Techassi: I hate it already.

src/duration/mod.rs
Techassi: Added in 9e9257d

src/duration/mod.rs
Techassi: Added in 9e9257d

src/duration/mod.rs
Techassi: Changed in 9e9257d

src/duration/mod.rs
Techassi: Changed in 9e9257d

src/duration/mod.rs
Techassi: Changed in 9e9257d

src/duration/mod.rs
nightkr: > because we want to move in that direction anyways.

We have some legacy stuff that uses it, but I'm not aware of a movement to inflict _more_ of it?

src/duration/mod.rs
nightkr: I'm fine with supporting ms if the user explicitly puts them in, I just think `0s` makes more sense as the fallback.

src/duration/mod.rs
```diff
@@ -0,0 +1,303 @@
+//! This module contains a common [`Duration`] struct which is able to parse
+//! human-readable duration formats, like `2y 2h 20m 42s` or`15d 2m 2s`. It
```
sbernauer, line 2:
```suggestion
//! human-readable duration formats, like `5s`, `24h`, `2y2h20m42s` or`15d2m2s`. It
```

src/duration/mod.rs
Techassi: Okay. Then I will re-add `ms` support, but default to `0s` for the `is_zero` case.

src/duration/mod.rs
Techassi: Changed in 58b04c3

src/duration/mod.rs
Techassi: See 58b04c3

nightkr:
Worked through a bunch of misc nits: 3282f9ecc70c76cddc91a807814af1fbae5b2e33...8ef94a7ce124c86323e0f6b42ac9c26d037d787f.

nightkr:
This also doesn't enforce unit order (unlike https://gist.github.com/nightkr/c75db0c27ac0815e4f8c41cbfed3aa61#file-crd_duration-rs-L89-L101)

Techassi:
> This also doesn't enforce unit order (unlike https://gist.github.com/nightkr/c75db0c27ac0815e4f8c41cbfed3aa61#file-crd_duration-rs-L89-L101)

Yes, that was one thing I wanted to continue working on. Had it on my personal TODO list :)

Techassi:
Commit fe1911e breaks `duration::test::parse_invalid::case_3`. This change needs to be reverted.

sbernauer:
Actual change LGTM, giving it a try with trino-op graceful shutdown

sbernauer:
```
error[E0277]: cannot multiply `{integer}` by `stackable_operator::duration::Duration`
  --> rust/operator-binary/src/operations/graceful_shutdown.rs:46:13
   |
46 |         + 2 * GRACEFUL_SHUTDOWN_GRACE_PERIOD
   |             ^ no implementation for `{integer} * stackable_operator::duration::Duration`
   |
   = help: the trait `Mul<stackable_operator::duration::Duration>` is not implemented for `{integer}`
```

sbernauer:
Looking at the generated `deploy/helm/trino-operator/crds/crds.yaml`, we have the following change I would have not expected.
```
                     gracefulShutdownTimeout:
                       default: 1h
                       description: Time period the trino workers have to gracefully shut down, e.g. `1h`, `30m` or `2d`. Consult the trino-operator documentation for details.
-                      type: string
+                      properties:
+                        nanos:
+                          format: uint32
+                          minimum: 0.0
+                          type: integer
+                        secs:
+                          format: uint64
+                          minimum: 0.0
+                          type: integer
+                      required:
+                        - nanos
+                        - secs
+                      type: object
```

sbernauer:
@Techassi as discussed, pushed the fixes, please have a look

sbernauer (dismissed):
Tested and works fine for the trino graceful shutdown, awesome!
However, I would give @nightkr the chance to review the code again

Techassi:
Yes, looks good! Waiting for final feedback from @nightkr

nightkr:
> Commit [fe1911e](https://github.com/stackabletech/operator-rs/commit/fe1911ec60b9b3c56ae4fe4df4366a90cf8d32ce) breaks `duration::test::parse_invalid::case_3`. This change needs to be reverted.

Fair, I should have fixed those test cases.

src/duration/mod.rs
```diff
@@ -178,19 +217,61 @@ impl SubAssign for Duration {
...
+    }
+}
+
+impl Div<Duration> for u32 {
+    type Output = Duration;
+
+    fn div(self, rhs: Duration) -> Duration {
+        rhs / self
+    }
+}
```
nightkr, line 250: Division is not symmetric (`x / y != y / x`).

Techassi: Yes, you are right. We should remove `impl Div<Duration> for u32`, because:

1. the implementation is wrong
2. doesn't really make sense (why would you divide a `u32` by `Duration`)

Techassi: Removed in 9a5c6da

nightkr:
> > Commit [fe1911e](https://github.com/stackabletech/operator-rs/commit/fe1911ec60b9b3c56ae4fe4df4366a90cf8d32ce) breaks `duration::test::parse_invalid::case_3`. This change needs to be reverted.
> 
> Fair, I should have fixed those test cases.

98149d7c57a381668b6c05e07bf4f27b198992b8

nightkr (approved):

sbernauer:
Awesome, thanks for working on this!

PR #556, opened 2023-02-16
[Merged by Bors] - Add affinity structs

sbernauer:
## Description

For https://github.com/stackabletech/issues/issues/323

## Review Checklist
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added (or not applicable)
- [ ] Documentation added (or not applicable)
- [ ] Changelog updated (or not applicable)
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)

PR #652, opened 2023-09-15
refactor!: Remove legacy node selector on RoleGroup

sbernauer:
## Description

Remove deprecated code from https://github.com/stackabletech/operator-rs/pull/556.
Users had enough time to migrate

```[tasklist]
### Steps
- [ ] https://github.com/stackabletech/documentation/pull/453
- [ ] Every operator has been upgraded to new operator-rs
- [ ] Grepped for `selector:` in every operator to update tests and docs
```

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- BREAKING: Remove legacy node selector on `RoleGroup` ([#652]).
```
maltesander, line 9: Is it "really" breaking if we remove deprecated stuff? If yes can you add the crd-change/breaking label?
```suggestion
- BREAKING: Remove legacy node selector on `RoleGroup` ([#652]).
```

maltesander (commented):
Thanks, LGTM! But i would bet we have it in some other operators / examples (e.g. https://github.com/stackabletech/opa-operator/blob/9c8c3bbdcf731b2f2d4f6af67bec1044c64ce56d/tests/templates/kuttl/smoke/10-install-opa.yaml.j2#L37C9-L38C23). Might be worth doing a grep on all repos?

CHANGELOG.md
sbernauer: I would consider it breaking - for users as well as our operators as it's going to break stuff. I think the deprecation is the call out that there will be a breaking change in the future.
Added the labels

sbernauer:
Regarding the grep: Totally agreed, did that on the docs repo: https://github.com/stackabletech/documentation/pull/453.
Also have added an checklist item to do it for all the operators once we bump them

CHANGELOG.md
nightkr: Agreed with @sbernauer. _Especially_ since we don't currently do a great job warning users that they're using deprecated features.

sbernauer:
Lets wait for Arch meeting

lfrancke:
@sbernauer Can you tell me what happened with this one? It has the 23.11 label but is not actually included.

fhennig:
Maybe @maltesander can chime in?

maltesander:
Yeah we can remove the label since this is not in 23.11. 
Otherwise the PR looks good, whenever i saw the labelselector in tests / examples i already removed it but probably a last check before merging this (or bumping the operators to a new tag) should be done / tracked.

sbernauer:
Yeah, PR is good to merge. I wasn't allowed to merge before because of priority reasons ;)

NickLarsenNZ (approved):
I approve the intention :+1:

PR #657, opened 2023-09-21
feat: Add Duration::from_minutes, hours and days

sbernauer:
# Description

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-- Add duration overflow check ([#665])
+- Add duration overflow check ([#665]).
+- Add `Duration::from_minutes`, `Duration::from_hours` and `Duration::from_days` ([#657]).
 
+[#657]: https://github.com/stackabletech/operator-rs/pull/657
 [#665]: https://github.com/stackabletech/operator-rs/pull/665
 
 ## [0.52.1] - 2023-10-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Add duration overflow check ([#665]).
-- Add `Duration::from_minutes`, `Duration::from_hours` and `Duration::from_days` ([#657]).
+- Add 'Duration::from_millis`, `Duration::from_minutes_unchecked`, `Duration::from_hours_unchecked` and `Duration::from_days_unchecked` ([#657]).
 
 [#657]: https://github.com/stackabletech/operator-rs/pull/657
 [#665]: https://github.com/stackabletech/operator-rs/pull/665
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Add duration overflow check ([#665]).
-- Add 'Duration::from_millis`, `Duration::from_minutes_unchecked`, `Duration::from_hours_unchecked` and `Duration::from_days_unchecked` ([#657]).
+- Add `Duration::from_millis`, `Duration::from_minutes_unchecked`, `Duration::from_hours_unchecked` and `Duration::from_days_unchecked` ([#657]).
 
 [#657]: https://github.com/stackabletech/operator-rs/pull/657
 [#665]: https://github.com/stackabletech/operator-rs/pull/665
--- a/src/duration/mod.rs
+++ b/src/duration/mod.rs
@@ -285,8 +285,10 @@ impl Duration {
 
     /// Creates a new [`Duration`] from the specified number of whole minutes.
     /// Panics if the Duration is higher than [`std::time::Duration`] supports
-    /// (which is 584.9 years). Please only use this function in constant contexts,
-    /// and not based upon user input!
+    /// (which is 584.9 years). It is recommended to only use this function in `const`
+    /// environments. It is, however, not recommended to use the function to construct
+    /// [`Duration`]s from user provided input. Instead, use [`Duration::from_str`] to
+    /// parse human-readable duration strings.
     pub const fn from_minutes_unchecked(minutes: u64) -> Self {
         let millis = match minutes.checked_mul(DurationUnit::Minutes.millis()) {
             Some(millis) => millis,
--- a/src/duration/mod.rs
+++ b/src/duration/mod.rs
@@ -15,7 +15,7 @@
 use std::{
     cmp::Ordering,
     fmt::Display,
-    num::{ParseIntError, TryFromIntError},
+    num::ParseIntError,
     ops::{Add, AddAssign, Deref, DerefMut, Div, Mul, Sub, SubAssign},
     str::FromStr,
 };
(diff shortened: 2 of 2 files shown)


sbernauer:
@Techassi @nightkr so, what is the plan here? I see the following options

1. `Duration::from_secs(15 * 24 * 60 * 60)` in operator code
2. const fn `Duration::from_days` (this PR)
3. Some macro solution, e.g. `duration!("2d")`

Both 2. and 3. would work for me.

I don't have enough Rust const knowledge, but my understanding is that Option 2. comes with a panic!() in case the Duration overflows. This is the same way `time::Duration::days(days: i64)` https://docs.rs/time/0.3.29/src/time/duration.rs.html#448 works and what we currently use e.g. in secret-op.

What would be your preferred Option?

Techassi:
> 1. `Duration::from_secs(15 * 24 * 60 * 60)` in operator code

I agree that this is quite tedious.

> 2. const fn `Duration::from_days` (this PR)

A panic is necessary, because `Result` cannot be used in constant environments (and it wouldn't make sense if it could). `std::time::Duration` will also panic when the internal `u64` would overflow. I don't have any strong opinions on re-panicing in our code. 

> 3. Some macro solution, e.g. duration!("2d")

This would need some more research (I would volunteer for that if we decide to move forward with it). Additionally, we have to maintain the macro code. This sometimes can get a little tricky, but this macros would probably be easy enough to maintain.

---

We can also choose to move forward using a "middle-ground" solution like exporting public constants like:

```rust
pub const ONE_MINUTE_IN_SECS: u64 = 60;
pub const ONE_HOUR_IN_SECS: u64 = ONE_MINUTE_IN_SECS * 60;
pub const ONE_DAY_IN_SECS: u64 = ONE_HOUR_IN_SECS * 24;
```

nightkr:
As I've written before, my primary concern is that making it too convenient to write long durations.. encourages people to use more long durations where doing so is largely an antipattern or just puts a band-aid over a bigger issue.

sbernauer:
I don't think the time periods are going to change based upon the presence or absence of an `Duration::from_XXX` function.
And e.g. `Duration::from_secs(6 * 60 * 60)` is just less readable and error prone than `Duration::from_hours(6)` (and 6 hours is a total legit time period for e.g. max query runtime, graceful shutdown period, scale down time period and much more)

lfrancke:
During todays architecture meeting we decided to go ahead with this feature (not necessarily this implementation).
We also decided that it makes more sense to add warnings at the time where this is used.
e.g. the operators can warn if a user uses a value that we consider too high.

sbernauer:
@siegfriedweber, @Techassi and I did go though the PR and reached a conclusion.
User input is handled with the Serialize function and is *not* handled by the code in this PR. The functions of this PR are now prefixed with `_unchecked` and are only meant for constants.
We don't want to merge this straight up, but give @nightkr and @fhennig a chance to comment or approve.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Add `Duration::from_minutes`, `Duration::from_hours` and `Duration::from_days` ([#657]).
```
siegfriedweber, line 9:
```suggestion
- Add 'Duration::from_millis`, `Duration::from_minutes_unchecked`, `Duration::from_hours_unchecked` and `Duration::from_days_unchecked` ([#657]).
```

siegfriedweber (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Add 'Duration::from_millis`, `Duration::from_minutes_unchecked`, `Duration::from_hours_unchecked` and `Duration::from_days_unchecked` ([#657]).
```
Techassi, line 9:
```suggestion
- Add `Duration::from_millis`, `Duration::from_minutes_unchecked`, `Duration::from_hours_unchecked` and `Duration::from_days_unchecked` ([#657]).
```

src/duration/mod.rs
```diff
@@ -242,10 +243,51 @@ impl Div<u32> for Duration {
...
+
     /// Creates a new [`Duration`] from the specified number of whole seconds.
     pub const fn from_secs(secs: u64) -> Self {
         Self(std::time::Duration::from_secs(secs))
     }
+
+    /// Creates a new [`Duration`] from the specified number of whole minutes.
+    /// Panics if the Duration is higher than [`std::time::Duration`] supports
+    /// (which is 584.9 years). Please only use this function in constant contexts,
+    /// and not based upon user input!
```
Techassi, line 259: Same for below.

```suggestion
    /// (which is 584.9 years). It is recommended to only use this function in `const`
    /// environments. It is, however, not recommended to use the function to construct
    /// [`Duration`]s from user provided input. Instead, use [`Duration::from_str`] to
    /// parse human-readable duration strings.
```

src/duration/mod.rs
```diff
@@ -242,10 +243,51 @@ impl Div<u32> for Duration {
...
+    pub const fn from_minutes_unchecked(minutes: u64) -> Self {
+        let millis = match minutes.checked_mul(DurationUnit::Minutes.millis()) {
+            Some(millis) => millis,
+            None => panic!("overflow in Duration::from_minutes"),
```
Techassi, line 263: Maybe add some more context to the panic message, like with which value (of minutes in this case) the value overflowed. This should make debugging a little more clear.

Techassi (changes requested):

src/duration/mod.rs
sbernauer: Got this on my try...
```
error[E0015]: cannot call non-const formatting macro in constant functions
   --> src/duration/mod.rs:265:75
    |
265 |             None => panic!("overflow in Duration::from_minutes by passing {minutes} minutes"),
    |                                                                           ^^^^^^^^^
    |
    = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants
    = note: this error originates in the macro `$crate::const_format_args` which comes from the expansion of the macro `panic` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Techassi: Ah... That's unfortunate. Let's leave it as is I guess.

Techassi (approved):

src/duration/mod.rs
```diff
@@ -242,10 +243,53 @@ impl Div<u32> for Duration {
...
+    }
+
     /// Creates a new [`Duration`] from the specified number of whole seconds.
     pub const fn from_secs(secs: u64) -> Self {
         Self(std::time::Duration::from_secs(secs))
     }
+
+    /// Creates a new [`Duration`] from the specified number of whole minutes.
+    /// Panics if the Duration is higher than [`std::time::Duration`] supports
+    /// (which is 584.9 years). It is recommended to only use this function in `const`
```
Techassi, line 258: We might want to update this to the correct value mentioned [here](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX).

Techassi (changes requested):

src/duration/mod.rs
sbernauer: Switched in 25a5c8a0b0c0876feb2a6e8e1aecd034a6194bf3

Techassi (approved):

siegfriedweber (approved):

PR #665, opened 2023-10-06
feat: Add duration overflow check

Techassi:
This PR adds a check when adding up different duration fragments while parsing using `Duration::from_str`. This function will now return an error, if the duration overflows.

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

### DIFF
--- a/src/duration/mod.rs
+++ b/src/duration/mod.rs
@@ -15,7 +15,7 @@
 use std::{
     cmp::Ordering,
     fmt::Display,
-    num::ParseIntError,
+    num::{ParseIntError, TryFromIntError},
     ops::{Add, AddAssign, Deref, DerefMut, Div, Mul, Sub, SubAssign},
     str::FromStr,
 };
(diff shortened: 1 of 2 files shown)


src/duration/mod.rs
```diff
@@ -53,6 +53,9 @@ pub enum DurationParseError {
 
     #[snafu(display("failed to parse fragment value as integer"))]
     ParseIntError { source: ParseIntError },
+
+    #[snafu(display("duration overflow occured while parsing {value}{unit}"))]
```
sbernauer, line 57: Can we please also include the original string in the error message?

src/duration/mod.rs
```diff
@@ -120,8 +123,13 @@ impl FromStr for Duration {
             // in 1.3.0 but u128 was only added in 1.26.0. See
             // - https://users.rust-lang.org/t/why-duration-as-from-millis-uses-different-primitives/89302
             // - https://github.com/rust-lang/rust/issues/58580
-            duration +=
-                std::time::Duration::from_millis((value * unit.millis()).try_into().unwrap());
+            match duration.checked_add(std::time::Duration::from_millis(
+                (value * unit.millis()).try_into().unwrap(),
```
sbernauer, line 127: Imagine the input `5000y`. Then `value * unit.millis()` would also fail, is that correct?
In that case we would need a `checked_mul` here as well. I'm happy to re-use the error variant and just say `The duration "<original-string, e.g. 5000y3d12h>" is too long`

src/duration/mod.rs
Techassi: Included in 25e0671

src/duration/mod.rs
Techassi: Checked in 25e0671

sbernauer (approved):
Code LGTM. Some tests would be great!

Techassi:
> Code LGTM. Some tests would be great!

Yes, adding tests would indeed be nice, however it might prove a little hard to achieve that according to `Duration::MAX`, see [here](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX).

PR #671, opened 2023-10-12
feat!: Add op impls, UTC helper function and interop with time crate for Duration

Techassi:
Part of https://github.com/stackabletech/operator-rs/issues/649

This PR adds op implementations between `Duration` and `std::time::Instant` and additionally adds the `now_utc` helper function.

It also adds interoperability with the `time` crate behind the new `time` feature. Enabling that feature will add basic math operations (`Add`, `AddAssign`, `Sub`, and `SubAssign`) between Duration and `time::OffsetDateTime`.


sbernauer (dismissed):

Cargo.toml
```diff
@@ -14,6 +14,9 @@ license.workspace = true
 edition.workspace = true
 repository.workspace = true
 
+[features]
+time = ["dep:time"]
```
sbernauer, line 18: What do you think of enabling this by default, as time is such a small crate (I guess)?

sbernauer (changes requested):

src/time/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod duration;
```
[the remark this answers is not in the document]
Techassi: You are right, makes sense!

Cargo.toml
Techassi: I made it opt-in because currently we need the interop only for one operator (secret-operator).

sbernauer: I know ;) Was just an idea to not need users to know this is behind a feature flag. But I don't care, was just an idea :)

src/time/mod.rs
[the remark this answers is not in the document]
Techassi: Updated in ef93ca3

CHANGELOG.md
```diff
@@ -21,11 +21,15 @@ All notable changes to this project will be documented in this file.
   - `ClusterCondition::is_good`
   - `ClusterOperationsConditionBuilder::new`
   - `commons::pdb::default_pdb_enabled`
-
-[#674]: https://github.com/stackabletech/operator-rs/pull/674
+- Add interoperability between the `time` crate and the `stackable_operator::time::Duration` struct. This is opt-in and
+  requires the `time` feature to be enabled. Additionally, adds `Add`, `AddAssign`, `Sub`, and `SubAssign` operations
+  between `Duration` and `std::time::INstant`. Further adds a new helper dunction `Duration::now_utc` which calculates
```
sbernauer, line 26:
```suggestion
  between `Duration` and `std::time::Instant`. Further adds a new helper function `Duration::now_utc` which calculates
```

CHANGELOG.md
```diff
@@ -21,11 +21,15 @@ All notable changes to this project will be documented in this file.
   - `ClusterCondition::is_good`
   - `ClusterOperationsConditionBuilder::new`
   - `commons::pdb::default_pdb_enabled`
-
-[#674]: https://github.com/stackabletech/operator-rs/pull/674
+- Add interoperability between the `time` crate and the `stackable_operator::time::Duration` struct. This is opt-in and
+  requires the `time` feature to be enabled. Additionally, adds `Add`, `AddAssign`, `Sub`, and `SubAssign` operations
+  between `Duration` and `std::time::Instant`. Further adds a new helper dunction `Duration::now_utc` which calculates
```
sbernauer, line 26:
```suggestion
  between `Duration` and `std::time::Instant`. Further adds a new helper function `Duration::now_utc` which calculates
```

sbernauer (changes requested):

CHANGELOG.md
Techassi: Waaay ahead of you, :) I already fixed it in dbca142

PR #672, opened 2023-10-12
refactor!: Change PodBuilder::termination_grace_period to take Duration struct

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/357

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Convert the format of the Vector configuration from TOML to YAML ([#670]).
-- BREAKING: Change `PodBuilder::termination_grace_period` to take `Duration` struct ([#672]).
+- BREAKING: Rename `PodBuilder::termination_grace_period_seconds` to `termination_grace_period`, and change it to take `Duration` struct ([#672]).
 
 [#670]: https://github.com/stackabletech/operator-rs/pull/670
 [#672]: https://github.com/stackabletech/operator-rs/pull/672
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Convert the format of the Vector configuration from TOML to YAML ([#670]).
+- BREAKING: Change `PodBuilder::termination_grace_period` to take `Duration` struct ([#672]).
 
 [#670]: https://github.com/stackabletech/operator-rs/pull/670
+[#672]: https://github.com/stackabletech/operator-rs/pull/672
 
 ## [0.54.0] - 2023-10-10
 
--- a/src/builder/pod/mod.rs
+++ b/src/builder/pod/mod.rs
@@ -30,7 +30,7 @@ use tracing::warn;
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("termination grace period is too long (got {duration}, maximum allowed is {max})", max = Duration::from_secs(i64::MAX as u64))]
-    DurationTooLong {
+    TerminationGracePeriodTooLong {
         source: TryFromIntError,
         duration: Duration,
     },
@@ -471,7 +471,7 @@ impl PodBuilder {
         let termination_grace_period_seconds = termination_grace_period
             .as_secs()
             .try_into()
-            .map_err(|err| Error::DurationTooLong {
+            .map_err(|err| Error::TerminationGracePeriodTooLong {
                 source: err,
                 duration: *termination_grace_period,
             })?;
@@ -721,7 +721,7 @@ mod tests {
         let result = pod_builder.termination_grace_period(&too_long_duration);
         assert!(matches!(
             result,
-            Err(Error::DurationTooLong {
+            Err(Error::TerminationGracePeriodTooLong {
                 source: TryFromIntError { .. },
                 duration,
             }) if duration == too_long_duration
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,8 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Convert the format of the Vector configuration from TOML to YAML ([#670]).
+- BREAKING: Change `PodBuilder::termination_grace_period` to take `Duration` struct ([#672]).
```
nightkr, line 10:
```suggestion
- BREAKING: Rename `PodBuilder::termination_grace_period_seconds` to `termination_grace_period`, and change it to take `Duration` struct ([#672]).
```

Techassi (commented):
Looks good in general. I added a small comment about the upstream change. This can be approved when the upstream decision is made.

src/error.rs
```diff
@@ -111,6 +111,9 @@ pub enum Error {
         container_name: String,
         violation: String,
     },
+
+    #[error("The duration {duration:?} is too long to process")]
+    DurationTooLong { duration: Duration },
```
nightkr, line 116: Is too long for what? Where does the duration come from? What is the limit? Where's the source error?

nightkr: If we expect the downstream operator to provide this context, make a separate error type instead of adding even more crap to the overbloated corpse that is `OperatorError`.

sbernauer: Please have another look.

src/builder/pod/mod.rs
```diff
@@ -452,12 +453,19 @@ impl PodBuilder {
...
-        termination_grace_period_seconds: i64,
-    ) -> &mut Self {
+        termination_grace_period: &Duration,
+    ) -> OperatorResult<&mut Self> {
+        let termination_grace_period_seconds = termination_grace_period
+            .as_secs()
+            .try_into()
+            .map_err(|_| Error::DurationTooLong {
+                duration: *termination_grace_period,
+            })?;
```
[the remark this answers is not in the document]
sbernauer: I fear this is a bigger thing. We have `Option<i32>` or `Option<i64>` multiple time in k8s-openapi.
E.g. the number of STS replicas is an i32 as well: https://github.com/stackabletech/hdfs-operator/blob/9830327fc493bca5f7d556ba34e04e6fa15ab587/rust/operator/src/hdfs_controller.rs#L699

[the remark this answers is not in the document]
nightkr: Yeah it'd be nice to resolve eventually but I wouldn't block on it.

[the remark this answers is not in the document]
Techassi: Yes. This is clearly a bigger issue, but fundamental issues lies within the OpenAPI spec, which doesn't define any *unsigned* integers, only signed ones (with different lengths). This can be fixed at different links in the chain:

- By K8s itself, by either specifying a `min` value of `0` or unsigned integers, by introducing a custom format or by using a JSON schema
- By k8s-openapi, which would manually fix known fields by using `u32` or `u64`. One example would be fields ending with `Seconds`.

src/builder/pod/mod.rs
```diff
@@ -25,8 +27,18 @@ use k8s_openapi::{
 };
 use tracing::warn;
 
+#[derive(Debug, thiserror::Error, PartialEq)]
+pub enum Error {
+    #[error("The duration {duration} is too long to put it into the Pods terminationGracePeriodSeconds. The maximum duration is {}", Duration::from_secs(i64::MAX as u64))]
+    DurationTooLong {
```
nightkr, line 33:
```suggestion
    TerminationGracePeriodTooLong {
```

The error message notes the field specifically, so we should do the same in code to avoid confusion.

src/builder/pod/mod.rs
```diff
@@ -25,8 +27,18 @@ use k8s_openapi::{
 };
 use tracing::warn;
 
+#[derive(Debug, thiserror::Error, PartialEq)]
+pub enum Error {
+    #[error("The duration {duration} is too long to put it into the Pods terminationGracePeriodSeconds. The maximum duration is {}", Duration::from_secs(i64::MAX as u64))]
```
nightkr, line 32:
```suggestion
    #[error("termination grace period is too long (got {duration}, maximum allowed is {max})", max = Duration::from_secs(i64::MAX as u64))]
```

src/builder/pod/mod.rs
```diff
@@ -25,8 +27,18 @@ use k8s_openapi::{
 };
 use tracing::warn;
 
+#[derive(Debug, thiserror::Error, PartialEq)]
```
nightkr, line 30:
```suggestion
#[derive(Debug, thiserror::Error)]
```

Error types are typically not `PartialEq`.

sbernauer: Yeah, that's a leftover of my assert_eq!() test, which didn't work because of `TryFromIntError`

nightkr (approved):

src/builder/pod/mod.rs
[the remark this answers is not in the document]
Techassi: I added this issue to our backlog. We will tackle this in the future.

Techassi (approved):

sbernauer:
Thanks for the review!

PR #674, opened 2023-10-12
feat: Mark various functions as const

Techassi:
This PR marks various function across `operator-rs` as `const`. This is purely a QoL change. Some downstream code can benefit from this, for example: https://github.com/stackabletech/superset-operator/blob/main/rust/crd/src/lib.rs#L39-L41


sbernauer (approved):

sbernauer:
I would leave it up to you if we want a changelog entry

Techassi:
I guess we can add one, because sadly the attribute to mark since when functions are `const` is not stabilized outside of the std library.

PR #680, opened 2023-10-27
feat!: Add OIDC AuthClass provider

sbernauer:
# Description

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/src/commons/authentication/mod.rs
+++ b/src/commons/authentication/mod.rs
@@ -63,7 +63,7 @@ impl AuthenticationClass {
 
 #[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct TlsClientUsage {
+pub struct TlsClientDetails {
     /// Use a TLS connection. If not specified no TLS will be used
     pub tls: Option<Tls>,
 }
@@ -103,7 +103,7 @@ pub enum CaCert {
     SecretClass(String),
 }
 
-impl TlsClientUsage {
+impl TlsClientDetails {
     /// This functions adds
     ///
     /// * The needed volumes to the PodBuilder
--- a/src/commons/authentication/mod.rs
+++ b/src/commons/authentication/mod.rs
@@ -5,13 +5,14 @@ pub mod tls;
 
 use crate::{
     builder::{ContainerBuilder, PodBuilder, VolumeMountBuilder},
-    commons::authentication::oidc::OidcAuthenticationProvider,
+    commons::secret_class::SecretClassVolume,
 };
+
 pub use crate::{
     client::Client,
     commons::authentication::{
-        ldap::LdapAuthenticationProvider, static_::StaticAuthenticationProvider,
-        tls::TlsAuthenticationProvider,
+        ldap::LdapAuthenticationProvider, oidc::OidcAuthenticationProvider,
+        static_::StaticAuthenticationProvider, tls::TlsAuthenticationProvider,
     },
     error::Error,
 };
@@ -21,8 +22,6 @@ use kube::CustomResource;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use strum::Display;
-
-use super::secret_class::SecretClassVolume;
 
 const SECRET_BASE_PATH: &str = "/stackable/secrets";
 
--- a/src/commons/authentication/mod.rs
+++ b/src/commons/authentication/mod.rs
@@ -3,7 +3,10 @@ pub mod oidc;
 pub mod static_;
 pub mod tls;
 
-use crate::builder::{ContainerBuilder, PodBuilder, VolumeMountBuilder};
+use crate::{
+    builder::{ContainerBuilder, PodBuilder, VolumeMountBuilder},
+    commons::authentication::oidc::OidcAuthenticationProvider,
+};
 pub use crate::{
     client::Client,
     commons::authentication::{
@@ -48,6 +51,7 @@ pub enum AuthenticationClassProvider {
     Ldap(LdapAuthenticationProvider),
     Tls(TlsAuthenticationProvider),
     Static(StaticAuthenticationProvider),
+    Oidc(OidcAuthenticationProvider),
 }
 
 impl AuthenticationClass {
--- a/src/commons/authentication/mod.rs
+++ b/src/commons/authentication/mod.rs
@@ -1,7 +1,9 @@
 pub mod ldap;
+pub mod oidc;
 pub mod static_;
 pub mod tls;
 
+use crate::builder::{ContainerBuilder, PodBuilder, VolumeMountBuilder};
 pub use crate::{
     client::Client,
     commons::authentication::{
@@ -11,10 +13,15 @@ pub use crate::{
     error::Error,
 };
 
+use k8s_openapi::api::core::v1::{Volume, VolumeMount};
 use kube::CustomResource;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use strum::Display;
+
+use super::secret_class::SecretClassVolume;
+
+const SECRET_BASE_PATH: &str = "/stackable/secrets";
 
 #[derive(Clone, CustomResource, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
 #[kube(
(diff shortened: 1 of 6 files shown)


src/commons/authentication/mod.rs
```diff
@@ -38,31 +37,98 @@ pub struct AuthenticationClassSpec {
...
+    /// A name/key which references an authentication class. To get the concrete
+    /// [`AuthenticationClass`], we must resolve it. This resolution can be
+    /// achieved by using [`ClientAuthenticationDetails::resolve_class`].
+    #[serde(rename = "authenticationClass")]
+    authentication_class_ref: String,
+
+    /// This field contains authentication provider specific configuration. It
+    /// is flattened into the final CRD.
+    ///
+    /// Use [`oidc_or_error`] to get the value or report an error to the user.
```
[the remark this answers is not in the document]
sbernauer: (Hopefully) fixed in ff778a4d6646b43390c9c44c27441d4c257b2d1c

sbernauer:
I did a review of the code others have written and the current state LGTM.
As I have written parts of the code as well someone else needs to review as well :)

src/commons/authentication/mod.rs
```diff
@@ -38,31 +37,106 @@ pub struct AuthenticationClassSpec {
...
+    /// [`AuthenticationClass`], we must resolve it. This resolution can be
+    /// achieved by using [`ClientAuthenticationDetails::resolve_class`].
+    #[serde(rename = "authenticationClass")]
+    authentication_class_ref: String,
+
+    /// This field contains authentication provider specific configuration. It
+    /// is flattened into the final CRD.
+    ///
+    /// Use [`ClientAuthenticationDetails::oidc_or_error`] to get the value or report an error to the user.
+    oidc: Option<oidc::ClientAuthenticationOptions<O>>,
```
Techassi, line 102: What happened here? This should be an enum with all supported authentication options:

```rust
pub enum ClientAuthenticationOptions {
    Ldap(ldap::ClientAuthenticationOptions),
    Oidc(oidc::ClientAuthenticationOptions),
}
```

src/commons/authentication/mod.rs
```diff
@@ -38,31 +37,106 @@ pub struct AuthenticationClassSpec {
...
+    }
+
+    pub fn authentication_class_name(&self) -> &String {
+        &self.authentication_class_ref
+    }
+
+    /// In case OIDC is configured, the user *needs* to provide some connection details,
+    /// such as the client credentials. Call this function in case the user has configured
+    /// OIDC, as it will error out then the OIDC client details are missing.
+    pub fn oidc_or_error(
```
Techassi, line 119: If we provide these kind of functions (as its super hard / impossible to solve this via types), we should do it for all methods, e.g. LDAP. But we might want to do this in a separate PR.

src/commons/authentication/mod.rs
Techassi: This is a breaking change, but I would have a really hard time pushing this OIDC-centric code. This kind of code gets forgotten and might block us to make changes in the future.

The good thing: Operators are not forced to use this common struct. It should be opt-in until we have CRD versioning in place.

src/commons/authentication/oidc.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    /// Root HTTP path of the identity provider. Defaults to `/`.
+    #[serde(default = "default_root_path")]
+    root_path: String,
+
+    /// Use a TLS connection. If not specified no TLS will be used.
+    #[serde(flatten)]
+    pub tls: TlsClientDetails,
+
+    /// If a product extracts some sort of "effective user" that is represented by a
```
Techassi, line 56: Three things:

1. This also should be in the ADR
2. Yes, ideally every product (operator) should use the same identifier cluster-wide, but can be do so in the real-world. The comment mentions hard-coded claims, so we might need to think about adding it to the client-level config instead. Or am I missing something here?
3. The wording needs to change a little, but let's first resolve 1 and 2.

src/commons/authentication/oidc.rs
```diff
@@ -0,0 +1,366 @@
...
+    #[serde(rename = "clientCredentialsSecret")]
+    pub client_credentials_secret_ref: String,
+
+    /// An optional list of extra scopes which get merged with the scopes
+    /// defined in the [`AuthenticationClass`].
+    #[serde(default)]
+    pub extra_scopes: Vec<String>,
+
+    #[serde(flatten)]
+    pub extra_fields_for_product: T,
```
Techassi, line 230: This might be a little wordy. Yes, the intention is to make this very clear to the developer. We might want to shorten it to `extra_fields`, and add doc comments instead. If we decide to keep the name as is, we still would want doc comments on that struct field.

src/commons/s3.rs
```diff
@@ -4,9 +4,12 @@
 //! Operator CRDs are expected to use the [S3BucketDef] as an entry point to this module
 //! and obtain an [InlinedS3BucketSpec] by calling [`S3BucketDef::resolve`].
 //!
-use crate::commons::{authentication::tls::Tls, secret_class::SecretClassVolume};
-use crate::error;
-use crate::{client::Client, error::OperatorResult};
+use crate::{
```
Techassi, line 7: Good to see these imports grouped, but they need to go under the third-party imports :)

src/error.rs
```diff
@@ -111,6 +111,9 @@ pub enum Error {
         container_name: String,
         violation: String,
     },
+
```
Techassi, line 114: And another one *(insert DJ Khaled meme here)*. For now this is necessary, but I started working on refactoring the `operator-rs` error situation. So that we finally have proper module level errors.

Techassi (changes requested):
First look, need to take a look again after the ADR is done.

src/commons/s3.rs
sbernauer: Fixed in 75d7cea

src/error.rs
sbernauer: Yeah, absolutely agreed. Was thinking the same while adding this error

src/commons/authentication/oidc.rs
sbernauer: As discussed: 151dcc6

src/commons/authentication/mod.rs
sbernauer: We will add such as function for ldap once we switched that to new new structure as well

src/commons/authentication/mod.rs
sbernauer: I tried my best, but could not get the enum to work...
I documented it in ab7b08edf5cc8b5a722c05f15a3b2c07e0cc35e5, would it be fine for you to continue like this for now?

src/commons/authentication/oidc.rs
sbernauer: Yeah, I think we are convinced that we need to support configuring the `principalClaim`. operators will error out when the product does not support the configured `principalClaim`.
Do you want to make suggestions regarding 3.), as 1.) is not under control of this PR? :)

NickLarsenNZ: For reference: https://github.com/stackabletech/superset-operator/blob/fbd82b6abf8ee6b9ff9d0e54d99aa17d60cbdc5b/rust/crd/src/authentication.rs#L144-L151

NickLarsenNZ: There could be a scenario where a customer already has Trino, and uses the `sub` (stable identifier) principal claim, and all rego rules that are needed. Then they want to add in Superset, which will error out if using the same AuthenticationClass.

I guess the solution would be to duplicate the AuthenticationClass, and just set the `principalClaim` accordingly.

If so, then I think we can resolve this for now, and separately discuss whether this is an AuthenticationClass level or a product level field.

src/commons/authentication/mod.rs
```diff
@@ -38,31 +37,110 @@ pub struct AuthenticationClassSpec {
...
+#[schemars(description = "")]
+pub struct ClientAuthenticationDetails<O = ()> {
+    /// A name/key which references an authentication class. To get the concrete
+    /// [`AuthenticationClass`], we must resolve it. This resolution can be
+    /// achieved by using [`ClientAuthenticationDetails::resolve_class`].
+    #[serde(rename = "authenticationClass")]
+    authentication_class_ref: String,
+
+    /// This field contains authentication provider specific configuration. It
+    /// is flattened into the final CRD.
```
Techassi, line 99: It is not flattened anymore, so we should remove that from the comment.

src/commons/authentication/oidc.rs
Techassi: Okay, then let's stick with the current approach. This needs a followup PR in https://github.com/stackabletech/documentation to adjust the ADR.

src/commons/authentication/oidc.rs
```diff
@@ -0,0 +1,368 @@
...
+        ]
+    }
+}
+
+/// An enum of supported OIDC or identity providers which can serve as a hint
+/// in the product operator. Some products require special handling of
+/// authentication related config options. This hint can be used to enable such
+/// special handling.
+#[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "PascalCase")]
```
Techassi, line 210: I feel like we should switch that to `lowercase` or `camelCase`.

Techassi (changes requested):
LGTM, only minor nits.

src/commons/authentication/oidc.rs
sbernauer: I think there is a naming convention of using PascalCase for scalar values in k8s. However I can't find it :/

src/commons/authentication/mod.rs
sbernauer: Good catch, was a leftover. Fixed in 3bad60c

src/commons/authentication/oidc.rs
Techassi: Alright, then let's keep it as is.

src/commons/authentication/oidc.rs
sbernauer: Created https://github.com/stackabletech/documentation/pull/498

PR #615, opened 2023-07-03
feat!: Add key/value pair validation

Techassi:
Fixes #614

PR #684, opened 2023-10-31
feat: Add key/value pair validation

### LINKED ISSUE #614: Add validation for key/value pairs used in various K8s resources
This issue tracks the implementation of label max length enforcement across our various builders for various K8s resources, e.g. volume names.

```[tasklist]
### Tasks
- [ ] https://github.com/stackabletech/operator-rs/pull/615
```

Techassi:
Fixes #614, supersedes #615, part of https://github.com/stackabletech/issues/issues/188

This PR adds mechanisms to safely create key/value pairs (labels and annotations). These enable us to attach a unified set of labels to deployed resources. The set of labels make it to possible to track resources installed by us, which in turn enables us to list, modify and delete resources with our management tools, namely `stackablectl`, and in the future `stackable-cockpit`.

```[tasklist]
## Tasks
- [x] Move `stackable_operator::label_selector::convert_label_selector_to_query_string` into `kvp` module
- [x] Remove all temporary `unwraps`
- [x] Add doc comments to all new structs, modules, types, etc...
```


Techassi:
We ideally want the creation of a K/V pair to be available in `const` environments. With the current approach making use of the (non-constant) `FromStr` trait, this is not possible. We might be able to sidestep this using a non-trait solution (until `const` in traits are stabilized and available). This however requires a more sophisticated parsing process, which looks at individual bytes without allocating.

Having K/V pair creation available in `const` environments would allow us to define common labels/attributes in `operator-rs` (or any other operator repository) which are parsed and validated during compile time. That would be huge, but at the moment rather cumbersome to implement.

The API is very close to it's final form. One open point is the usage of the new API in other `operator-rs` code, like the k8s resource builders. An important design decision is to make the usage through these functions convenient to use while making sure the K/V pairs are valid.

crates/operator-macros/src/lib.rs
NickLarsenNZ, line 1: Could be nice to have some doc comments to show how to use the macros

NickLarsenNZ (approved):
Looks really good

crates/operator-macros/src/lib.rs
Techassi: Added in a4e20aa

NickLarsenNZ (dismissed):
lgtm

Techassi:
This PR is now at a point, where everything should work without any breaking changes. That is due to multiple `unwraps`, which will be removed from this point on wards. Basically all previously in-fallible label/annotation builder functions will now be fallible to to the new required validation.

src/kvp/mod.rs
```diff
@@ -337,6 +348,85 @@ where
...
+                return Ok(false);
+            }
+        }
+
+        Ok(true)
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct ObjectLabels<'a, T> {
```
NickLarsenNZ, line 394: nit: Add a doc comment for what this is used for. Do they represent the existing labels on the resource?

src/builder/meta.rs
```diff
@@ -153,6 +153,9 @@ impl ObjectMetaBuilder {
     }
 
     pub fn build(&self) -> ObjectMeta {
+        // NOTE (Techassi): Shouldn't this take self instead of &self to consume
+        // the builder and build ObjectMeta without cloning?
+
```
NickLarsenNZ, line 158: I think you're right. They do the same here: https://rust-unofficial.github.io/patterns/patterns/creational/builder.html#example

src/builder/pod/volume.rs
```diff
@@ -335,7 +335,7 @@ impl SecretOperatorVolumeSourceBuilder {
             if Some(SecretFormat::TlsPkcs12) != self.format {
                 warn!(format.actual = ?self.format, format.expected = ?Some(SecretFormat::TlsPkcs12), "A TLS PKCS12 password was set but ignored because another format was requested")
             } else {
-                annotations.insert(Annotation::tls_pkcs12_password(&password).unwrap());
+                annotations.insert(Annotation::tls_pkcs12_password(password).unwrap());
```
NickLarsenNZ, line 338: Just trying to understand this. We're putting a passphrase into an annotation?
Maybe it's fine, but this can be visible in various K8s UIs, control-plane logs (maybe).

NickLarsenNZ (dismissed):
Looking good!

src/kvp/mod.rs
Techassi: Yes, this is part of the above todo list.

src/builder/meta.rs
Techassi: I guess we should address this in a follow-up PR. This is not really part of this PR and would make the PR even bigger than it already is.

I will create a tracking issue, where we can track all the technical debt / cruft which needs fixing.

src/builder/pod/volume.rs
nightkr: This is just a workaround for applications that only support opening encrypted PKCS#12 stores.

src/kvp/mod.rs
```diff
@@ -0,0 +1,546 @@
...
+    KeyValuePairParse { source: KeyValuePairError<E> },
+}
+
+/// [`KeyValuePairs`] is a list of [`KeyValuePair`]. This collection **doesn't**
+/// provide any de-duplication mechanism, meaning multiple [`KeyValuePair`]s
+/// with the same content can be present at the same time. However, converting
+/// to a [`BTreeMap<String, String>`] removes any duplicate data. Order matters
+/// in this case: later labels overwrite previous onces.
+#[derive(Clone, Debug, Default)]
+pub struct KeyValuePairs<V: ValueExt>(BTreeSet<KeyValuePair<V>>);
```
nightkr, line 207: Couldn't this be a `BTreeMap<K, V>`? Why does this need a dedicated type anyway?

src/kvp/value.rs
```diff
@@ -0,0 +1,17 @@
...
+    ops::Deref,
+    str::FromStr,
+};
+
+use serde::Serialize;
+
+/// Trait which ensures the value of [`KeyValuePair`][crate::kvp::KeyValuePair]
+/// is validated. Different value implementations should use [`FromStr`] to
+/// parse and validate the value based on the requirements.
+pub trait ValueExt:
```
nightkr, line 13:
```suggestion
pub trait Value:
```

The `Ext` prefix is used for traits that just add extension methods to other types. If it has more than one impl, it's not an `Ext` trait. If anything needs a bound on the trait, it's probably not an `Ext` trait.

crates/operator-macros/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+/// ```
+/// let annotation = annotation!("stackable.tech/vendor=Hello Wörld!");
+/// ```
+#[macro_export]
+macro_rules! annotation {
+    ($Input:literal) => {{
+        use std::str::FromStr;
+        stackable_operator::kvp::Annotation::from_str($Input)
+    }};
+}
```
nightkr, line 30: Why are these macros at all, actually?

src/kvp/value.rs
```diff
@@ -0,0 +1,17 @@
...
+    Deref<Target = String> + FromStr<Err = Self::Error> + Clone + Display + Eq + Ord + Serialize
```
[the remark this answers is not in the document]
Techassi: Yes, this can be changed to `&str`. I just didn't want to deal with lifetimes for the initial API design.

nightkr:
When would the former API be used? What hole is it trying to fill?

src/kvp/value.rs
Techassi: Yes, agreed. I will change that.

src/kvp/key.rs
```diff
@@ -0,0 +1,360 @@
...
+        let (prefix, name) = if parts.len() == 1 {
+            (None, KeyName::from_str(parts[0]).context(KeyNameSnafu)?)
+        } else {
+            (
+                Some(KeyPrefix::from_str(parts[0]).context(KeyPrefixSnafu)?),
+                KeyName::from_str(parts[1]).context(KeyNameSnafu)?,
+            )
+        };
+
+        Ok(Self { prefix, name })
```
nightkr, line 76:
```suggestion
let (prefix, name) = if let Some((prefix, name)) = input.split_once('/') {
    ensure!(!name.contains('/'), NestedPrefixSnafu);
    (Some(prefix), name)
} else {
    (None, input)
}

Ok(Self {
    prefix: prefix.map(KeyPrefix::from_str).transpose().context(KeyPrefixSnafu)?,
    name: KeyName::from_str(name).context(KeyNameSnafu)?,
})
```

nightkr: Or alternatively,

```rust
let parts = input.split('/').collect::<Vec<_>>();
match &parts[..] {
  [name] => (None, name),
  [prefix, name] = (Some(prefix), name),
  _ => return NestedPrefixSnafu.fail(),
}
```

nightkr: But this splitting also seems like a case of "why are we parsing these consts instead of just constructing valid data to begin with"?

crates/operator-macros/src/lib.rs
Techassi: A shorter way of constructing a label / annotation. The idea was that this can eventually be used in const environments like this:

```rust
pub const APP_NAME = label!("app.kubernetes.io/name=airflow");
```

nightkr:
Then we can consider the problem when it actually arises. We shouldn't preemptively make it easier to do the wrong thing just in case.

src/kvp/value.rs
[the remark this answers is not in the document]
nightkr: The types would be `Deref<Target = str>` and `&str`, no more or fewer lifetimes than in the current `String` variant.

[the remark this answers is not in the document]
Techassi: Oh sorry, my bad.Mmy mint went to the inner type being a `&str`.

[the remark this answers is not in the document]
Techassi: Changed in 139b973

crates/operator-macros/src/lib.rs
nightkr: That just depends on it being `const`. The value that we can provide would be compile-time validation... but the current API is incompatible with that anyway since it returns a `Result<Value>` rather than the `Value` itself.

src/builder/meta.rs
NickLarsenNZ: Agree, a small PR is a good PR

src/builder/pod/volume.rs
NickLarsenNZ: @Techassi good point, it is not introduced in this PR.

@nightkr, while we're here, is it a reference to where to find the PKCS12 passphrase? Or the actual secret? 
I can imagine some organisations having a problem with this (unless they have alternatives when using PKCS12 - eg: mounting as a Volume or env var from a Secret).

src/kvp/annotation/value.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::kvp::ValueExt;
+
+#[derive(Debug, PartialEq, Snafu)]
+pub enum AnnotationValueError {
+    #[snafu(display("value contains non-utf8 characters"))]
+    ValueNotUtf8 { source: std::str::Utf8Error },
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct AnnotationValue(String);
```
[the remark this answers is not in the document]
NickLarsenNZ: I think it's fine. Annotation values just need to be valid UTF-8 (from what I recall).
And it maintains the same pattern as LabelValue which needs validation.

src/builder/pod/volume.rs
nightkr: It is the actual password. The password is _only_ used to encrypt the PKCS#12 store between the secret-operator and the application. The password is not used to encrypt any secrets at rest (certificates are stored as unencrypted PEMs in `Secret`s).

src/kvp/key.rs
Techassi: Changed in b3a897e

crates/operator-macros/src/lib.rs
Techassi: Yes that depends on it being `const`, which is a whole other discussion. When we would eventually have `const` support, we could use `panic` or compile-time result evaluation which would fail compilation on `Err(_)`.

Of course we can remove these macros for this PR. I just found them neat :)

src/kvp/annotation/value.rs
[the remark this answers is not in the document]
Techassi: Removing that newtype and replacing it with a plain `String` might be challenging, as it must fulfill the `Value` trait. I can test it tho...

src/kvp/annotation/value.rs
```diff
@@ -0,0 +1,42 @@
...
+impl ValueExt for AnnotationValue {
+    type Error = AnnotationValueError;
+}
+
+impl FromStr for AnnotationValue {
+    type Err = AnnotationValueError;
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        let input = std::str::from_utf8(input.as_bytes()).context(ValueNotUtf8Snafu)?;
+        Ok(Self(input.to_owned()))
```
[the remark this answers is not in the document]
Techassi: You are right, we can skip validation and straight up return the value.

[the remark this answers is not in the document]
Techassi: Changed in 28f73c8

src/kvp/annotation/value.rs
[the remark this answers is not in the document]
Techassi: Yes, I agree with that :)

src/kvp/mod.rs
```diff
@@ -0,0 +1,546 @@
...
+/// A [`KeyValuePair`] is a pair values which consist of a [`Key`] and value.
+/// These pairs can be used as Kubernetes labels or annotations. A pair can be
+/// parsed from a string with the following format: `(<PREFIX>/)<NAME>=<VALUE>`.
+///
+/// ### Links
+///
+/// - <https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/>
+/// - <https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/>
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct KeyValuePair<V>
```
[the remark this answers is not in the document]
Techassi: We can change the `KeyValuePair` struct to a newtype struct like: `KeyValuePair<V>((Key, V))`. I like having a dedicated type with which we can provide a developer friendly API.

src/kvp/mod.rs
Techassi: I mainly wanted to have a controlled way of creating a list of `KeyValuePair`. One example is `Labels::recommended()`.

crates/operator-macros/src/lib.rs
nightkr: Compile-time validation would be the use-case I can see for macroing it.. but even then I'm honestly not sure how common it'd be to have a fully static label set?

src/kvp/mod.rs
nightkr: Couldn't that be `kvp::label::recommended()`?

src/kvp/mod.rs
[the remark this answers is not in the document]
nightkr: > I like having a dedicated type with which we can provide a developer friendly API.

What special APIs are you thinking about here?

`(K, V)` is a common Rust convention, and allows for seamless roundtripping with `HashMap`/`BTreeMap` (see [`Extend`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#impl-Extend%3C(K,+V)%3E-for-HashMap%3CK,+V,+S%3E), [`From`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#impl-From%3C%5B(K,+V);+N%5D%3E-for-HashMap%3CK,+V,+RandomState%3E), [`FromIterator`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#impl-FromIterator%3C(K,+V)%3E-for-HashMap%3CK,+V,+S%3E), [`IntoIterator`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#impl-IntoIterator-for-HashMap%3CK,+V,+S%3E), [`iter()`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.iter), etc. I'm not saying that it can't be done, but I'd like to understand what you expect to gain from doing it.

crates/operator-macros/src/lib.rs
Techassi: Yeah this is mostly useful for stuff like `stackable.tech/vendor=Stackable` or `stackable.tech/managed-by=stackablectl`.

nightkr: Yup. My question would be how often we set those without also setting the corresponding stacklet annotations.

src/kvp/mod.rs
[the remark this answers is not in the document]
Techassi: > What special APIs are you thinking about here?

For example `KeyValuePair::new()`, `kvp.key()` or `kvp.value()`. The custom type also "ecapsulates" specific behaviour and different impls like `Display` or `Serialize`.

To get the PR forward, an intermediate solution could be to implement `Deref<Target=(Key, V)> where V: Value` for `KeyValuePair`. We can then explore using a plain `(K, V)` in a future PR.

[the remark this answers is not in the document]
nightkr: > For example KeyValuePair::new(), kvp.key() or kvp.value().

So.. `(k, v)`, `kvp.0`, and `kvp.1`, respectively.

> The custom type also "ecapsulates" specific behaviour and different impls like Display or Serialize.

`Display` I can sort of understand, but what on earth do you need a non-standard `Serialize` for here? That just looks like a very special-cased workaround for abusing `BTreeSet` instead of `BTreeMap`.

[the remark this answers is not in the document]
Techassi: > So.. `(k, v)`, `kvp.0`, and `kvp.1`, respectively.

Yes, but the decision to use a dedicated type was inspired by this pattern: https://www.shuttle.rs/blog/2022/07/28/patterns-with-rust-types#contract-based-programming-in-rust--sub-typing

> `Display` I can sort of understand, but what on earth do you need a non-standard `Serialize` for here? That just looks like a very special-cased workaround for abusing `BTreeSet` instead of `BTreeMap`.

I don't know what `BTreeSet` and `BTreeMap` has to do with this? We are talking about a single `KeyValuePair`, which doesn't use any of these collection types. The serialization of the type works very similar to a map, because the identifier is part of the inner data instead the item (struct/enum) name serde uses per default.

src/kvp/mod.rs
Techassi: Which would return a `BTreeMap<Key, LabelValue>` or a `BTreeSet<KeyValuePair<LabelValue>>`?

src/kvp/mod.rs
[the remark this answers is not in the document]
nightkr: > Yes, but the decision to use a dedicated type was inspired by this pattern: https://www.shuttle.rs/blog/2022/07/28/patterns-with-rust-types#contract-based-programming-in-rust--sub-typing

That makes sense for the making sure that the keys/values are valid, but I don't see the value to treating them in isolation.

> I don't know what BTreeSet and BTreeMap has to do with this? We are talking about a single KeyValuePair, which doesn't use any of these collection types. The serialization of the type works very similar to a map, because the identifier is part of the inner data instead the item (struct/enum) name serde uses per default.

`BTreeMap<K, V>` already serializes to a map of:

```yaml
k1: v1
k2: v2
# ...
```

`BTreeSet<(K, V)>` serializes natively to `[[k1, v1], [k2, v2], ...]`, so you do this newtype dance to override both to recreate the map-style serialization. Actually, looking through it again it doesn't look like the `impl Serialize for KeyValuePair` is used as all, since `KeyValuePairs` serializes each key/value separately already?

src/kvp/mod.rs
nightkr: `BTreeMap`. `BTreeSet` doesn't have the semantics you want. For another example, saying `a=b` always takes precedence over `a=a` is just bizarre.

crates/operator-macros/src/lib.rs
nightkr: The macros as defined in this PR don't help anyone and might as well be removed. I'm not entirely against adding compile-time validated macros in the future, as long as we can minimize the maintenance burden that they add (as in: we should make sure that we can reuse the same validation infra on both sides).

Techassi: Okay, then I will remove the macros for this PR. Supporting them in the future shouldn't need much maintenance work.

Techassi: Removed in 7e6279d

src/kvp/mod.rs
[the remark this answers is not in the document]
Techassi: > That makes sense for the making sure that the keys/values are valid, but I don't see the value to treating them in isolation.

Okay, that's why I proposed to implement `Deref<Target = (Key, V)>` or having a function like `.into_inner()` which consumes self and returns `(Key, V)`. Do you have any other suggestions?

> Actually, looking through it again it doesn't look like the `impl Serialize for KeyValuePair` is used as all, since `KeyValuePairs` serializes each key/value separately already?

The `Serialize` for a single `KeyValuePair` is only used, if we want to serialize a single k/v pair, like:

```rust
#[derive(Serialize)]
struct Foo {
    bar: KeyValuePair<LabelValue>,
}
```

Sadly we cannot combine the `Serialize` impl for both a single and multiple k/v pairs.

The serde support might not be necessary, as we most likely wont directly use these structs in our CRDs. So we can opt to remove them for now.

src/kvp/mod.rs
Techassi: Added in 1e6979f.

src/kvp/mod.rs
nightkr: Or... we could just use the collection directly, and not reinvent the wheel.

src/kvp/mod.rs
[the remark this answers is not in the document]
Techassi: I removed serde support for now.

src/kvp/mod.rs
Techassi: @NickLarsenNZ and I decided to move forward as is. We can refine the feature in later PRs. I already have some stuff in mind :)

NickLarsenNZ (approved):
LGTM. I like the contains and contains_key that take an impl TryInto. Makes the usage really clean.

PR #701, opened 2023-12-15
refactor: Add delegated function impls for `Labels` and `Annotations`

Techassi:
Part of https://github.com/stackabletech/issues/issues/188

~Final merge is blocked until https://github.com/Kobzol/rust-delegate/pull/64 is officially released.~ Version `0.12.0` was released on 2023-12-22.


sbernauer (dismissed):
Whup whup

src/kvp/label/mod.rs
```diff
@@ -305,4 +278,25 @@ impl Labels {
 
         Ok(labels)
     }
+
+    delegate! {
```
lfrancke, line 282: Maybe it's just me but it's not obvious what this is doing. I have no idea what "delegate!" is.
If in doubt, add a comment

Techassi: Sure. I will add a comment. It basically forwards the impl to the inner type.

lfrancke: You don't have to start a new PR for this.
It was more a general remark: Help your future self and other future people with logging and comments.

Comments should (in my opinion) mostly about the "why" not the "what", sometimes - if it's not obvious - about the "how this works"

Techassi: This should now be resolved by 20274e7

sbernauer (approved):

lfrancke (approved):

PR #702, opened 2023-12-18
Add operator_image argument to ProductOperatorRun

NickLarsenNZ:
# Description

Add the operator_image argument to ProductOperatorRun for all operators to use.
This change came from this comment chain: https://github.com/stackabletech/opa-operator/pull/433#discussion_r1371496911

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -197,6 +197,8 @@ pub struct ProductOperatorRun {
     /// Tracing log collector system
     #[arg(long, env, default_value_t, value_enum)]
     pub tracing_target: TracingTarget,
+    #[arg(long, env)]
+    pub operator_image: String,
 }
 
 /// A path to a [`ProductConfigManager`] spec file
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -274,6 +274,7 @@ mod tests {
     const DEPLOY_FILE_PATH: &str = "deploy_config_spec_properties.yaml";
     const DEFAULT_FILE_PATH: &str = "default_file_path_properties.yaml";
     const WATCH_NAMESPACE: &str = "WATCH_NAMESPACE";
+    const OPERATOR_IMAGE: &str = "OPERATOR_IMAGE";
 
     #[test]
     fn verify_cli() {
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -197,6 +197,10 @@ pub struct ProductOperatorRun {
     /// Tracing log collector system
     #[arg(long, env, default_value_t, value_enum)]
     pub tracing_target: TracingTarget,
+    /// The full name of the operator image. This env var is set by the helm-charts and should
```
sbernauer, line 200: Nit: Please have one empty newline as spacing between all this struct fields

sbernauer (changes requested):
Please add a changelog entry similar to

BREAKING: Add mandatory operator argument `--operator-image`, which can also be set via the `OPERATOR_IMAGE` env var

NickLarsenNZ:
> Please add a changelog entry similar to
> 
> BREAKING: Add mandatory operator argument `--operator-image`, which can also be set via the `OPERATOR_IMAGE` env var

Added in dcfe39407b3cc0ae4f887a35afea1d3ecd5e80ca

src/cli.rs
NickLarsenNZ: Yes, I prefer that too, but followed what was already done

Updated in b3ed2a06a5afa6b016915d462850ed3d93edab61

sbernauer: Thanks!
He have decided on the spaces in an Arch meeting and documented it in our code style guideline (same sheet as arch meeting agenda) :)

sbernauer (approved):
Thanks!

fhennig:
I see this flag is set by the Helm Chart. But can we also document why the operator needs to know the image name?

Also, is this something that the user should be able to override? if so this needs to be documented as well.

nightkr:
> I see this flag is set by the Helm Chart. But can we also document why the operator needs to know the image name?

It's required when the operator needs to deploy itself as a sidecar (currently used by the in-progress OPA UIF).

> Also, is this something that the user should be able to override? if so this needs to be documented as well.

I'd say not specifically, the way to override this would be to specify the custom image to the helm chart (which would bubble down to this).

NickLarsenNZ:
I'm closing this PR as it is a premature change. We can do it when a second operator needs to know the image.

PR #705, opened 2024-01-03
feat: add ldap endpoint helper 

NickLarsenNZ:
… to liberate each operator from having to do it

# Description

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/src/commons/authentication/ldap.rs
+++ b/src/commons/authentication/ldap.rs
@@ -2,6 +2,7 @@ use k8s_openapi::api::core::v1::{Volume, VolumeMount};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use snafu::{ResultExt, Snafu};
+use url::{ParseError, Url};
 
 use crate::{
     builder::{ContainerBuilder, PodBuilder, VolumeMountBuilder},
@@ -11,12 +12,17 @@ use crate::{
     },
 };
 
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
 #[derive(Debug, Snafu)]
-pub enum AuthenticationProviderError {
+pub enum Error {
     #[snafu(display(
         "failed to convert bind credentials (secret class volume) into named Kubernetes volume"
     ))]
     BindCredentials { source: SecretClassVolumeError },
+
+    #[snafu(display("failed to parse LDAP endpoint url"))]
+    ParseLdapEndpointUrl { source: ParseError },
 }
 
 #[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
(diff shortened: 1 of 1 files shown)


src/commons/authentication/ldap.rs
```diff
@@ -49,6 +55,22 @@ pub struct AuthenticationProvider {
...
+    pub fn endpoint_url(&self) -> Result<Url> {
+        let url = Url::parse(&format!(
+            "{protocol}{server_hostname}:{server_port}",
+            protocol = match self.tls.tls {
+                None => "ldap://",
+                Some(_) => "ldaps://",
+            },
+            server_hostname = self.hostname,
+            server_port = self.port()
+        ))
```
Techassi, line 68: Nit: Is there any particular reason we first assign `self.*` to a formatting variable. I don't know if there is any difference performance-wise, but we might want to think about a rule to add to our style guide anyways...

Techassi (approved):

sbernauer (approved):
Thanks!

src/commons/authentication/ldap.rs
sbernauer: Do I understand you correctly that you want anonymous formatting arguments?
I would have expected no performance difference and would slightly prefer named arguments for readability

Techassi: I would expect no performance difference either, but you never know. I'm happy leaving it as is (and yes, named arguments are better for readability), but I then would prefer that we write this down in our code style guide :)

Techassi: I will create an Arch Meeting item for this. I might also add a PR in our https://github.com/stackabletech/documentation repo to have the code style guide visible in our public documentation.

sbernauer: Agreed :+1:

NickLarsenNZ: I'm not sure I understood what you meant, but it sounded like you are talking about inlining?

You can't (AFAIK) do `format!("{self.hostname}")` for example.

Did you mean to declare variables and then inline them?

```rust
let protocol = match ... {};
let server_hostname = self.hostname;
let server_port = self.port;

let url = Url::parse(&format!("{protocol}{server_hostname}:{server_port}"))
// ...
```

NickLarsenNZ: Ah I see. It's just about named vs unnamed. I think named is best, so that order doesn't matter and there is less chance of accidents.

But I also prefer inlining when possible (no extra args).

sbernauer: Just had a look and commented on your Arch meeting entry @Techassi

PR #711, opened 2024-01-11
feat: Add key/value pair QoL improvements

Techassi:
This PR adds two QoL improvements to the key/value pair API:

- Add `TryFrom<[(K, V); N]>` impl for labels and annotations
- Add `Labels::parse_insert()` and `Annotations::parse_insert()` functions. They take `impl TryInto<_>` as a parameter.

### DIFF
--- a/src/kvp/annotation/mod.rs
+++ b/src/kvp/annotation/mod.rs
@@ -189,7 +189,7 @@ impl Annotations {
 
     /// Tries to insert a new annotation by first parsing `annotation` as an
     /// [`Annotation`] and then inserting it into the list. This function will
-    /// overide any existing annotation already present.
+    /// overwrite any existing annotation already present.
     pub fn parse_insert(
         &mut self,
         annotation: impl TryInto<Annotation, Error = AnnotationError>,
@@ -198,7 +198,7 @@ impl Annotations {
         Ok(())
     }
 
-    /// Inserts a new [`Annotation`]. This function will overide any existing
+    /// Inserts a new [`Annotation`]. This function will overwrite any existing
     /// annotation already present.
     pub fn insert(&mut self, annotation: Annotation) -> &mut Self {
         self.0.insert(annotation.0);
@@ -213,7 +213,7 @@ impl Annotations {
         to self.0 {
             /// Tries to insert a new [`Annotation`]. It ensures there are no duplicate
             /// entries. Trying to insert duplicated data returns an error. If no such
-            /// check is required, use the `insert` function instead.
+            /// check is required, use [`Annotation::insert`] instead.
             pub fn try_insert(&mut self, #[newtype] annotation: Annotation) -> Result<(), AnnotationsError>;
 
             /// Extends `self` with `other`.
--- a/src/kvp/annotation/mod.rs
+++ b/src/kvp/annotation/mod.rs
@@ -44,14 +44,14 @@ pub type AnnotationError = KeyValuePairError<Infallible>;
 #[derive(Debug)]
 pub struct Annotation(KeyValuePair<AnnotationValue>);
 
-impl<T, K> TryFrom<(T, K)> for Annotation
+impl<K, V> TryFrom<(K, V)> for Annotation
 where
-    T: AsRef<str>,
     K: AsRef<str>,
+    V: AsRef<str>,
 {
     type Error = AnnotationError;
 
-    fn try_from(value: (T, K)) -> Result<Self, Self::Error> {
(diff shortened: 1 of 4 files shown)


NickLarsenNZ:
Just so I understand it correctly, does that mean instead of:

```rust
    let labels: Labels =
        BTreeMap::from([("k", "v")])
            .try_into()
            .context(BuildLabelSnafu)?;
```

I would now do:

```rust
    let labels = Labels::try_from([("k", "v")])
            .context(BuildLabelSnafu)?;
```

Techassi:
Yes! See https://github.com/stackabletech/operator-rs/blob/feat/kvps-qol-changes/src/kvp/mod.rs#L335-L343 for a small example.

src/kvp/annotation/mod.rs
```diff
@@ -45,13 +44,14 @@ pub type AnnotationError = KeyValuePairError<Infallible>;
 #[derive(Debug)]
 pub struct Annotation(KeyValuePair<AnnotationValue>);
 
-impl<T> TryFrom<(T, T)> for Annotation
+impl<T, K> TryFrom<(T, K)> for Annotation
 where
     T: AsRef<str>,
+    K: AsRef<str>,
```
NickLarsenNZ, line 50: nit: maybe we want to use `K` and `V` generics?

src/kvp/annotation/mod.rs
```diff
@@ -150,6 +150,19 @@ impl TryFrom<BTreeMap<String, String>> for Annotations {
...
+{
+    type Error = AnnotationError;
+
+    fn try_from(value: [(T, K); N]) -> Result<Self, Self::Error> {
+        let kvps = KeyValuePairs::try_from(value)?;
+        Ok(Self(kvps))
+    }
+}
+
```
NickLarsenNZ, line 165: praise: yes! this will be nice so I don't have to convert BTreeMaps into Labels. :raised_hands:

src/kvp/annotation/mod.rs
NickLarsenNZ: _same applies for labels_

NickLarsenNZ (dismissed):
Nice one! This is immediately useful (thought probably in the next round of operator maintenance)

src/kvp/annotation/mod.rs
```diff
@@ -174,17 +187,19 @@ impl Annotations {
...
+        annotation: impl TryInto<Annotation, Error = AnnotationError>,
+    ) -> Result<(), AnnotationError> {
+        self.0.insert(annotation.try_into()?.0);
+        Ok(())
     }
 
     /// Inserts a new [`Annotation`]. This function will overide any existing
-    /// annotation already present. If this behaviour is not desired, use the
-    /// `try_insert` function instead.
+    /// annotation already present.
```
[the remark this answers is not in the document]
Techassi: Yeah I never know what the "right" version is. From now on will stick with "overwrite".

src/kvp/annotation/mod.rs
Techassi: Sure, I will change that.

Techassi: This might be a little tricky, for example:

https://github.com/stackabletech/operator-rs/blob/008056d743afc0d9acdb04b721a84fe495f6b7e2/src/kvp/mod.rs#L167-L184

`V` is already used as the final value type.

src/kvp/annotation/mod.rs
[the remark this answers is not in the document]
Techassi: Changed in f856ae7

src/kvp/annotation/mod.rs
Techassi: Yeah that's a very nice improvement indeed.

src/kvp/annotation/mod.rs
NickLarsenNZ: Ah ok, personally I would do:

```rust
impl<const N: usize, T, K, V> TryFrom<[(K, V); N]> for KeyValuePairs<T>
 where 
     K: AsRef<str>, 
     V: AsRef<str>, 
     T: Value + std::default::Default, 
```

But will leave that up to you. It's not a blocker in my eyes.

Techassi: Oh that's a nice solution. I will update this.

Techassi: Changed in 0e5ca4d

NickLarsenNZ (approved):
Nicely done :rocket:

NickLarsenNZ (approved):
:+1:

PR #715, opened 2024-01-12
feat: Add TryFromIterator trait and implement it for `Labels` and `Annotations`

### LINKED ISSUE #713: Implement Labels::try_from<&BTreeMap<_,_>>
I'm trying to convert a `&BTreeMap<String,String>` into Labels.

https://github.com/stackabletech/hdfs-operator/blob/934beeaa5b5c5e85324ccf93ba4e46cef130c440/rust/operator-binary/src/pod_svc_controller.rs#L57-L67

Code with inferred types:

Showing inferred types

```
a value of type `stackable_operator::kvp::Labels` cannot be built from an iterator over elements of type `(std::string::String, std::string::String)`
the trait `FromIterator<(std::string::String, std::string::String)>` is not implemented for `stackable_operator::kvp::Labels`
the trait `FromIterator<KeyValuePair<LabelValue>>` is implemented for stackable_operator::kvp::Labels`
```

I guess we need:
- `Labels::try_from(&btree_map)` (eg: `&BTreeMap<AsRef<str>, AsRef<str>>`)
- `Iterator::collect::<Labels>()`

Techassi:
Fixes #713

It does not directly solve the issue, but `.collect()` only works on types which implement `FromIterator` which must never fail. Implementing `FromIterator for Result<Self, Error>` is also not an option, because we cannot implement a foreign trait on a foreign type. There are now two possible solutions.

#### Old

```rust
let map = BTreeMap::from([
    ("stackable.tech/managed-by", "stackablectl"),
    ("stackable.tech/vendor", "Stackable"),
]);

let labels = map
    .iter()
    .map(|p| Label::try_from(p))
    .collect::<Result<Vec<Label>, LabelError>>()
    .unwrap();

let labels = Labels::from_iter(labels);
```

#### New

```rust
let map = BTreeMap::from([
    ("stackable.tech/managed-by", "stackablectl"),
    ("stackable.tech/vendor", "Stackable"),
]);

// Works on any iterator, like Vec<(String, String)> or &[(&str, &str)]
let labels = Labels::try_from_iter(map).unwrap();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,17 @@ All notable changes to this project will be documented in this file.
 
 - Add `TryFrom<[(K, V); N]>` implementation for `Annotations` and `Labels` ([#711]).
 - Add `parse_insert` associated function for `Annotations` and `Labels` ([#711]).
+- Add generic types for `TryFrom<BTreeMap<K, V>>` impl ([#714]).
+- Add `TryFromIterator` trait, which tries to construct `Self` from an iterator. It is a falliable version of
+  `FromIterator` ([]).
+- Add `TryFromIterator` impl for `Labels` and `Annotations` ([]).
 
 ### Changed
 
 - Adjust `try_insert` for `Annotations` and `Labels` slightly ([#711]).
 
 [#711]: https://github.com/stackabletech/operator-rs/pull/711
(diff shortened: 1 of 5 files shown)


NickLarsenNZ (approved):
LGTM

PR #721, opened 2024-01-18
Add Logging::for_container helper

nightkr:
# Description

Extracted from https://github.com/stackabletech/hdfs-operator/pull/457.

```[tasklist]
# Author

```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/src/product_logging/spec.rs
+++ b/src/product_logging/spec.rs
@@ -1,7 +1,6 @@
 //! Logging structure used within Custom Resource Definitions
 
-use std::collections::BTreeMap;
-use std::fmt::Display;
+use std::{borrow::Cow, collections::BTreeMap, fmt::Display};
 
 use crate::config::{
     fragment::{self, Fragment, FromFragment},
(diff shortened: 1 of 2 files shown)


src/product_logging/spec.rs
```diff
@@ -81,6 +80,19 @@ where
     pub containers: BTreeMap<T, ContainerLogConfig>,
 }
 
+impl<T> Logging<T>
+where
+    T: Clone + Display + Ord,
+{
+    /// Get the logging configuration for `container`, falling back to the default.
+    pub fn for_container(&self, container: &T) -> Cow<ContainerLogConfig> {
```
sbernauer, line 88: WDYT of calling it like this?
```suggestion
    pub fn for_container_or_default(&self, container: &T) -> Cow<ContainerLogConfig> {
```

nightkr: I'd disagree, for the consumer there should semantically be no difference between "no configuration set" or "configuration set but empty". What this `HashMap` is trying (poorly) to emulate here is something like:

```rust
struct Containers {
  #[serde(default)]
  hdfs: ContainerLogConfig,
  #[serde(default)]
  vector: ContainerLogConfig,
}
```

sbernauer: Yeah, makes sense :+1:

sbernauer (approved):

PR #723, opened 2024-01-22
feat: Add support for Helm Values file

Techassi:
This adds support to directly work with an abstracted view of a Helm Value file by having an in-memory representation to ease the deployment using dynamic values. This is inspired by [DynamicObject](https://docs.rs/kube/latest/kube/api/struct.DynamicObject.html).

The implementation makes the assumption, that all our operators use a common set of key/values in their `values.yaml` files. Currently, we don't use Rust as a source of truth for these definitions. Instead, the values files are maintained manually. I'm experimenting with this in my private time.

Reviewers should also make suggestions, if we should add or remove certain keys.

```[tasklist]
### Tasks
- [ ] https://github.com/stackabletech/operator-rs/pull/724
- [ ] https://github.com/stackabletech/operator-rs/pull/726
- [ ] https://github.com/stackabletech/operator-rs/pull/725
- [x] Add tests for DynamicValues serialization and deserialization
```


src/helm/mod.rs
```diff
@@ -114,7 +116,11 @@ mod test {
             .labels_mut()
             .insert("stackable.tech/demo".into(), "logging".into());
 
-        let output = std::fs::read_to_string("fixtures/helm/output.yaml").unwrap();
-        assert_eq!(serde_yaml::to_string(&values).unwrap(), output);
+        let expected = std::fs::read_to_string("fixtures/helm/output.yaml").unwrap();
```
NickLarsenNZ, line 119: should the file path come in via `#[file(...)]` for `rstest`? Or is this a different usage than the ones that do?

NickLarsenNZ (approved):
LGTM, I like the test fixtures (and how they get pulled in via `rstest`.

src/helm/mod.rs
NickLarsenNZ: eg:

```rust
    #[rstest]
    fn test_serialize(
        #[files("fixtures/helm/input-required.yaml")] input: PathBuf,
        #[files("fixtures/helm/output.yaml")] output: PathBuf
    ) {
        let contents = std::fs::read_to_string(input).unwrap();
        let mut values: DynamicValues = serde_yaml::from_str(&contents).unwrap();

        values
            .labels_mut()
            .insert("stackable.tech/demo".into(), "logging".into());

        let expected = std::fs::read_to_string(output).unwrap();
...
}
```

Techassi: For now, there is only one output fixture file. I also don't know if `rstest` can currently "match" up multiple files so that the input file will be compared to the right output file.

For now, it should be fine as is. If we need more output files in the future, we can do some research on `rstest` and we might need to implement upstream changes.

Techassi: The example to mentioned will currently generate *two* test cases, one for each globbed file. In this case `test_input_required` and `test_output`. To enable the feature we want here, we definitely need upstream changes.

PR #725, opened 2024-01-23
refactor: Change Deref target of KeyPrefix and KeyName to str

Techassi:
Came up in https://github.com/stackabletech/hdfs-operator/pull/454.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- Change Deref target of `KeyPrefix` and `KeyName` to `str` ([#725]).
+
+[#725]: https://github.com/stackabletech/operator-rs/pull/725
+
 ## [0.62.0] - 2024-01-19
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-- Change Deref target of `KeyPrefix` and `KeyName` to `str` ([#725]).
+- Change Deref target of `KeyPrefix` and `KeyName` from `String` to `str` ([#725]).
 
 [#725]: https://github.com/stackabletech/operator-rs/pull/725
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- Change Deref target of `KeyPrefix` and `KeyName` to `str` ([#725]).
```
sbernauer, line 9:
```suggestion
- Change Deref target of `KeyPrefix` and `KeyName` from `String` to `str` ([#725]).
```

NickLarsenNZ (dismissed):
LGTM - only thing that is arguably missing would be a test to show how it gets used (eg: https://github.com/stackabletech/hdfs-operator/pull/454#discussion_r1462961182)

Techassi:
Sure, I can add a few tests which should cover the use-case you mentioned in `hdfs-operator`.

NickLarsenNZ (approved):

sbernauer (dismissed):

src/kvp/key.rs
```diff
@@ -399,4 +402,30 @@ mod test {
...
+    #[rstest]
+    #[case("app.kubernetes.io/name", true)]
+    #[case("name", false)]
+    fn key_prefix_deref(#[case] key: &str, #[case] expected: bool) {
+        let label = Label::try_from((key, "zookeeper")).unwrap();
+
+        let is_valid = label
+            .key()
+            .prefix()
+            .is_some_and(|prefix| *prefix == "app.kubernetes.io");
```
NickLarsenNZ, line 415: Shouldn't this be:

```rust
.prefix()
.as_deref() == "app.kubernetes.io";
```

As Nat was suggesting in the coomment linked in the PR review?

Or maybe I'm misunderstanding what she meant.

NickLarsenNZ (approved):
LGTM

src/kvp/key.rs
NickLarsenNZ: Yeah, so the `as_deref()` comparison to a `&str` doesn't work (as per your linked comment), but I think this is still a net-positive improvement.

Techassi: Mark as resolved for now, tracking done in https://github.com/stackabletech/operator-rs/issues/727

PR #726, opened 2024-01-24
refactor: Add context to key/value pair error variants

Techassi:
Came up in https://github.com/stackabletech/hdfs-operator/pull/454

This PR adds more context to the key/value pair error variants which should make it easier for the user to identify the erroneous pair.

### DIFF
--- a/src/kvp/mod.rs
+++ b/src/kvp/mod.rs
@@ -30,12 +30,12 @@ where
 {
     /// Indicates that the key failed to parse. See [`KeyError`] for more
     /// information about the error causes.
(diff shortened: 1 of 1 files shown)


NickLarsenNZ (approved):
LGTM

NickLarsenNZ:
@Techassi just a note, the CHANGELOG wasn't updated, but maybe this change was small enough to not need one?

Techassi:
Yeah. I feel like this was small enough to not have an entry. This change also isn't developer-facing and only affects the error display. So it should be fine as is...

PR #736, opened 2024-02-13
feat: Add TLS certificate util crate

Techassi:
Follow-up of #730, discussed in https://github.com/stackabletech/documentation/pull/522, tracked by https://github.com/stackabletech/issues/issues/361

This PR adds a crate which handles TLS certificate generation, reading and writing.


Techassi:
The basic work is mostly done, but a few pieces are still missing:

- Add a manager, which contains one or more CAs, which will be rotated automatically once they expire.
- Add dynamic TLS cert resolver in the webhook server

Techassi:
>     * Add a manager, which contains one or more CAs, which will be rotated automatically once they expire.
> 
>     * Add dynamic TLS cert resolver in the webhook server

I decided to move this into a separate PR as this PR is already big enough. Both features can be implemented on top of the changes introduced in this PR.

stackable-certs/src/keys/ecdsa.rs
NickLarsenNZ, line 1: **question**: Do we only support ECDSA p256? Does p384 require an extra import (just judging by the `p256` crate name)?

stackable-certs/src/ca/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+        // There are multiple default extensions included in the profile. For
+        // the root profile, these are:
+        //
+        // - BasicConstraints marked as critical and CA = true
+        // - SubjectKeyIdentifier with the 160-bit SHA-1 hash of the subject
+        //   public key.
+        // - KeyUsage with KeyCertSign and CRLSign bits set. Ideally we also
+        //   want to include the DigitalSignature bit, which for example is
+        //   required for CA certs which want to sign an OCSP response.
+        //   Currently, the root profile doesn't include that bit.
```
NickLarsenNZ, line 184: TIL :smile:

stackable-certs/src/ca/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+            signer,
+        )
+        .context(CreateCertificateBuilderSnafu)?;
+
+        // Add extension constructed above
+        builder
+            .add_extension(&aki)
+            .context(AddCertificateExtensionSnafu)?;
+
+        info!("create and sign CA certificate");
```
NickLarsenNZ, line 211: Being a _lib_ crate, I think maybe this should be...

```suggestion
        debug!("create and sign CA certificate");
```

stackable-certs/src/ca/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+            serial_numbers: HashSet::new(),
+            certificate_pair: CertificatePair {
+                key_pair: signing_key_pair,
+                certificate,
+            },
+        })
+    }
+
+    #[instrument(skip(key_pair))]
+    pub fn generate_leaf_certificate<T>(
```
NickLarsenNZ, line 224: **suggestion (non-blocker)**: `pub fn` should probably have doc-comments.

NickLarsenNZ (dismissed):
That's a lot of work. Looking really good.

I left a bunch of comments, sorry if it's too many (though some being disregarded will lead to many others being able to be marked as resolved).

stackable-certs/src/ca/k8s.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl<S> K8sCertificatePair for CertificateAuthority<S>
+where
+    S: KeypairExt,
+    <S::SigningKey as signature::Keypair>::VerifyingKey: EncodePublicKey,
+{
+    type Error = SecretError<S::Error>;
+
+    fn from_secret(
+        secret: Secret,
```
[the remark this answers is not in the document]
Techassi: Sure, we can have a guard clause in this function.

stackable-certs/src/ca/mod.rs
Techassi: We should maybe add this to the code style guide.

stackable-certs/src/ca/mod.rs
Techassi: Yes, I will add doc comments for it.

stackable-certs/src/keys/rsa.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use rand_core::{CryptoRngCore, OsRng};
+use rsa::{pkcs8::DecodePrivateKey, RsaPrivateKey};
+use signature::Keypair;
+use snafu::{ResultExt, Snafu};
+use tracing::instrument;
+
+use crate::keys::KeypairExt;
+
+pub const DEFAULT_RSA_BIT_SIZE: usize = 2048;
```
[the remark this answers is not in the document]
Techassi: Sure, we can do that!

stackable-certs/src/keys/rsa.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    #[snafu(display("failed to deserialize the signing (private) key from PEM-encoded PKCS8"))]
+    DeserializeSigningKey { source: rsa::pkcs8::Error },
+}
+
+#[derive(Debug)]
+pub struct SigningKey(rsa::pkcs1v15::SigningKey<sha2::Sha256>);
+
+impl SigningKey {
+    // NOTE (@Techassi): Should we maybe enfore bit sizes >= 2048?
```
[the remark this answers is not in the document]
Techassi: So yeah let's add a guard for it.

stackable-certs/src/lib.rs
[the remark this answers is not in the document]
Techassi: We spoke about it and are happy to leave it as is.

A certificate basically binds the public key half of a key pair to some X509 subject (with additional metadata).

stackable-webhook/src/options.rs
```diff
@@ -120,14 +123,14 @@ impl OptionsBuilder {
     /// [`OptionsBuilder::tls_autogenerate()`] function.
     pub fn tls_mount(
         mut self,
-        public_key_path: impl Into<PathBuf>,
+        certificate_path: impl Into<PathBuf>,
```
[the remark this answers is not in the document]
Techassi: https://github.com/stackabletech/operator-rs/pull/736/files#r1512482718

stackable-webhook/src/options.rs
```diff
@@ -120,14 +123,14 @@ impl OptionsBuilder {
...
         private_key_path: impl Into<PathBuf>,
-        private_key_encoding: PrivateKeyEncoding,
+        private_key_type: PrivateKeyType,
     ) -> Self {
         self.tls = Some(TlsOption::Mount {
-            public_key_path: public_key_path.into(),
             private_key_path: private_key_path.into(),
-            private_key_encoding,
+            certificate_path: certificate_path.into(),
```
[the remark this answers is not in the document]
Techassi: https://github.com/stackabletech/operator-rs/pull/736/files#r1512482718

stackable-webhook/src/options.rs
```diff
@@ -146,9 +149,9 @@ impl OptionsBuilder {
 pub enum TlsOption {
     AutoGenerate,
     Mount {
-        private_key_encoding: PrivateKeyEncoding,
-        public_key_path: PathBuf,
+        private_key_type: PrivateKeyType,
         private_key_path: PathBuf,
+        certificate_path: PathBuf,
```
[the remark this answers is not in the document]
Techassi: https://github.com/stackabletech/operator-rs/pull/736/files#r1512482718

stackable-webhook/src/tls.rs
```diff
@@ -57,27 +55,48 @@ impl TlsServer {
                 todo!()
             }
             TlsOption::Mount {
-                public_key_path,
+                certificate_path,
```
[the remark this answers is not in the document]
Techassi: https://github.com/stackabletech/operator-rs/pull/736/files#r1512482718

stackable-webhook/src/tls.rs
```diff
@@ -57,27 +55,48 @@ impl TlsServer {
...
-                    private_key_path,
-                    private_key_encoding,
-                )
-                .context(TlsCertificateChainSnafu)?
-                .into_parts();
+                // TODO (@Techassi): Remove unwraps
+                let (cert, pk) = match private_key_type {
+                    PrivateKeyType::Ecdsa => {
+                        let pair = CertificatePair::<ecdsa::SigningKey>::from_files(
+                            certificate_path,
```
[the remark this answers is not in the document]
Techassi: https://github.com/stackabletech/operator-rs/pull/736/files#r1512482718

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+    DeserializePrivateKey {
+        source: E,
+        key_encoding: KeyEncoding,
+    },
+
+    #[snafu(display("failed to write file"))]
+    WriteFile { source: std::io::Error },
+
+    #[snafu(display("failed to read file"))]
+    ReadFile { source: std::io::Error },
```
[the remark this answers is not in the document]
Techassi: Yeah it should be there.

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+        let doc = self
+            .key_pair
+            .signing_key()
+            .to_pkcs8_der()
+            .context(SerializePrivateKeySnafu {
+                key_encoding: KeyEncoding::Der,
+            })?;
+
+        let bytes = doc.to_bytes().deref().to_owned();
+        let der = PrivateKeyDer::from(PrivatePkcs8KeyDer::from(bytes));
```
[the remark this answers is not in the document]
Techassi: Basically the whole thing is damned.

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// Reads in a PEM-encoded certificate from `certificate_path` and private
+    /// key file from `private_key_path` and finally constructs a CA from the
+    /// contents.
+    fn from_files(
+        certificate_path: impl AsRef<Path>,
+        private_key_path: impl AsRef<Path>,
+    ) -> Result<Self, Self::Error>;
+
+    fn to_certificate_file(
```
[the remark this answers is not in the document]
Techassi: I will add a doc comment for it, which will basically say it's always Base64 PEM.

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+/// let private_key_path = PathBuf::private_key_path("path/to/my/tls-pk");
+/// assert_eq!(private_key_path.display().to_string(), "path/to/my/tls-pk.key");
+/// ```
+pub trait PathBufExt {
+    /// Returns a path to `<path>.crt`.
+    ///
+    /// The default extension is defined by [`CERTIFICATE_FILE_EXT`].
+    fn certificate_path(path: impl AsRef<Path>) -> PathBuf {
+        PathBuf::from(path.as_ref()).with_extension(CERTIFICATE_FILE_EXT)
+    }
```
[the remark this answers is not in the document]
Techassi: Regarding the first question: `with_extension` replaces the file extension as a whole. So there will be no chaining of multiple extensions. See https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.with_extension

This is opt-in and people can use different extensions if they want. The default we provide is based on the Kubernetes file extensions.

stackable-certs/src/keys/ecdsa.rs
Techassi: We would need a separate crate for that. It would also involve adding a parameter for type of ECDSA key we want to generate.

stackable-certs/src/keys/rsa.rs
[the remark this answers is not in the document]
Techassi: Changed in 9c21a1c and 0d5c001

stackable-certs/src/keys/rsa.rs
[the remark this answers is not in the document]
Techassi: Added in 0d5c001

stackable-certs/src/lib.rs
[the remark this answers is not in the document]
Techassi: Added in 2b80739

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+    S: KeypairExt,
+    <S::SigningKey as signature::Keypair>::VerifyingKey: EncodePublicKey,
+{
+    type Error = CertificatePairError<S::Error>;
+
+    fn from_files(
+        certificate_path: impl AsRef<Path>,
+        private_key_path: impl AsRef<Path>,
+    ) -> Result<Self, Self::Error> {
+        let certificate_pem = std::fs::read(certificate_path).context(ReadFileSnafu)?;
```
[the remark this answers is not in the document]
Techassi: See https://github.com/stackabletech/operator-rs/pull/743

stackable-certs/src/ca/k8s.rs
```diff
@@ -17,6 +17,9 @@ where
     #[snafu(display("failed to retrieve secret {secret:?}"))]
     GetSecret { source: kube::Error, secret: String },
 
+    #[snafu(display("invalid secret type, expected kubernetes.io/tls"))]
+    InvalidSecretType,
```
NickLarsenNZ, line 21: nit (non-blocker): I think the Secret _type_ string shouldn't be written in two places. Either it should be a `const` or passed into the Snafu.

stackable-certs/src/keys/rsa.rs
```diff
@@ -10,7 +10,8 @@ use tracing::instrument;
 
 use crate::keys::KeypairExt;
 
-pub const DEFAULT_RSA_BIT_SIZE: usize = 2048;
+pub const DEFAULT_BIT_SIZE: usize = 4096;
+pub const MINIMUM_BIT_SIZE: usize = 2048;
```
NickLarsenNZ, line 14: **thought (non-blocker)**: Almost looks like an enum would suffice for the valid key lengths (if we are using the standard base2.

But that would add a bit more code with having to impl TryFrom/TryInto for usize.

stackable-certs/src/lib.rs
```diff
@@ -11,7 +11,7 @@
 //!
 //! - `k8s`: This enables various traits and functions to work with
 //!   certificates and Kubernetes secrets.
-//! - `rustls`: This enables interoperability between this crates types
+//! - `webhook`: This enables interoperability between this crates types
```
NickLarsenNZ, line 14: **praise**: This makes more sense now :+1:

NickLarsenNZ (approved):
LGTM

stackable-certs/src/ca/k8s.rs
Techassi: Good catch! I will add a `const` for that.

stackable-certs/src/lib.rs
Techassi: Yeah it definitely does.

The old name kinda made sense (because it enabled a lot of rustls related dependencies), but the new one actually describes what it is achieving.

stackable-certs/src/keys/rsa.rs
Techassi: Sure, we can do that.

I don't know if we need a `TryFrom` impl, but we can definitely provide a `From<Enum> for usize` impl.

stackable-certs/src/ca/k8s.rs
Techassi: Fixed in 7a6fc12

stackable-certs/src/keys/rsa.rs
NickLarsenNZ: I like that!

NickLarsenNZ (approved):
LGTM

NickLarsenNZ (dismissed):
Nice one

stackable-certs/src/ca/consts.rs
```diff
@@ -1,23 +1,5 @@
...
-/// country.
-pub const ROOT_CA_SUBJECT: &str = concatcp!(
-    ROOT_CA_COMMON_NAME_DN,
-    ",",
-    ORGANIZATION_DN,
-    ",",
-    COUNTRY_DN
-);
+/// The root CA subject name containing only the common name.
+pub const ROOT_CA_SUBJECT: &str = "CN=Stackable Data Platform Internal CA";
```
NickLarsenNZ, line 5: I like this more :smile:

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+/// Provides functions to work with CAs stored in Kubernetes secrets.
+///
+/// Namely these function enable:
+///
+/// - decoding a certificate from a Kubernetes secret
+/// - encoding a certificate as a Kubernetes secret
+#[cfg(feature = "k8s")]
+pub trait CertificatePairExt: Sized {
```
nightkr, line 165: This still isn't an extension trait in the traditional sense. I'm not sure it should be a trait at all, but if it is then it should have a better name.

stackable-certs/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    type Error: std::error::Error;
+
+    fn from_secret(
+        secret: Secret,
+        key_certificate: &str,
+        key_private_key: &str,
+    ) -> Result<Self, Self::Error>;
+
+    #[allow(async_fn_in_trait)]
```
nightkr, line 174: This warning exists for a reason, we need the future to be `Send`. Async trait _definitions_ should be `fn foo(...) -> impl Future<Output = ...> + Send + '_`, not `async fn foo(...) -> ...` (if they're not using `#[async_trait]`).

stackable-certs/src/keys/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+use p256::pkcs8::EncodePrivateKey;
+use signature::{Keypair, Signer};
+use x509_cert::spki::{EncodePublicKey, SignatureAlgorithmIdentifier, SignatureBitStringEncoding};
+
+pub mod ecdsa;
+pub mod rsa;
+
+// NOTE (@Techassi): This can _maybe_ be slightly simplified by adjusting the
+// trait and using a blanket impl on types which implement Deref<Target = _>.
+pub trait KeypairExt
```
nightkr, line 44:
```suggestion
pub trait Keypair
```

This isn't an extension trait, it's semantically meaningful.

src/commons/secret.rs
```diff
@@ -0,0 +1,36 @@
+use std::fmt::Display;
+
+use k8s_openapi::api::core::v1::Secret;
+use kube::runtime::reflector::ObjectRef;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+// Redefine SecretReference instead of reusing k8s-openapi's, in order to make name/namespace mandatory.
```
nightkr, line 8: Should be a doccomment. We should also clarify that this is _must_ only be used in cluster-scoped objects. Namespaced objects _must not_ be able to define cross-namespace secret references.

stackable-certs/src/lib.rs
Techassi: I could instead implement the function directly on the struct and use the feature gates on the functions directly.

stackable-certs/src/lib.rs
Techassi: Yeah... True. I will change it.

src/commons/secret.rs
Techassi: I basically copied this straight from https://github.com/stackabletech/secret-operator/blob/main/rust/crd-utils/src/lib.rs#L12-L37 but I can adjust the comment as you requested.

stackable-certs/src/keys/mod.rs
Techassi: I actually need to revert this change, because it clashes with `signature::Keypair` when both traits are required to be in scope.

stackable-certs/src/lib.rs
Techassi: Changed in a372b7a

stackable-certs/src/lib.rs
Techassi: Removed in a372b7a

src/commons/secret.rs
Techassi: Added in 17a6aef

stackable-certs/src/keys/mod.rs
Techassi: Do you have an idea for a better name, which doesn't clash with the external trait?

stackable-certs/src/lib.rs
nightkr: They can also be freestanding functions, this isn't Java. :P

stackable-certs/src/keys/mod.rs
nightkr: `CertificateKeypair`.. maybe? But it also feels kind of redundant to end up needing both traits that do more or less the same thing.

src/commons/secret.rs
nightkr: This isn't about them being mandatory, an optional `namespace` field would still be problematic in a namespaced object.

stackable-certs/src/keys/rsa.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// The bit size of an RSA key pair. To retrieve the bit size as an integer,
+/// use numeric casting via the `as` keyword.
+///
+/// This can either be:
+///
+/// - [`BitSize::Default`], with a value of `4096`
+/// - [`BitSize::Minimum`], with a value of `2048`
+#[derive(Debug, Default)]
+pub enum BitSize {
```
nightkr, line 85:
```suggestion
pub enum KeySize {
```

stackable-certs/src/keys/rsa.rs
```diff
@@ -0,0 +1,89 @@
...
+    #[default]
+    Default = 4096,
+    Minimum = 2048,
```
nightkr, line 88: Why do we even support 2048?

nightkr, line 88: I'd prefer to have a `bits(self)` method, rather than assume that the discriminant is the key size.

stackable-certs/src/keys/rsa.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/// The bit size of an RSA key pair. To retrieve the bit size as an integer,
+/// use numeric casting via the `as` keyword.
+///
+/// This can either be:
+///
+/// - [`BitSize::Default`], with a value of `4096`
+/// - [`BitSize::Minimum`], with a value of `2048`
+#[derive(Debug, Default)]
```
nightkr, line 84:
```suggestion
#[derive(Debug, Default, Clone, Copy)]
```

stackable-certs/src/keys/rsa.rs
Techassi: If we don't support 2048, we can completely drop the enum and hard-code 4096.

Techassi: Added in 7a96031

src/commons/secret.rs
Techassi: Adjusted in 694e337. Is this better now?

stackable-certs/src/lib.rs
Techassi: I know haha :) But I like the style `CertificateAuthority::from_secret()`

stackable-certs/src/keys/mod.rs
Techassi: Renamed in c96fa62

There are a few ways to simplify the trait bounds. If we want to do that, I would do that in a separate PR.

Or we could get rid of it entirely, if we only support one type of key, either RSA or ECDSA.

stackable-certs/src/keys/rsa.rs
Techassi: Removed in 862d2bb

stackable-certs/src/ca/mod.rs
```diff
@@ -294,13 +331,84 @@ where
...
+
+    /// Create a [`CertificateAuthority`] from a Kubernetes [`SecretReference`].
+    #[instrument(
+        name = "create_certificate_authority_from_k8s_secret_ref",
+        skip(secret_ref, client)
+    )]
+    pub async fn from_secret_ref(
+        secret_ref: &SecretReference,
+        key_certificate: &str,
+        key_private_key: &str,
```
NickLarsenNZ, line 392: **nit (non-blocker)**: the duplicate `key` irks me a little.

```suggestion
        certificate: &str,
        private_key: &str,
```

NickLarsenNZ: There is another instance further up in this file.

stackable-certs/src/ca/mod.rs
```diff
@@ -294,13 +331,84 @@ where
...
+        client: &Client,
+    ) -> Result<Self, SecretError<S::Error>> {
+        let secret_api = client.get_api::<Secret>(&secret_ref.namespace);
+        let secret = secret_api
+            .get(&secret_ref.name)
+            .await
+            .context(GetSecretSnafu {
+                secret_ref: secret_ref.to_owned(),
+            })?;
```
NickLarsenNZ, line 401: **suggestion (non-blocker)**: As discussed, this could/will allocate, so we _should_ use `with_context`.

If it is an `Option` (`snafu::OptionExt` is already imported):

```suggestion
            .with_context(|| GetSecretSnafu {
                secret_ref: secret_ref.to_owned(),
            })?;
```

or if it was a `Result`:

```suggestion
            .with_context(|_| GetSecretSnafu {
                secret_ref: secret_ref.to_owned(),
            })?;
```

NickLarsenNZ (approved):
Left some comments. None of them should be blockers. Will let you decide whether to include them.

stackable-certs/src/ca/mod.rs
```diff
@@ -68,20 +68,58 @@ pub enum Error {
     ParseAuthorityKeyIdentifier { source: x509_cert::der::Error },
 }
 
+/// Defines all error variants which can occur when loading a CA from a
+/// Kubernetes [`Secret`].
+#[derive(Debug, Snafu)]
+pub enum SecretError<E>
```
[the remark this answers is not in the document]
Techassi: I would argue that this is implied by context (for now). Currently, this code is only used for Kubernetes specific tasks.

If this will ever change, we can rename it in the future.

stackable-certs/src/ca/mod.rs
```diff
@@ -68,20 +68,58 @@ pub enum Error {
...
+    #[snafu(display("the secret {secret:?} does not contain TLS private key data"))]
+    NoPrivateKeyData { secret: ObjectRef<Secret> },
+
+    #[snafu(display("failed to read PEM-encoded certificate chain from secret {secret:?}"))]
+    ReadChain {
+        source: x509_cert::der::Error,
+        secret: ObjectRef<Secret>,
+    },
+
+    #[snafu(display("failed to parse Base64 encoded byte string"))]
```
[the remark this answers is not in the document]
Techassi: Good catch. I will adjust the message.

stackable-certs/src/ca/mod.rs
Techassi: Yeah this is a little weird, but we are talking about the _key_ of the key/value pair in the secret.

I don't know if there is any better way of naming these parameters.

stackable-certs/src/ca/mod.rs
Techassi: Fixed in dd51591

stackable-certs/src/ca/mod.rs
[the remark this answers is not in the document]
Techassi: Adjusted in b90dc1f

stackable-certs/src/ca/mod.rs
NickLarsenNZ: Oh, I didn't even consider it meant that (makes sense in the context), though I would maybe expect `key` to then be the suffix for both.

Eg:

```suggestion
        certificate_key: &str,
        private_key_key: &str,
```

or for more clarity, to [avoid ambiguity](https://docs.stackable.tech/home/nightly/contributor/code-style-guide#_long_versus_abbreviated)

```suggestion
        certificate_secret_key: &str,
        private_key_secret_key: &str,
```

Techassi: Added a doc comment in 4eb2743

NickLarsenNZ (approved):
LGTM :rocket:

PR #739, opened 2024-02-14
chore: Adjust changelog to include changes from #730

Techassi:
This is a follow-up PR to add an entry to the changelog for changes introduced in #730. The previous PR accidentally didn't include this entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## Added
+
+- Add `stackable_webhook` crate which provides utilities to create webhooks with TLS termination ([#730]).
+
+[#730]: https://github.com/stackabletech/operator-rs/pull/730
+
 ## Changed
 
 - Remove `resources` key from `DynamicValues` struct ([#734]).


sbernauer (approved):

NickLarsenNZ (approved):
Easily missed.

Is there a way in Github Actions to warn but not block?
Would be nice to have a check for changelog edits in PRs but not enforce it.

Techassi:
> Easily missed.
> 
> Is there a way in Github Actions to warn but not block?
> Would be nice to have a check for changelog edits in PRs but not enforce it.

We could use conventional commits for auto-generation of the change log. But that would require some more tooling and research. Happy to explore this further, as I already suggested it a few months back.

NickLarsenNZ:
> We could use conventional commits for auto-generation of the change log. But that would require some more tooling and research. Happy to explore this further, as I already suggested it a few months back.

I'm keen to automate this, but wonder how much friction is could generate? I'm a fan of conventional commits (for this reason, but also search-ability) but others might find it a bit of a showstopper to their flow. Especially due to the whole squash thing where commit messages are quite free-style.

I know you said you'd explore it further, but my question is: would it be a pre-commit/pre-push hook run locally? Or something that runs in Github Actions which then pushes a commit?

Either way, I'm pro-exploration.

PR #749, opened 2024-03-11
feat: Add re-export for ConversionReview type in `stackable-webhook`

Techassi:
This adds a little more convenience when using `ConversionWebhookServer`. Users now don't need an additional explicit dependency on `kube` to import the required type.

Came up in https://github.com/stackabletech/documentation/pull/561

### DIFF
--- a/stackable-webhook/src/servers/conversion.rs
+++ b/stackable-webhook/src/servers/conversion.rs
@@ -3,7 +3,10 @@ use std::fmt::Debug;
 use axum::{extract::State, routing::post, Json, Router};
 use tracing::{debug, instrument};
 
-// Selected re-exports
+// Re-export this type because users of the conversion webhook server require
+// this type to write the handler function. Instead of importing this type from
+// kube directly, consumers can use this type instead. This also eliminates
+// keeping the kube dependency version in sync between here and the operator.
 pub use kube::core::conversion::ConversionReview;
 
 use crate::{options::Options, StatefulWebhookHandler, WebhookHandler, WebhookServer};
--- a/stackable-webhook/src/servers/conversion.rs
+++ b/stackable-webhook/src/servers/conversion.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 
 use axum::{extract::State, routing::post, Json, Router};
-use kube::core::conversion::ConversionReview;
+pub use kube::core::conversion::ConversionReview;
 use tracing::{debug, instrument};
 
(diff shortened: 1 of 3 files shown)


stackable-webhook/src/servers/conversion.rs
```diff
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 
 use axum::{extract::State, routing::post, Json, Router};
-use kube::core::conversion::ConversionReview;
+pub use kube::core::conversion::ConversionReview;
```
sbernauer, line 4: I'm honestly not sure, but I **think** in the past we tried to get rid of re-exports in operator-rs (something something about picking versions and stuff), but @nightkr knows best.

Techassi: I would be happy to leave it as is, but we then require an explicit dependency. This is mostly relevant for the guide in https://github.com/stackabletech/documentation/pull/561.

The current example needs adjustment and we should also add a hint for the developer to keep the dependency version in sync between `stackable-webhook` and the operator.

Techassi: One small follow-up: I would keep the change in this line https://github.com/stackabletech/operator-rs/blob/d34e77bafbea6d2565235448910339678b83d0d7/stackable-webhook/src/lib.rs#L38

It makes a little more sense to import `ConversionWebhookServer` (and other servers in the future) from the `servers` module instead of the root level.

nightkr: > I'm honestly not sure, but I think in the past we tried to get rid of re-exports in operator-rs (something something about picking versions and stuff), but @nightkr knows best.

Largely, the other way around. The operators _need_ to use the same version of kube-rs as op-rs, so it's preferable if we just have to decide the version in op-rs and then re-export from there.

Techassi:
Thanks for the approval, I will add the above mentioned changes https://github.com/stackabletech/operator-rs/pull/749#discussion_r1521059405, which requires another approval when done.

stackable-webhook/src/servers/conversion.rs
Techassi: Adjusted in 5314e8e

nightkr: I don't think that comment adds any value IMO, it's _already_ selected by virtue of having the `pub`. If you do think it warrants a comment, mention ___WHY___ it warrants a re-export.

Techassi: Clarified in 46dfae2

Issue #737, opened 2024-02-13
Use workspace dependencies


Techassi:
Blocked by #730 and #736.

We want to use a single place to specify all our dependencies across all workspace members. Currently, all workspace members still manage their dependencies in their own `Cargo.toml` file.

This refactor should bring this repo more in line with other repositories, like https://github.com/stackabletech/stackable-cockpit

PR #751, opened 2024-03-21
refactor: Use workspace dependencies, move crates into common folder

Techassi:
Part of https://github.com/stackabletech/issues/issues/531, fixes https://github.com/stackabletech/operator-rs/issues/737


crates/stackable-operator-derive/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "stackable-operator-derive"
 description = "Derive macros for the Stackable Operator Framework"
-version.workspace = true
+version = "0.1.0"
```
NickLarsenNZ, line 4: Why do we hard code the in each crate?
Do we want to version everything independently (wondering what the burden is)?

IMO, we should re-export the crates and use feature flags so we can simply add one dependency to the operators and toggle the features that way. Not a strong opinion though.

NickLarsenNZ (approved):
LGTM, left a comment (non-blocker)

Cargo.toml
```diff
@@ -1,29 +1,21 @@
+[workspace]
+members = ["crates/*"]
```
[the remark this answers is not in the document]
Techassi: It is very nice indeed :)

crates/stackable-operator-derive/Cargo.toml
Techassi: > Do we want to version everything independently (wondering what the burden is)?

In my opinion, we should do that. Always greater flexibility. We might need some tooling in the future.

> IMO, we should re-export the crates and use feature flags so we can simply add one dependency to the operators and toggle the features that way. Not a strong opinion though.

That's part of my long-term vision. I'm working towards that goal step-by-step and we eventually will use selected re-exports for various stuff.

PR #752, opened 2024-03-22
fix: Implement custom schema for `config.affinity.nodeSelector`

sbernauer:
# Description

Fixes https://github.com/stackabletech/issues/issues/554.

IMHO it's not worth an **immediate** platform bugfix release, as there is a workaround available.

The effect can be seen e.g. in Trino:
```
diff --git a/deploy/helm/trino-operator/crds/crds.yaml b/deploy/helm/trino-operator/crds/crds.yaml
index e41fdfb..5de8694 100644
--- a/deploy/helm/trino-operator/crds/crds.yaml
+++ b/deploy/helm/trino-operator/crds/crds.yaml
@@ -305,6 +305,8 @@ spec:
                                   type: object
                               type: object
                             nodeSelector:
+                              additionalProperties:
+                                type: string
                               nullable: true
                               type: object
                             podAffinity:
@@ -647,6 +649,8 @@ spec:
                                     type: object
                                   type: array
                               type: object
+                          required:
+                            - nodeSelector
                           type: object
```

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


Techassi:
> IMHO it's not worth an immediate bugfix release, as there is a workaround available.

I would be happy to release a new version, because there are now multiple things which are unreleased.

Techassi:
I would however do one additional PR regarding the changelog before bumping the version.

CHANGELOG.md
```diff
@@ -9,8 +9,13 @@ All notable changes to this project will be documented in this file.
 - Add `stackable_webhook` crate which provides utilities to create webhooks with TLS termination ([#730]).
 - Add `ConversionReview` re-export in `stackable_webhook` crate ([#749]).
 
+## Fixed
+
+- Fixed wrong schema (and thus CRD) for `config.affinity.nodeSelector`, making it unusable ([#752]).
```
Techassi, line 14: suggestion: The fix makes it _usable_ again. So I would suggest to just remove that note.

```suggestion
- Fix wrong schema (and thus CRD) for `config.affinity.nodeSelector` ([#752]).
```

crates/stackable-operator/src/commons/affinity.rs
```diff
@@ -51,6 +60,10 @@ pub struct StackableNodeSelector {
 
 impl Atomic for StackableNodeSelector {}
 
+pub fn stackable_node_selector_schema(gen: &mut schemars::gen::SchemaGenerator) -> Schema {
```
Techassi, line 63: note: Can we maybe add a small comment to this function to better understand where and how this is used?

Techassi (changes requested):

crates/stackable-operator/src/commons/affinity.rs
sbernauer: I tried containing all the docs in a single place on `StackableNodeSelector`.
WDYT of

We need a custom JsonSchema for StackableNodeSelector, please have a look at the documentation there

Techassi: Yeah something along those lines is fine!

sbernauer: Added, please have a look

Techassi (approved):

nightkr:
I tried diffing the output before/after this patch: https://gist.github.com/nightkr/9446c9e3747aa6f58ce616cd86c1f13e

As far as I can tell, the actual issue here is that schemars doesn't propagate `additionalProperties` when used with `#[serde(flatten)]`? We should probably file an upstream ticket for that and link to it in the code.

crates/stackable-operator/src/commons/affinity.rs
```diff
@@ -51,6 +60,11 @@ pub struct StackableNodeSelector {
 
 impl Atomic for StackableNodeSelector {}
 
+/// We need a custom JsonSchema for [`StackableNodeSelector`], please have a look at the documentation there.
+pub fn optional_stackable_node_selector_schema(gen: &mut schemars::gen::SchemaGenerator) -> Schema {
+    Option::<BTreeMap<String, String>>::json_schema(gen)
+}
```
nightkr, line 66: This could be an explicit `impl JsonSchema for StackableNodeSelector`, which should make this non-breaking

nightkr:
This looks like https://github.com/GREsau/schemars/issues/259.

nightkr:
Another workaround is to add `#[schemars(deny_unknown_fields)]`  to `StackableNodeSelector`: https://gist.github.com/nightkr/af2236b2d3cf77eb71242124d588a638

sbernauer:
Many thanks for tracking this down @nightkr! What would you suggest as the next step?
1. Merge as is
2. Use `#[schemars(deny_unknown_fields)]`
3. Wait for a few days to see if https://github.com/GREsau/schemars/pull/275 get's merged
4. Something else

nightkr:
@sbernauer I'd say 2 and add a tracking issue for 3. schemars doesn't look _too_ lively these days so we shouldn't block on it.

sbernauer:
Makes sense to me! Adopted Option 2 and created https://github.com/stackabletech/operator-rs/issues/756.
A quick review would be great, so that @Techassi can release :)

crates/stackable-operator/src/commons/affinity.rs
sbernauer: Good to know!

crates/stackable-operator/src/commons/affinity.rs
```diff
@@ -42,8 +42,18 @@ pub struct StackableAffinity {
     pub node_selector: Option<StackableNodeSelector>,
 }
 
+/// We can not simply use [`BTreeMap<String, String>`] in [`StackableAffinity`], as the fields need to be [`Atomic`].
+/// We can not mark it as [`Atomic`], as [`crate::config::fragment::FromFragment`] is already implemented for
+/// [`BTreeMap<String, String>`].
+///
+/// We `#[serde(flatten)]` the contained [`BTreeMap<String, String>`], so `serde_yaml` can deserialize everything as
+/// expected.
```
nightkr, line 50: This probably shouldn't be a doccomment, since we don't want it in the CRD docs.

NickLarsenNZ: Marking as resolved

NickLarsenNZ (approved):
I trust that this works, and as I understand it we simple remove the `[schemars(deny_unknown_fields)]` as soon as that schemars issue is resolved.

PR #753, opened 2024-03-22
chore(deps): Bump Rust dependencies

Techassi:
This PR bumps various Rust dependencies. The changes pass the tests and `cargo check`.

The `opentelemetry-jaeger` crate was recently deprecated for `opentelemetry-otlp`, because Jaeger now natively supports the OTEL protocol. This requires a follow-up PR implementing the required changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,14 @@ All notable changes to this project will be documented in this file.
 ## Changed
 
 - Remove `resources` key from `DynamicValues` struct ([#734]).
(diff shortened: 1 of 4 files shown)


sbernauer (approved):
LGTM. I let you decide if you want to hear from Lars in terms of the license (but I think there is no real alternative)

Techassi:
Spoke with @lfrancke and we decided to go forward for now. If we ever need to move away from that license, we can do so in the future.

PR #757, opened 2024-03-25
chore: implement PartialEq for most Snafu Error enums

NickLarsenNZ:
# Description

This change is to encourage testing error variants by type rather than text contents.

- Bump strum to 0.26.2
- Derive `PartialEq` on most _Snafu_ Error enums.
- Add custom implementations of `PartialEq` for _Snafu_ error enums that use types that are not directly comparable.

>[!IMPORTANT] 
> Some variants are impossible to compare, so these implementations are restricted to `#[cfg(test)]` because they cause panics.

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-certs/src/ca/mod.rs
+++ b/crates/stackable-certs/src/ca/mod.rs
@@ -32,7 +32,7 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 /// Defines all error variants which can occur when creating a CA and/or leaf
 /// certificates.
-#[derive(Debug, Snafu)]
+#[derive(Debug, PartialEq, Snafu)]
 pub enum Error {
     #[snafu(display("failed to generate RSA signing key"))]
     GenerateRsaSigningKey { source: rsa::Error },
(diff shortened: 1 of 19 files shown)


crates/stackable-certs/src/ca/mod.rs
```diff
@@ -68,6 +68,48 @@ pub enum Error {
     ParseAuthorityKeyIdentifier { source: x509_cert::der::Error },
 }
 
+/// Custom implementation of [`std::cmp::PartialEq`] because some inner types
+/// don't implement it.
+///
+/// Note that this implementation is restritced to testing because there is a
+/// variant that is impossible to compare, and will cause a panic if it
```
Techassi, line 75: nitpick: Seems like `is` is missing here:

```suggestion
/// variant that is impossible to compare, and will cause a panic if it is
```

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -68,6 +68,48 @@ pub enum Error {
...
+/// attemped.
+#[cfg(test)]
+impl PartialEq for Error {
```
Techassi, line 78: note: It ain't pretty, but it works :)

Techassi (changes requested):
Looking good! Only requires a small change.

As already discussed, we might want to thrive to add more unit tests in the crates themself (where testing might be easier), instead of relying on external tests (in operators). It then might be possible to eliminate some of the manually implemented `PartialEq` implementations.

Techassi (approved):

Issue #694, opened 2023-12-05
tracking: Remove technical debt / framework rework


Techassi:
```[tasklist]
### Tasks
- [ ] Add module-scoped errors with the help of `snafu`
- [ ] Improve doc comments / `schemars` descriptions for our CRDs
- [ ] #617
- [ ] #607
- [ ] #564
```

PR #766, opened 2024-04-16
fix: Remove duplicate authority key identifier extension

Techassi:
Without removing this duplicate extension, the Go X.509 parsing code would error out stating that the "certificate contains duplicate extensions", which was indeed correct.

We accidentally included the authority key identifier twice, once in the leaf profile and once by manually adding the extension after building the cert.

We now removed the manually added extension. This resolved the Go error and the HTTP client was able to establish a TLS-secured connection to the dummy webhook.

See merge: https://go-review.googlesource.com/c/go/+/383215
See code: https://github.com/golang/go/blob/315b6ae682a2a4e7718924a45b8b311a0fe10043/src/crypto/x509/parser.go#L965-L968

---

To be able to access the OID of the duplicate extension, it was required to change Go's standard library. I will provide an upstream patch to improve the error message.

Also see the follow-up issue #694 to add tests / assertions around correct certificate generation.


NickLarsenNZ (approved):
LGTM

PR #769, opened 2024-04-25
chore: Bump k8s compilation version to 1.29. Also bump all dependencies

sbernauer:
# Description

We need to go throw the breaking change https://github.com/kubernetes/kubernetes/pull/118653, which should not break the API (only the internally used structs)

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/src/builder/pod/volume.rs
+++ b/crates/stackable-operator/src/builder/pod/volume.rs
@@ -3,8 +3,8 @@ use k8s_openapi::{
         CSIVolumeSource, ConfigMapVolumeSource, DownwardAPIVolumeSource, EmptyDirVolumeSource,
         EphemeralVolumeSource, HostPathVolumeSource, PersistentVolumeClaim,
         PersistentVolumeClaimSpec, PersistentVolumeClaimTemplate,
-        PersistentVolumeClaimVolumeSource, ProjectedVolumeSource, ResourceRequirements,
-        SecretVolumeSource, Volume, VolumeMount,
+        PersistentVolumeClaimVolumeSource, ProjectedVolumeSource, SecretVolumeSource, Volume,
+        VolumeMount, VolumeResourceRequirements,
     },
     apimachinery::pkg::api::resource::Quantity,
 };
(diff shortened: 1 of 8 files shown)


crates/stackable-operator/src/builder/pod/volume.rs
```diff
@@ -358,9 +358,9 @@ impl SecretOperatorVolumeSourceBuilder {
                 metadata: Some(ObjectMetaBuilder::new().annotations(annotations).build()),
                 spec: PersistentVolumeClaimSpec {
                     storage_class_name: Some("secrets.stackable.tech".to_string()),
-                    resources: Some(ResourceRequirements {
+                    resources: Some(VolumeResourceRequirements {
```
Techassi, line 361: question: This is a non-breaking change for operator users, but a breaking one for library users? The type changes but the (de)serialization is the same, right?

We should mark the changelog entry as breaking in my opinion.

crates/stackable-operator/src/builder/pod/volume.rs
```diff
@@ -358,9 +358,9 @@ impl SecretOperatorVolumeSourceBuilder {
...
                         requests: Some([("storage".to_string(), Quantity("1".to_string()))].into()),
-                        ..ResourceRequirements::default()
+                        ..Default::default()
```
Techassi, line 363: question: I assume the defaults are the same?

crates/stackable-operator/src/commons/affinity.rs
```diff
@@ -85,6 +85,8 @@ pub fn affinity_between_role_pods(
             namespace_selector: None,
             namespaces: None,
             topology_key: TOPOLOGY_KEY_HOSTNAME.to_string(),
+            match_label_keys: None,
```
Techassi, line 88: question: Were these fields added in K8s `1.29`? Setting them to `None` should be fine then.

crates/stackable-operator/src/builder/pod/volume.rs
sbernauer: I mean technically we (as operator-rs) are not making a breaking change and I think we do not re-export k8s-openapi. But I'll add a changelog entry anyways :)

crates/stackable-operator/src/builder/pod/volume.rs
sbernauer: Yep (everything is None)

crates/stackable-operator/src/commons/affinity.rs
sbernauer: Yep, https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#matchlabelkeys

Techassi (approved):

PR #771, opened 2024-04-25
ci: Only run clippy annotation job when PR is not a draft

Techassi:
Only run the clippy annotation job when the PR is not marked as a draft. This prevents the bot from spamming comments / reviews, when the PR is known to be in a broken state.


sbernauer (approved):
Makes sense to me

PR #562, opened 2023-03-05
Support specifying externalTrafficPolicy in Services created by listener-operator

sbernauer:
## Description

Needed by https://github.com/stackabletech/listener-operator/pull/57

## Review Checklist
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added (or not applicable)
- [ ] Documentation added (or not applicable)
- [ ] Changelog updated (or not applicable)
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)

PR #773, opened 2024-05-06
Support specifying externalTrafficPolicy in Services created by listener-operator

sbernauer:
# Description

Replaces https://github.com/stackabletech/operator-rs/pull/562
Needed by https://github.com/stackabletech/listener-operator/pull/57
First raised a year ago [here](https://stackable-workspace.slack.com/archives/C043FMTVBF1/p1677761384978149)

If i use the ListenerClass `external-stable` on IONOS k8s the `LoadBalancer` svc is not reachable. I did find out it's because we set `externalTrafficPolicy: Local` on the svc [here](https://github.com/stackabletech/listener-operator/blob/405502fcf85b1dc7b626e1214c70791dd53b07ad/rust/operator-binary/src/listener_controller.rs#L172).
This PR simply allows users to configure that field, which is than passed to the created Service. We leave the default as-is at `Local`, so we don't have any breaking change.

This change is needed for external-stable to work on IONOS!

[Some background info on that feature](https://blog.getambassador.io/externaltrafficpolicy-local-on-kubernetes-e66e498212f9)

## CRD change
```yaml
---
apiVersion: listeners.stackable.tech/v1alpha1
kind: ListenerClass
metadata:
  name: public
spec:
  serviceType: LoadBalancer
  # NEW:
  serviceExternalTrafficPolicy: Local # or Cluster, defaults to Local
```

```[tasklist]
# Tasks
- [ ] operator-rs change
- [ ] Read and use new field in listener-op
```

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


crates/stackable-operator/src/commons/listener.rs
```diff
@@ -68,6 +68,34 @@ pub enum ServiceType {
...
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, Default, JsonSchema, PartialEq, Eq)]
+pub enum TrafficPolicy {
+    /// Obscures the client source IP and may cause a second hop to another node, but allows Kubernetes to spread the load between all nodes.
+    #[default]
+    Cluster,
+    /// Preserves the client source IP and avoid a second hop for LoadBalancer and NodePort type Services, but makes clients responsible for spreading the load.
+    Local,
```
NickLarsenNZ, line 87: **style**: empty newlines between variants/fields with comments or attributes.

```suggestion
    /// Obscures the client source IP and may cause a second hop to another node, but allows Kubernetes to spread the load between all nodes.
    #[default]
    Cluster,

    /// Preserves the client source IP and avoid a second hop for LoadBalancer and NodePort type Services, but makes clients responsible for spreading the load.
    Local,
```

crates/stackable-operator/src/commons/listener.rs
```diff
@@ -68,6 +68,34 @@ pub enum ServiceType {
...
+
+impl TrafficPolicy {
+    pub fn to_kubernetes_literal(&self) -> String {
+        match self {
+            TrafficPolicy::Cluster => "Cluster".to_string(),
+            TrafficPolicy::Local => "Local".to_string(),
+        }
+    }
+}
+
```
NickLarsenNZ, line 98: I think we can probably just impl `std::string::ToString`? Or is there something special about the Kubernetes literal (if there was, I'd assume we'd be implementing a trait anyway)?

crates/stackable-operator/src/commons/listener.rs
```diff
@@ -68,6 +68,34 @@ pub enum ServiceType {
...
 
+impl ServiceType {
+    pub fn to_kubernetes_literal(&self) -> String {
+        match self {
+            ServiceType::NodePort => "NodePort".to_owned(),
+            ServiceType::LoadBalancer => "LoadBalancer".to_owned(),
+            ServiceType::ClusterIP => "ClusterIP".to_owned(),
+        }
+    }
+}
```
NickLarsenNZ, line 79: I think we can probably just impl `std::string::ToString`? Or is there something special about the Kubernetes literal (if there was, I'd assume we'd be implementing a trait anyway)?

NickLarsenNZ (approved):
LGTM, left some comments

crates/stackable-operator/src/commons/listener.rs
sbernauer: Thinking about this, let's just name it `KubernetesTrafficPolicy` to make clear it's naming is coming from k8s and use strum::Display

crates/stackable-operator/src/commons/listener.rs
```diff
@@ -50,16 +50,19 @@ use crate::builder::pod::volume::ListenerOperatorVolumeSourceBuilder;
...
 }
 
 /// The method used to access the services.
-#[derive(Serialize, Deserialize, Clone, Copy, Debug, JsonSchema, PartialEq, Eq)]
-pub enum ServiceType {
+//
+// Please note that this represents a Kubernetes type, so the name of the enum variant needs to exactly match the
+// Kubernetes service type.
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, JsonSchema, PartialEq, Eq, strum::Display)]
+pub enum KubernetesServiceType {
```
NickLarsenNZ, line 65: **style** (non-blocker): spaces between variants/fields with comments/attributes.

_left it as a non-blocker, since it is existing code_

NickLarsenNZ (approved):
LGTM

crates/stackable-operator/src/commons/listener.rs
NickLarsenNZ: 1e8571c

crates/stackable-operator/src/commons/listener.rs
NickLarsenNZ: 1e8571c

crates/stackable-operator/src/commons/listener.rs
sbernauer: Fixed in 5e6453795b9945d7e3fd8b54fa7f9a6ab0442bc5

NickLarsenNZ (approved):
LGTM

crates/stackable-operator/src/commons/listener.rs
```diff
@@ -50,24 +50,48 @@ use crate::builder::pod::volume::ListenerOperatorVolumeSourceBuilder;
...
+/// See [Kubernetes docs](https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/).
+//
+// Please note that this represents a Kubernetes type, so the name of the enum variant needs to exactly match the
+// Kubernetes traffic policy.
+#[derive(
+    Serialize, Deserialize, Clone, Debug, Default, JsonSchema, PartialEq, Eq, strum::Display,
+)]
+pub enum KubernetesTrafficPolicy {
+    /// Obscures the client source IP and may cause a second hop to another node, but allows Kubernetes to spread the load between all nodes.
+    #[default]
```
nightkr, line 88: I would keep Local as the default. We try pretty hard to shove traffic onto the right node, and we should keep testing that as the primary configuration. Cluster is a fallback option for providers that break Local mode (IONOS so far).

sbernauer: Was a copy/paste error, good catch! Should be fixed in 1060548c728ef78e6b98030243b8e1965e0fc72c

nightkr: Nit: I'd shove it into the `impl` block, like we do for the other defaults.

sbernauer: Makes sense. I also switched from  #[schemars(default )] to  #[serde(default )], I hope that is ok

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Support specifying externalTrafficPolicy in Services created by listener-operator ([#773]).
+- BREAKING: Remove `Default` impl for `ListenerSpec` ([#773]).
```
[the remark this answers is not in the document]
sbernauer: 51a46faa8b0c7bbf72604b9665cc4411065bebd9

crates/stackable-operator/src/commons/listener.rs
```diff
@@ -100,14 +123,28 @@ pub struct ListenerSpec {
...
 }
 
 impl ListenerSpec {
     const fn default_publish_not_ready_addresses() -> Option<bool> {
         Some(true)
     }
+
+    // `#[serde(default)]` only supports functions, so we need to wrap the literal in a function :)
+    fn default_service_external_traffic_policy() -> KubernetesTrafficPolicy {
+        Self::default().service_external_traffic_policy
```
nightkr, line 146: It's overkill to construct the whole object just to take the SETP, better to have the derivative default reference the method too instead.

crates/stackable-operator/src/commons/listener.rs
```diff
@@ -100,14 +123,28 @@ pub struct ListenerSpec {
...
     /// Whether incoming traffic should also be directed to Pods that are not `Ready`.
-    #[schemars(default = "Self::default_publish_not_ready_addresses")]
+    #[serde(default = "ListenerSpec::default_publish_not_ready_addresses")]
     pub publish_not_ready_addresses: Option<bool>,
+
+    /// `externalTrafficPolicy` that should be set on the [`Service`] object.
+    ///
+    /// The default is `Local` (in contrast to `Cluster`), as we try pretty hard to shove traffic onto the right node,
+    /// and we should keep testing that as the primary configuration. Cluster is a fallback option for providers that
+    /// break Local mode (IONOS so far).
```
nightkr, line 133: Probably want to reword this if we're putting it in user facing docs.

NickLarsenNZ: Perhaps:

```
as we aim to direct traffic to the nearest node
````

_Other edits needed, but I didn't get to them_

nightkr: "to a node running the workload" perhaps.

sbernauer: Please have a look at f59ff2dfe796d91e5dd18fe747faf31266d2930b

crates/stackable-operator/src/commons/listener.rs
sbernauer: Didn't get derivative to pull a function value at the first attempt. It now works, so 0ff36cf

sbernauer:
There were no further comments during the voting phase, merging this

PR #778, opened 2024-05-07
fix(stackable-webhook): Explicitly set the TLS provider for the ServerConfig, and enable "safe" protocols

NickLarsenNZ:
Rustls now requires a global provider to be set, or explicitly passed in via the server/client config.

Normally this would be set by the application, and not in a library, however this library is to remove so much boiler plate from our applications.

# Description

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


crates/stackable-webhook/src/tls.rs
```diff
@@ -97,7 +103,10 @@ impl TlsServer {
             .private_key_der()
             .context(EncodePrivateKeyDerSnafu)?;
 
-        let mut config = ServerConfig::builder()
+        let tls_provider = default_provider();
+        let mut config = ServerConfig::builder_with_provider(tls_provider.into())
+            .with_safe_default_protocol_versions()
```
NickLarsenNZ, line 108: This enables both TLS1.2 and TLS1.3. I didn't put much thought into this option.

Alternatively, select the versions via [`with_protocol_versions(...)`](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html#method.with_protocol_versions)

Techassi: Using `with_protocol_versions` is a little more explicit. I'm not 100% sure if we need to be that explicit, or if using the default ones (TLS1.2 and TLS1.3) is fine.

Techassi (dismissed):

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Done in c3122c9

Techassi (approved):

PR #790, opened 2024-05-21
feat(stackable-versioned): Generate From implementations for automatic versions upgrades

Techassi:
# Description

Tracked by https://github.com/stackabletech/issues/issues/507, follow-up of #764, ~blocked by #793~

This PR adds code generation for `From<OLD> for NEW` implementations to enable conversion between versions. The macro also allows to implement the `From` trait by hand if customization is required.

Additionally, it adds support to customize the default function (used in the automatically generated `From` impl) for added fields.

### Generated `From` implementations

```rust
#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1beta1"),
    version(name = "v1")
)]
pub struct Foo {
    #[versioned(
        added(since = "v1beta1"),
        deprecated(since = "v1", note = "not needed")
    )]
    deprecated_bar: usize,
    baz: bool,
}

// Produces ...

#[automatically_derived]
pub mod v1alpha1 {
    pub struct Foo {
        pub baz: bool,
    }
}
#[automatically_derived]
#[allow(deprecated)]
impl From<v1alpha1::Foo> for v1beta1::Foo {
    fn from(__sv_foo: v1alpha1::Foo) -> Self {
        Self {
            bar: std::default::Default::default(),
            baz: __sv_foo.baz,
        }
    }
}
#[automatically_derived]
pub mod v1beta1 {
    pub struct Foo {
        pub bar: usize,
        pub baz: bool,
    }
}
#[automatically_derived]
#[allow(deprecated)]
impl From<v1beta1::Foo> for v1::Foo {
    fn from(__sv_foo: v1beta1::Foo) -> Self {
        Self {
            deprecated_bar: __sv_foo.bar,
            baz: __sv_foo.baz,
        }
    }
}
#[automatically_derived]
pub mod v1 {
    pub struct Foo {
        #[deprecated = "not needed"]
        pub deprecated_bar: usize,
        pub baz: bool,
    }
}

// Which can be used like this ...

let foo_v1alpha1 = v1alpha1::Foo { baz: true };
let foo_v1beta1 = v1beta1::Foo::from(foo_v1alpha1);
let foo_v1 = v1::Foo::from(foo_v1beta1);

assert_eq!(foo_v1.deprecated_bar, 0);
assert!(foo_v1.baz);
```

### Skip generating `From` implementations

The generation can be skipped at the container level (no implementations are generated) or at the version level (only the generation from that to the next version is skipped).

```rust
#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1beta1"),
    version(name = "v1"),
    options(skip(from))
)]
pub struct Foo {
    #[versioned(
        added(since = "v1beta1"),
        deprecated(since = "v1", note = "not needed")
    )]
    deprecated_bar: usize,
    baz: bool,
}

// Or ...

#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1beta1", skip(from)),
    version(name = "v1")
)]
pub struct Foo {
    #[versioned(
        added(since = "v1beta1"),
        deprecated(since = "v1", note = "not needed")
    )]
    deprecated_bar: usize,
    baz: bool,
}
```

### Custom default function for added fields

```rust
#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1beta1"),
    version(name = "v1")
)]
pub struct Foo {
    #[versioned(
        added(since = "v1beta1", default = "default_bar"),
        deprecated(since = "v1", note = "not needed")
    )]
    deprecated_bar: usize,
    baz: bool,
}

// Can be used like this ...

fn default_bar() -> usize {
    42
}

let foo_v1alpha1 = v1alpha1::Foo { baz: true };
let foo_v1beta1 = v1beta1::Foo::from(foo_v1alpha1);

assert_eq!(foo_v1beta1.bar, 42);
assert!(foo_v1beta1.baz);
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


crates/stackable-versioned-macros/tests/from.rs
NickLarsenNZ, line 1: note: the `skip_*` tests don't have assertions. I assume they are there to ensure code compiles.

crates/stackable-versioned/CHANGELOG.md
```diff
@@ -4,12 +4,20 @@ All notable changes to this project will be documented in this file.
...
+
+[#784]: https://github.com/stackabletech/operator-rs/pull/784
+[#790]: https://github.com/stackabletech/operator-rs/pull/790
+[#793]: https://github.com/stackabletech/operator-rs/pull/793
+
 - Improve action chain generation ([#784]).
 
-[#784](https://github.com/stackabletech/operator-rs/pull/784)
-[#CHANGEME](https://github.com/stackabletech/operator-rs/pull/CHANGEME)
+[#784](ttps://github.com/stackabletech/operator-rs/pull/784)
```
NickLarsenNZ, line 20: This is a duplicate line, and it is missing the `h` in http.

Techassi: Fixed in e8cc4d1

crates/stackable-versioned-macros/tests/from.rs
Techassi: Yeah there isn't really anything to assert. We would have to assert that code _isn't_ generated. We could maybe do that by using the `#[should_panic]` attribute.

NickLarsenNZ: From memory, there was a thing for testing generated code. Will see if I can't find it, but it could have been for something else.

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -8,6 +8,222 @@ mod attrs;
...
+/// # use stackable_versioned_macros::versioned;
+/// #[versioned(
+///     version(name = "v1alpha1", deprecated)
+/// )]
+/// struct Foo {}
+/// ```
+///
+/// Additionally, it is ensured that each version is unique. Declaring the same
+/// version multiple times will result in an error. Furthermore, declaring the
+/// versions out-of-order ist prohibited by default. It is possible to opt-out
```
NickLarsenNZ, line 58:
```suggestion
/// versions out-of-order is prohibited by default. It is possible to opt-out
```

crates/stackable-versioned-macros/tests/from.rs
Techassi: I eventually want so try if we can use snapshot testing for making sure the generated code matches a known output.

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -8,6 +8,222 @@ mod attrs;
...
+/// )]
+/// struct Foo {
+///     #[versioned(
+///         added(since = "v1beta1"),
+///         renamed(since = "v1", from = "gau"),
+///         deprecated(since = "v2", note = "not empty")
+///     )]
+///     deprecated_bar: usize,
+///     baz: bool,
+/// }
```
NickLarsenNZ, line 103: At a quick glance, this looks the same as the first code block.

NickLarsenNZ (changes requested):

crates/stackable-versioned-macros/src/lib.rs
Techassi: Removed in 3be5393

NickLarsenNZ (approved):
LGTM

crates/stackable-versioned-macros/tests/from.rs
NickLarsenNZ: I was thinking of [trybuild](https://docs.rs/trybuild/latest/trybuild/)

PR #796, opened 2024-05-24
fix(stackable-telemetry): Disable export of h2 events via OTLP

NickLarsenNZ:
It causes infinite cascading events to be emitted. 

See https://github.com/open-telemetry/opentelemetry-rust/issues/761


sbernauer (approved):
Seems to make sense to me. Is the stackable-telemetry crate already used?

PR #799, opened 2024-05-28
Add labels to listener volume builder

razvan:
# Description

Part of https://github.com/stackabletech/issues/issues/555

Required for https://github.com/stackabletech/hdfs-operator/pull/534

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

```[tasklist]
# Author
- [x] Changes are OpenShift compatible
- [x] CRD changes approved
- [x] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [x] Code contains useful comments
- [x] Documentation added or updated
- [x] Changelog updated
- [x] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

PR #798, opened 2024-05-27
chore: Add functionality to convert LogLevel to an OPA log level

xeniape:
# Description

This PR adds a function to `LogLevel` which converts the `LogLevel` to a log level String understood by OPA.

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

```[tasklist]
# Author
- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

PR #800, opened 2024-06-03
chore: Release stackable-operator 0.68.1

Techassi:
Release stackable-operator 0.68.1 which includes these changes:

- Add functionality to convert LogLevel to an OPA log level ([#798]).
- Add labels to listener volume builder ([#799]).

[#798]: https://github.com/stackabletech/operator-rs/pull/798
[#799]: https://github.com/stackabletech/operator-rs/pull/799

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2842,7 +2842,7 @@ dependencies = [
 
 [[package]]
 name = "stackable-operator"
-version = "0.68.0"
+version = "0.68.1"
 dependencies = [
  "chrono",
  "clap",
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [0.68.1] - 2024-06-03
+
 ### Added
 
 - Add functionality to convert LogLevel to an OPA log level ([#798]).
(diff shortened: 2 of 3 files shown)


sbernauer (changes requested):
We should use 0.69.0 as there are new features and especially a breaking change in this.

Please also mark https://github.com/stackabletech/operator-rs/pull/799 as BREAKING in the changelog.

Techassi:
I wouldn't classify adding labels to PVCs a breaking change. If it is breaking, it should've been marked as such in the original PR (#799).

sbernauer:
It was missed in the PR and I wasn't quick enough to review :)

I would use something like

```
- BREAKING: Add labels to listener volume builder.
  `PodBuilder::add_listener_volume_by_listener_class`, `PodBuilder::add_listener_volume_by_listener_name` and `ListenerOperatorVolumeSourceBuilder::new` now require you to pass the labels for the created volumes ([#799]).
```

Techassi:
Ah well, another instance where manually maintaining the changelog file caused an issue. I guess we consider the changes as breaking because we altered the API. It shouldn't be breaking from a functionality point-of-view.

I will adjust the changelog and the version.

---

It might be worth to differentiate between these two instances. Should we open decision / discussion for that?

sbernauer:
> Ah well, another instance where manually maintaining the changelog file caused an issue.

The label `changelog/breaking` was also missing, I don't see how automation could have prevented that :see_no_evil:
Maybe there is a tool out there that automatically detects breaking changes, but that might be an overkill for our small lib with mostly only us as users.

> Should we open decision / discussion for that?

Would be great, but I'm personally happy as is, given our limited time resources. For me it's important to state *why* this change is breaking, than users can decide on there own if they care or not

sbernauer (approved):

PR #804, opened 2024-06-06
feat(stackable-versioned): Add collision check for renamed fields

Techassi:
Tracked by https://github.com/stackabletech/issues/issues/507

This PR adds a check to ensure that renamed fields don't collide in any of the declared versions. This is currently not ensured and can lead to two or more fields using the same name in a particular version.


NickLarsenNZ (approved):
LGTM

PR #806, opened 2024-06-07
fix(stackable-telemetry/webhook): Correctly extract connect and host info in Axum trace layer

Techassi:
This correctly extract the `server.address`, `server.port`, `client.address` and `client.port` fields in the OpenTelemetry compatible traces.

This fix requires changes in the `stackable-webhook` crate to instruct various parts of the connection handling chain to extract the required data, which can be consumed by the `stackable-telemetry` crate.


crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -242,6 +286,28 @@ impl RequestExt for Request {
...
+        // it includes a non-default port.
+        let port: u16 = port.parse().context(ParsePortSnafu)?;
+        Ok((host.to_owned(), port))
+    } else {
+        // If there is no port included in the header value, the port is implied.
+        // Port 443 for HTTPS and port 80 for HTTP.
+        let port = match scheme {
+            Some("https") => DEFAULT_HTTPS_PORT,
+            Some("http") => DEFAULT_HTTP_PORT,
+            _ => return InvalidSchemeSnafu.fail(),
```
NickLarsenNZ, line 304: ...following on from a previous suggestion:

```suggestion
            Some(scheme) => return InvalidSchemeSnafu { scheme }.fail(),
            _ => return InvalidSchemeSnafu { scheme: String::new() }.fail(),
```

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -242,6 +286,28 @@ impl RequestExt for Request {
...
 }
 
+fn server_host_to_tuple(
+    host: &str,
+    scheme: Option<&str>,
+) -> Result<(String, u16), ServerHostError> {
+    if let Some((host, port)) = host.split_once(':') {
+        // First, see if the host header value contains a colon indicating that
+        // it includes a non-default port.
+        let port: u16 = port.parse().context(ParsePortSnafu)?;
```
NickLarsenNZ, line 296: ...following on from a previous suggestion:

```suggestion
        let port: u16 = port.parse().context(ParsePortSnafu { port })?;
```

_untested, maybe there is a borrow issue here_

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -338,6 +404,7 @@ impl SpanExt for Span {
             otel.status_message = Empty,
             http.request.method = http_method,
             http.response.status_code = Empty,
+            http.route = Empty,
```
NickLarsenNZ, line 407: **note (non-blocker)**: with the attribute keys, as soon as we have one with a Rust reserved word, we need to wrap them in quotes.

Eg: `network.type` (type being a reserved word).

With a bit of luck, something like https://github.com/tokio-rs/tracing/pull/2254 can be implemented soon so that we can consistently use the Semantic Convention consts in these span definitions as well as in the `record` functions.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -363,9 +428,12 @@ impl SpanExt for Span {
...
 
-        if let Some(server_socket_address) = req.server_socket_address() {
-            span.record("server.address", server_socket_address.ip().to_string())
-                .record("server.port", server_socket_address.port());
+        if let Ok((host, port)) = req.server_host() {
+            // NOTE (@Techassi): We cast to i64, because otherwise the field
+            // will NOT be recorded as a number but as a string. This is likely
+            // an issue in the tracing-opentelemetry crate.
+            span.record("server.address", host)
+                .record("server.port", port as i64);
```
NickLarsenNZ, line 436: **note (non-blocker)**: As discussed via DM, we will not use the Semantic Convention consts in the `record` functions until they can be used in the `Span` definitions (and `instrument(fields(...))` too.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -375,25 +443,22 @@ impl SpanExt for Span {
             span.record("client.address", client_socket_address.ip().to_string());
 
             if opt_in {
-                span.record("client.port", client_socket_address.port());
+                // NOTE (@Techassi): We cast to i64, because otherwise the field
+                // will NOT be recorded as a number but as a string. This is
+                // likely an issue in the tracing-opentelemetry crate.
+                span.record("client.port", client_socket_address.port() as i64);
```
NickLarsenNZ, line 449: **note (non-blocker)**: IMO, we should get rid of the `opt-in` concept here, and just decide on the fields we record.

The whole opt-in thing was more for implementers (in this case, us) to choose which of those they want to add to instrumentation.

crates/stackable-webhook/src/lib.rs
```diff
@@ -173,6 +173,7 @@ impl WebhookServer {
     async fn run_server(self) -> Result<()> {
         debug!("run webhook server");
 
+        // TODO (@Techassi): Make opt-in configurable from the outside
```
NickLarsenNZ, line 176: I would prefer we drop the concept of opt-in fields all together. We should just decide which additional fields are relevant, and record them.

crates/stackable-webhook/src/tls.rs
```diff
@@ -139,10 +139,24 @@ impl TlsServer {
...
+        // doing that. Calling `into_make_service_with_connect_info` returns a
+        // new struct `IntoMakeServiceWithConnectInfo` which implements the
+        // Tower Service trait. This service is called after the TCP connection
+        // has been accepted.
+        //
+        // Inspired by:
+        // - https://github.com/tokio-rs/axum/discussions/2397
+        // - https://github.com/tokio-rs/axum/blob/b02ce307371a973039018a13fa012af14775948c/examples/serve-with-hyper/src/main.rs#L98
+
+        let mut router = self
```
NickLarsenNZ, line 153: **question**: As I understand this, `router: IntoMakeServiceWithConnectInfo`. What was the type before?

NickLarsenNZ (dismissed):
LGTM, left some comments and suggestions, none of which are blockers.

crates/stackable-webhook/src/lib.rs
NickLarsenNZ: I'll mark as resolved, as it is covered here: https://github.com/stackabletech/operator-rs/pull/806#discussion_r1636282926

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -30,6 +34,10 @@ mod injector;
 pub use extractor::*;
 pub use injector::*;
 
+const X_FORWARDED_HOST_HEADER_KEY: &str = "X-Forwarded-Host";
+const DEFAULT_HTTPS_PORT: u16 = 443;
+const DEFAULT_HTTP_PORT: u16 = 80;
```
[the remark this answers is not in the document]
Techassi: Yes, I'm also surprised. I also had a quick look and also didn't find anything.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -163,13 +171,25 @@ where
     }
 }
 
+#[derive(Debug, Snafu)]
+pub enum ServerHostError {
+    #[snafu(display("failed to parse port as u16 from string"))]
+    ParsePort { source: ParseIntError },
```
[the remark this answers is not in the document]
Techassi: Nice idea! I will add it.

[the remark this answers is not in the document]
Techassi: Added in 47ffb40

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -163,13 +171,25 @@ where
...
+
+    #[snafu(display("encountered invalid request scheme"))]
+    InvalidScheme,
```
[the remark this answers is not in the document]
Techassi: Added in 47ffb40

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: Added in 47ffb40. I also turned it into an Option and combined it with `unwrap_or_default()`.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: Added in 47ffb40

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: Sure, I can remove the opt-in feature. In the future, this might become a problem tho. If we can support recording the HTTP request headers as fields, this can result in a large set of additional fields.

We can also remove it for now and decide later, if we need such a mechanism and in which form.

crates/stackable-webhook/src/tls.rs
Techassi: Before, it was just `router: Router` iirc.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: Yes, that feature would be really nice to have. I have planned to invest some time in the upstream project if needed.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: Added quotes in d41179b

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: Added quotes in d41179b

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
NickLarsenNZ: > If we can support recording the HTTP request headers as fields, this can result in a large set of additional fields.

We could make that decision on whether we want to record such detail. Also, there are protections to limit the amount of attributes that can be sent in the span.

Nothing needed right now, we can remove it later (if we want).

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
NickLarsenNZ: Updated in fbb65ef2fb05ff501849c319e195bfc052231e54

NickLarsenNZ (approved):
LGTM

PR #811, opened 2024-06-12
feat: More granular instrumentation and remove long running spans

NickLarsenNZ:
# Description

- [x] Remove long running spans.
    - They are not so useful, and cause trace visualization applications to report the missing parent span for its children.
- [x] Add more granular spans for futures.
- [x] Carry across context between the tls server handling connections, and the axum router endpoints.
    - Without this, two separate traces appear instead of a single trace with a hierarcy of spans from the low level tcp listener up to the route handlers.
- [x] Use fully qualified tracing events and spans (per: https://github.com/stackabletech/decisions/issues/15).
- [x] Update crates (pending: https://github.com/open-telemetry/opentelemetry-rust/issues/1880)

### Full chain

> [!NOTE]
> All spans are in the same trace when the client has not sent a trace ID in the request.

image

image

### Linked traces

When a client sends trace information, we can pick that up and use the trace context from there. Notice the traceparent send is the same one as what is received. The webhook is using the clients trace context from thereon.

image

> [!NOTE]
> Unfortunately trace ids cannot be updated on spans, so once we are able to extract the trace-id from the headers of a client request, we can only use it from that point forward.
> But, we are able to add links to spans to make life easier.

Client makes a request, and injects the trace-id into headers.
A link appears on the span where the parent context was updated.

image

Clicking on the link will take you to the trace with the spans that existed before the trace-id was extracted from the headers.

image

This trace also has a span with a link back to the other.

Similarly in Jaeger

image

```[tasklist]
# Reviewer
- [x] Changelog updated
```


crates/stackable-webhook/src/tls.rs
```diff
@@ -171,32 +172,98 @@ impl TlsServer {
...
-                    .serve_connection_with_upgrades(tls_stream, hyper_service)
-                    .await
-                {
-                    warn!(%err, %remote_addr, "failed to serve connection");
+            let span = tracing::debug_span!("accept tcp connection");
+            tokio::spawn(
+                async move {
+                    let span = tracing::trace_span!(
+                        "accept tls connection",
+                        // otel.name = "accept tls connection",
```
Techassi, line 180: nit: Remove that line. Maybe add a comment stating why we don't need that field here.

crates/stackable-webhook/src/tls.rs
```diff
@@ -171,32 +172,98 @@ impl TlsServer {
...
+                        "server.address" = Empty,
+                        "server.port" = Empty,
+                        "client.address" = Empty,
+                        "client.port" = Empty,
+                        "network.local.address" = Empty,
+                        "network.local.port" = Empty,
+                        "network.peer.address" = Empty,
+                        "network.peer.port" = Empty,
+                        // network.protocol.name = Empty, // at this layer, we don't know the application protocol
+                        // network.protocol.version = Empty, // doesn't make sense for tcp
```
Techassi, line 193: nit: Remove those fields and add the comments at the top.

crates/stackable-webhook/src/tls.rs
```diff
@@ -171,32 +172,98 @@ impl TlsServer {
...
+                        "network.transport" = "tcp",
+                        "network.type" = self.socket_addr.semantic_convention_network_type(),
+                    );
+
+                    if let Ok(local_addr) = tcp_stream.local_addr() {
```
Techassi, line 198: question: Does the local address ever change during runtime? If not, we could retrieve the address once instead of retrieving it for every incoming connection.

crates/stackable-telemetry/src/tracing.rs
```diff
@@ -186,15 +181,9 @@ impl Tracing {
...
-                .with_trace_config(
-                    trace::config()
-                        .with_sampler(Sampler::AlwaysOn) // TODO (@NickLarsenNZ): Make this configurable. See also Sampler::ParentBased
-                        .with_id_generator(RandomIdGenerator::default())
-                        .with_resource(Resource::new(vec![KeyValue::new(
-                            resource::SERVICE_NAME,
-                            self.service_name,
-                        )])),
-                )
+                .with_trace_config(trace::config().with_resource(Resource::new(vec![
```
Techassi, line 184: question: Was the sampler not needed anymore. We discussed explicitly setting the ID generator, because it uses the random ID generator by default.

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -363,11 +362,13 @@ pub trait SpanExt {
 
 impl SpanExt for Span {
     fn from_request(req: &Request, opt_in: bool) -> Self {
+        let _otel_context_guard = Context::current();
```
Techassi, line 365: question: I'm curious why this is needed. Shouldn't the following spans / events use the current context by default?

Techassi (commented):
LGTM, just some questions and nitpicks.

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Oh, I thought I had removed it. I think I got stuck wondering what the otel lib does when it is missing. My assumption is that it uses the span name. Either way, looks fine in Grafana/Jaeger.

I think I'll just remove it. We can add an explanation when we do set it perhaps?

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Should I just remove them completely? Comments and all?
They were leftovers from development.

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: The local address is whatever the connection happened on, so yes it _can_ change during runtime, but can't change per connection.

Further explanation: If the app listens on 0.0.0.0, then clients can hit any IP on any reachable interface.

crates/stackable-webhook/src/tls.rs
Techassi: > Either way, looks fine in Grafana/Jaeger.

Yeah it seems to work fine. I think `otel.name` is only needed when the span name needs to be dynamic.

> I think I'll just remove it. We can add an explanation when we do set it perhaps?

Agreed, that makes more sense.

crates/stackable-webhook/src/tls.rs
Techassi: > Further explanation: If the app listens on 0.0.0.0, then clients can hit any IP on any reachable interface.

Ah, you are right. I didn't think about that possibility.

crates/stackable-telemetry/src/tracing.rs
NickLarsenNZ: The default ID Generator is the [RandomIdGenerator], so it seems redundant to make it explicit. The only other generate I know of is the [XrayIdGenerator] (maybe it needs to be configurable one day, but we should wait for the need).

As for the sampler, the default uses [ParentBased] and falling back to [AlwaysOn] (it allows parent spans that might be probability based to also pass that info on).

Default impl can be found here: https://github.com/open-telemetry/opentelemetry-rust/blob/b933bdd82dadbadedb42a1f572caba1e0b8b9391/opentelemetry-sdk/src/trace/config.rs#L95-L96

[RandomIdGenerator]: https://docs.rs/opentelemetry_sdk/0.22.0/opentelemetry_sdk/trace/struct.RandomIdGenerator.html
[XrayIdGenerator]: https://docs.rs/opentelemetry_sdk/0.22.0/opentelemetry_sdk/trace/struct.XrayIdGenerator.html
[ParentBased]: https://docs.rs/opentelemetry_sdk/0.22.0/opentelemetry_sdk/trace/enum.Sampler.html#variant.ParentBased
[AlwaysOn]: https://docs.rs/opentelemetry_sdk/0.22.0/opentelemetry_sdk/trace/enum.Sampler.html#variant.AlwaysOn

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
NickLarsenNZ: Yes, but as you and I both saw, the context wasn't being carried across and the Axum route spans then had no parent and generated new trace IDs, thereby splitting one trace into two separate traces.

This was based on how it worked before #806 was merged, so I haven't gone back to see if it might not be needed now.

What do you suggest here?

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Futhermore, the local_addr is only used here, and since it is modifying the span, it needs to be below it.

An alternative would be to deal with it directly on span creation, like:

```rust
                    let span = tracing::trace_span!(
                        "accept tls connection",
                        // otel.name = "accept tls connection",
                        "otel.kind" = ?SpanKind::Server,
                        "otel.status_code" = Empty,
                        "otel.status_message" = Empty,
                        "server.address" = tcp_stream.local_addr().unwrap_or(Empty).ip().to_string(),
                        "server.port" = tcp_stream.local_addr().unwrap_or(Empty).port() as u64,
                        "client.address" = Empty,
                        "client.port" = Empty,
                        "network.local.address" = tcp_stream.local_addr().unwrap_or(Empty).ip().to_string(),
                        "network.local.port" = tcp_stream.local_addr().unwrap_or(Empty).port() as u64,
                        "network.peer.address" = Empty,
                        "network.peer.port" = Empty,
                        "network.transport" = "tcp",
                        "network.type" = self.socket_addr.semantic_convention_network_type(),
                    );
```

But that seems messy.

NickLarsenNZ: ^ Actually that doesn't work, because the unwrap_or wants a SocketAddr

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Done in 9cf0316cae86aa34424c49171b2ee80fde01bb38

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Done in 9cf0316cae86aa34424c49171b2ee80fde01bb38

crates/stackable-webhook/src/tls.rs
Techassi: Yeah, that is messy indeed. Let's leave it as is.

crates/stackable-telemetry/src/tracing.rs
Techassi: Thank you for the great explanation!

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: > What do you suggest here?

If you really want, you can test it again. In case we still need it, we should add a comment stating why we need it - so that we don't stumble over this in the future.

NickLarsenNZ: It seems not needed.

Removed in d5d90bd0081d711002c87c15bb9d01ac1eb173fe

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -418,9 +419,42 @@ impl SpanExt for Span {
...
+        if new_span_context != current_span_context {
+            tracing::trace!(
+                opentelemetry.trace_id.from = ?current_span_context.trace_id(),
+                opentelemetry.trace_id.to = ?new_span_context.trace_id(),
+                "set parent span context based on context extracted from request headers"
+            );
+            Span::current().add_link(new_parent.span().span_context().clone());
+            span.add_link(Context::current().span().span_context().to_owned());
+            span.set_parent(new_parent);
+        }
```
Techassi, line 457: nitpick(style): Add newlines here:

```suggestion
        let current_span_context = Context::current().span().span_context().clone();
        
        if new_span_context != current_span_context {
            tracing::trace!(
                opentelemetry.trace_id.from = ?current_span_context.trace_id(),
                opentelemetry.trace_id.to = ?new_span_context.trace_id(),
                "set parent span context based on context extracted from request headers"
            );
            
            Span::current().add_link(new_parent.span().span_context().clone());
            span.add_link(Context::current().span().span_context().to_owned());
            span.set_parent(new_parent);
        }
```

crates/stackable-telemetry/src/tracing.rs
```diff
@@ -41,6 +45,8 @@ pub enum Error {
 ///
 /// #[tokio::main]
 /// async fn main() -> Result<(), Error> {
+///     // IMPORTANT: Name the guard variable appropriately, do not just use
```
Techassi, line 48: suggestion: Let's include this notice in the doc comment directly (maybe in addition to this note).

Techassi (changes requested):
LGTM, just some small remarks.

crates/stackable-telemetry/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 - Add support for setting the environment variable for each configured tracing subscriber ([#801]).
 - Use OpenTelemetry Context in Axum instrumentation layer, adjust log and span level, simplify trace config ([#811]).
+  - tracing: Upgrade opentelemetry crates, simplify trace config, fix shutdown conditions, use new way to shutdown LoggerProvider.
+  - instrumentation/axum: demote event severity for errors easily caused by clients, replace parent span context if given in http header and link to previous trace contexts
```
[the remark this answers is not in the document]
NickLarsenNZ: b494f1c

crates/stackable-operator/CHANGELOG.md
```diff
@@ -12,12 +12,14 @@ All notable changes to this project will be documented in this file.
   - [Blog Post - Breaking Change](https://kube.rs/blog/2024/06/11/watcher-memory-improvements/#breaking-change)
   - [kube#1494](https://github.com/kube-rs/kube/pull/1494)
   - [kube#1504](https://github.com/kube-rs/kube/pull/1504)
+- Upgrade opentelemetry crates ([#811])
```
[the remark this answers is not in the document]
NickLarsenNZ: b494f1c

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
NickLarsenNZ: 51ad319

PR #815, opened 2024-06-21
fix(stackable-webhook): Move certificate generation to the blocking thread

NickLarsenNZ:
> [!NOTE]
> Easier to review with whitespace hidden: https://github.com/stackabletech/operator-rs/pull/815/files?diff=split&w=1

# Description

As per https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html,

> In general, issuing a blocking call or performing a lot of compute in a future without yielding is problematic, as it may prevent the executor from driving other futures forward. This function runs the provided closure on a thread dedicated to blocking operations. See the [CPU-bound tasks and blocking code](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) section for more information.

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

### DIFF
--- a/crates/stackable-webhook/src/tls.rs
+++ b/crates/stackable-webhook/src/tls.rs
@@ -56,6 +56,9 @@ pub enum Error {
 
     #[snafu(display("failed to set safe TLS protocol versions"))]
     SetSafeTlsProtocolVersions { source: tokio_rustls::rustls::Error },
+
+    #[snafu(display("failed to run task in blocking thread"))]
+    TokioSpawnBlocking { source: tokio::task::JoinError },
 }
 
 /// Custom implementation of [`std::cmp::PartialEq`] because some inner types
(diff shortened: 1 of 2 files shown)


crates/stackable-webhook/src/tls.rs
```diff
@@ -94,37 +97,43 @@ pub struct TlsServer {
...
-            .context(InvalidTlsPrivateKeySnafu)?;
-
-        config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
-        let config = Arc::new(config);
-
-        Ok(Self {
-            socket_addr,
-            config,
-            router,
+        let task = tokio::task::spawn_blocking(move || {
```
Techassi, line 100: suggestion: Let's add a comment here, why we use the `spawn_blocking` method. And let's also add the link you provided in the PR description here.

Techassi (approved):
LGTM, just one minor suggestion.

crates/stackable-webhook/src/tls.rs
NickLarsenNZ: Yeah boi

NickLarsenNZ: Done in f6902446c614ad83e9c3ceec78d36df66b09f279

Techassi (approved):

PR #817, opened 2024-06-26
feat/misc utils

razvan:
# Description

Needed for https://github.com/stackabletech/hbase-operator/pull/506

Useful for operators that need to support incompatible product versions.

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
   - [kube#1494](https://github.com/kube-rs/kube/pull/1494)
   - [kube#1504](https://github.com/kube-rs/kube/pull/1504)
 - Upgrade opentelemetry crates ([#811]).
+- Added `ProductImage.product_version()` utility function ([#817])
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


nightkr (approved):
Functionality LGTM, could use a more descriptive PR title

sbernauer (commented):
Honestly I would like to object this change. Yes, currently we can save ~ 10 CPU cycles :D, but in the future we will have
not only automatic stackable version detection, but the operator might also automatically pick the product version for you (probably from the LTS line).
This might involve network calls or some constants compiled in the code or something completely else.

I would prefer to not build on the fact that the product version determination is currently cheap (n network call) and simple.

For now I would prefer sticking to `product_image.resolve().product_version`.

razvan:
I can revert if this is a veto. No problem.

Though I agree that the code comment insinuates this is about saving CPU cycles, it was more about failing fast and having concise code.

sbernauer:
I guess you are right, I forgot about the need of passing `image_base_name: &str, operator_version: &str` to `resolve`.
So let's keep the API as you have introduced it ;)

I created a PR for a small refactoring, would be awesome if you could have a look! https://github.com/stackabletech/operator-rs/pull/818

PR #820, opened 2024-06-28
refactor(stackable-versioned): Remove duplicate code

Techassi:

This PR unifies struct/enum and field/variant code. This allows common parts to be shared and only implemented once. ~~Currently blocked until #813 is merged.~~

```[tasklist]
# Reviewer
- [x] Changelog updated
```


crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -223,7 +223,7 @@ pub fn versioned(attrs: TokenStream, input: TokenStream) -> TokenStream {
     // Module(ItemMod).
     let input = syn::parse_macro_input!(input as DeriveInput);
 
-    gen::expand(attrs, input)
+    codegen::expand(attrs, input)
```
NickLarsenNZ, line 226: Now I have to make the same change in `protogen` :sweat_smile:

Techassi: I made the change, because in the future `gen` will be a reserved keyword and would require us to use `use crate::r#gen` for imports. Also, in doc comments `r#gen` didn't link to the item. I don't know if there is an upstream issue for that, instead I decided to rename the module to make that work again.

crates/stackable-versioned-macros/src/codegen/common/item.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+/// This trait enables access to the common attributes across field and variant
+/// attributes.
+pub(crate) trait Attributes {
+    fn common_attrs_owned(self) -> ItemAttributes;
+    fn common_attrs(&self) -> &ItemAttributes;
+}
+
+/// This struct combines common common code for versioned fields and variants.
```
NickLarsenNZ, line 63:
```suggestion
/// This struct combines common code for versioned fields and variants.
```

crates/stackable-versioned-macros/src/codegen/common/item.rs
```diff
@@ -0,0 +1,315 @@
...
+        let attrs = A::try_from(&item)?;
+        attrs.validate_versions(container_attrs, &item)?;
+
+        let item_attrs = attrs.common_attrs_owned();
+
+        // Constructing the action chain requires going through the actions
+        // starting at the end, because the base container always represents the
+        // latest (most up-to-date) version of that struct. That's why the
+        // following code needs to go through the actions in reverse order, as
+        // otherwise it is impossible to extract the item ident for each version.
```
NickLarsenNZ, line 112:
```suggestion
        // Constructing the action chain requires going through the actions
        // starting at the end, because the container definition always
        // represents the latest (most up-to-date) version of that struct.
        // That's why the following code needs to go through the actions in
        // reverse order, as otherwise it is impossible to extract the item
        // ident for each version.
```

crates/stackable-versioned-macros/src/codegen/chain.rs
```diff
@@ -14,6 +14,15 @@ impl<K, V> Neighbors<K, V> for BTreeMap<K, V>
 where
     K: Ord + Eq,
 {
+    /// Returns the values of keys which are neighbors of `key`.
+    ///
+    /// Imagine a map which contains the following keys: 1, 3, 5. Calling this
```
NickLarsenNZ, line 19:
```suggestion
    /// Given a map which contains the following keys: 1, 3, 5. Calling this
```

crates/stackable-versioned-macros/src/attrs/variant.rs
```diff
@@ -1,12 +1,22 @@
...
+
+use crate::attrs::common::{ItemAttributes, ItemType};
+
+/// This struct describes all available variant attributes, as well as the
+/// variant name to display better diagnostics.
+///
+/// Data stored in this struct is validated using darling's `and_then` attribute.
+/// During darlings validation, it is not possible to validate that action
+/// versions match up with declared versions on the container. This validation
+/// can be done using the associated [`FieldAttributes::validate_versions`][1]
```
NickLarsenNZ, line 13:
```suggestion
/// can be done using the associated [`VariantAttributes::validate_versions`][1]
```

crates/stackable-versioned-macros/src/attrs/common/item.rs
```diff
@@ -1,13 +1,114 @@
...
+    T: Attributes,
+    I: Spanned,
+{
+    fn validate_versions(
+        &self,
+        container_attrs: &ContainerAttributes,
+        item: &I,
+    ) -> Result<(), darling::Error> {
+        // NOTE (@Techassi): Can we maybe optimize this a little?
+        let mut errors = Error::accumulator();
```
NickLarsenNZ, line 42:
```suggestion
        // NOTE (@Techassi): Can we maybe optimize this a little?

        let mut errors = Error::accumulator();
```

NickLarsenNZ (approved):
LGTM

crates/stackable-versioned-macros/src/codegen/vstruct/field.rs
```diff
@@ -0,0 +1,187 @@
...
+                // of actions. If it is, some action occurred in exactly that
+                // version and thus code is generated for that field based on
+                // the type of action.
+                // If not, the provided version has no action attached to it.
+                // The code generation then depends on the relation to other
+                // versions (with actions).
+
+                let field_type = &self.inner.ty;
+
+                // NOTE (@Techassi): https://rust-lang.github.io/rust-clippy/master/index.html#/expect_fun_call
```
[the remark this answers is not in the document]
Techassi: Yes, that would indeed be very cool.

PR #825, opened 2024-07-10
chore: Add crate release helper scripts

NickLarsenNZ:
# Description

Added script to help with release crates:
- `.scripts/verify_crate_versions.sh` - used to ensure `CHANGELOG.md` versions match what is in `Cargo.toml`. Useful when creating a release PR.
- `.scripts/tag_and_push_release.sh` - used after a release PR has been merged. It will fetch the origin, and tag each _not yet tagged_ release to the selected commit, and optionally push the tags.

> [!TIP]
> Should we add `.scripts/verify_crate_versions.sh` to the pre-commit hooks? This can save a lot of time and human error.

### DIFF
--- a/.github/workflows/pr_pre-commit.yaml
+++ b/.github/workflows/pr_pre-commit.yaml
@@ -1,0 +1,15 @@
+---
+name: pre-commit
+
+on:
+  pull_request:
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+        with:
+          python-version: '3.12'
+      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
--- a/.github/workflows/pr_pre-commit.yaml
+++ b/.github/workflows/pr_pre-commit.yaml
@@ -6,12 +6,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  CARGO_INCREMENTAL: '0'
-  CARGO_PROFILE_DEV_DEBUG: '0'
   RUST_TOOLCHAIN_VERSION: "1.79.0"
-  RUSTFLAGS: "-D warnings"
-  RUSTDOCFLAGS: "-D warnings"
-  RUST_LOG: "info"
 
 jobs:
   pre-commit:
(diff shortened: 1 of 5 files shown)


NickLarsenNZ:
> Do we maybe want to add `shellcheck` to the reviewdog workflow to make sure these scripts are valid?

Oh good point. I ran shellcheck across them first, so never realised it wasn't already a check.

Let me add it in this PR.

Techassi:
I'm also in favour of adding the `.scripts/verify_crate_versions.sh` script to the pre-commit config.

NickLarsenNZ:
I've added shellcheck to both reviewdog and pre-commit (copied from the docker-images repo), and added the verify_crate_versions.sh to the pre-commit config.

I have not added a pre-commit github action, so CI will not pickup mismatched versions (between Cargo.toml and CHANGELOG.md).

Would you like me to add [that workflow](https://github.com/stackabletech/docker-images/blob/main/.github/workflows/pr_pre-commit.yaml) in?

Techassi:
> Would you like me to add [that workflow](https://github.com/stackabletech/docker-images/blob/main/.github/workflows/pr_pre-commit.yaml?rgh-link-date=2024-07-10T13%3A32%3A13Z) in?

Tbh, I would keep it as is for now. I would rather invest more time in better tooling / improving the workflows (in the future).

NickLarsenNZ:
> Tbh, I would keep it as is for now. I would rather invest more time in better tooling / improving the workflows (in the future).

I more meant pre-commit running in Github Actions (not so much this temporary script). Much is duplicated by review dog (same dilemma in docker-images).

Techassi (approved):

Techassi:
> > Tbh, I would keep it as is for now. I would rather invest more time in better tooling / improving the workflows (in the future).
> 
> I more meant pre-commit running in Github Actions (not so much this temporary script). Much is duplicated by review dog (same dilemma in docker-images).

Sure, let's add it. Shouldn't be much work anyway. We should keep in mind that much of that work will eventually be replaced (for example when moving this repo to it's own org).

NickLarsenNZ:
> Sure, let's add it. Shouldn't be much work anyway. We should keep in mind that much of that work will eventually be replaced (for example when moving this repo to it's own org).

Do you mean either reviewdog or pre-commit wouldn't be used anymore? At least the functionality those two provide will be needed in some capacity.

Side-note: I would be in favour of dropping reviewdog for pre-commit, since the latter can work locally and in CI (maybe reviewdog can too though).

Techassi (approved):

Techassi:
> Do you mean either reviewdog or pre-commit wouldn't be used anymore? At least the functionality those two provide will be needed in some capacity.

No, they still will be used, but the workflows around it will probably change.

> Side-note: I would be in favour of dropping reviewdog for pre-commit, since the latter can work locally and in CI (maybe reviewdog can too though).

Yeah, interesting idea. Worth exploring in the future.

PR #838, opened 2024-08-09
Add iter::reverse_if helper

nightkr:
# Description

This is an alternative to

```rust
if cond {
    // must collect here, since foo.rev() is a distinct type from foo
    foo.rev().collect::<Vec<_>>()
} else {
    foo.collect()
}
```

That is both much more convenient to use and avoids the buffering and allocation.

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- `iter::reverse_if` helper ([#838]).
+
+[#838]: https://github.com/stackabletech/operator-rs/pull/838
+
 ## [0.73.0] - 2024-08-09
 
 ### Added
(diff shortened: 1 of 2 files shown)


NickLarsenNZ (approved):
The impl looks good to me. I am curious about two things:
- how this came to be (where will it be used)
- what does _marble_ mean in the tests?

nightkr:
> how this came to be (where will it be used)

https://github.com/stackabletech/hdfs-operator/blob/808f926dc0ad585bac972b0817f1a8548f524725/rust/operator-binary/src/hdfs_controller.rs#L334-L344

> what does marble mean in the tests?

The style of black-box testing iterator/async implementation as "this sequence of inputs should generate this sequence of outputs". I originally got it from Rx (where it's both a [helpful education tool](https://rxmarbles.com/) and a [testing framework](https://rxjs.dev/guide/testing/marble-testing)).

PR #848, opened 2024-08-22
refactor(stackable-versioned): Adjust variant rename validation error span

Techassi:

Variant rename validation errors are now reported on the `from` attribute instead of the variant ident.


NickLarsenNZ (approved):
LGTM

PR #850, opened 2024-08-23
feat(stackable-versioned): Forward container visibility

Techassi:

This PR adds visibility forwarding instead of hardcoding `pub` as the generated module visibility. It now uses the visibility of the definition container.


NickLarsenNZ (approved):
LGTM

PR #853, opened 2024-09-03
fix: Lower Logback checkIncrement value

labrenbe:
# Description
Partially fixes https://github.com/stackabletech/issues/issues/574.

By default Logback checks the size of log files every 60s to determine if log rotation should be triggered. This can be too slow in NiFi with high load resulting in log files becoming larger than the provisioned `emptyDir` volume. Lowering the `checkIncrement` property allows for more timely file size checks.

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


crates/stackable-operator/src/product_logging/framework.rs
```diff
@@ -589,6 +589,7 @@ pub fn create_logback_config(
     </rollingPolicy>
     <triggeringPolicy class="ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy">
       <MaxFileSize>{max_log_file_size_in_mib}MB</MaxFileSize>
+      <checkIncrement>5s</checkIncrement>
```
siegfriedweber, line 592: `5s` sets `checkIncrement` to 5 milliseconds.

The plural `s` is ignored in the pattern of the unit part:
https://github.com/qos-ch/logback/blob/v_1.5.7/logback-core/src/main/java/ch/qos/logback/core/util/Duration.java#L43

Therefore, this value has effectively no unit which defaults to milliseconds:
https://github.com/qos-ch/logback/blob/v_1.5.7/logback-core/src/main/java/ch/qos/logback/core/util/Duration.java#L96

The unit `second` must be used instead:

```suggestion
      <checkIncrement>5 seconds</checkIncrement>
```

siegfriedweber (changes requested):
This change must be added to the changelog.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -39,6 +40,7 @@ All notable changes to this project will be documented in this file.
 [#840]: https://github.com/stackabletech/operator-rs/pull/840
 [#841]: https://github.com/stackabletech/operator-rs/pull/841
 [#843]: https://github.com/stackabletech/operator-rs/pull/843
+[#853]: https://github.com/stackabletech/operator-rs/pull/853
```
siegfriedweber, line 43: This line must be moved to the Unreleased section.

siegfriedweber (changes requested):

crates/stackable-operator/CHANGELOG.md
labrenbe: Thanks for spotting that

siegfriedweber (approved):
LGTM

PR #856, opened 2024-09-05
fix(ci): Install hadolint in pre-commit workflow

Techassi:
Part of https://github.com/stackabletech/issues/issues/637

This fixes two things:

- Install the hadolint binary to be able to run the pre-commit hook
- Set the default language version of `node` to `system`. This fixes issues on NixOS systems.


NickLarsenNZ (approved):
LGTM

PR #859, opened 2024-09-06
feat(stackable-versioned): Add 'use super::*' to generated version module

Techassi:

This PR enables the user to use imported items in container definitions. Previously, the version modules limited the visibility of such items inside said module. Now, using `use super::*` the items can be accessed.

This issue was reported by @sbernauer after a first quick testing round.


NickLarsenNZ (approved):
Like a superstar (get it?)

PR #860, opened 2024-09-06
fix(stackable-versioned): Emit right type for unchanged fields

Techassi:

Previously, the type used in fields marked as NoChange in a particular version was incorrect. It used the type of the definition container. The NoChange action now also tracks the field type and can thus generate the correct type.

Discovered by @sbernauer.


NickLarsenNZ (approved):
LGTM. I assume the "good" trybuild validates this (since the `added` fields was removed).

PR #862, opened 2024-09-11
Check for `ownerReferences` in `ClusterResources`

### LINKED ISSUE #861: `ClusterResources::delete_orphaned_resources` also deletes objects that are not owned directly by it
For example:

- user creates kafka stacklet
- kafka-op creates kafka listener (which is managed by kafka-op)
- list-op creates kafka listener service (which is managed by list-op), but propagates labels from the listener
- kafka-op sees its ownership labels on the listener service and tries to delete it, starting a tug of war with list-op
- list-op sees that the service was deleted, recreates it
- rinse, repeat, ad infinitum

Arguably, this is actually two separate bugs that should both be resolved separately:
- `delete_orphaned_resources` should also do a sanity check against `owner_references` before deleting
- list-op should use its own `ClusterResources`, and overwrite the ownership labels inherited from the listener
  - https://github.com/stackabletech/listener-operator/issues/221

nightkr:
Fixes #861

-------

# Description

Fixes #861.

```[tasklist]
# Author
  - [x] Tested manually
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/src/cluster_resources.rs
+++ b/crates/stackable-operator/src/cluster_resources.rs
@@ -388,6 +388,9 @@ pub struct ClusterResources {
     /// The name of the application
     app_name: String,
 
+    /// The uid of the cluster object
+    cluster_uid: String,
+
     // TODO (Techassi): Add doc comments
     operator_name: String,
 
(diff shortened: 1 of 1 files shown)


maltesander (approved):
LGTM!

PR #679, opened 2023-10-25
refactor!: Simplify common S3 structs and add helper functions

sbernauer:
# Description

For example usage see https://github.com/stackabletech/trino-operator/tree/fix/s3-attributes

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

```[tasklist]
# Author
- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

PR #863, opened 2024-09-11
fix!: Make fields on S3 structs mandatory, add S3 helpers

sbernauer:
# Description

Reviving https://github.com/stackabletech/operator-rs/pull/679 (replacing it)

Example usage is in https://github.com/stackabletech/trino-operator/tree/chore/new-s3-structs

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


NickLarsenNZ (commented):
Partial review (as I see changes are still coming in)

crates/stackable-operator/src/commons/s3/helpers.rs
```diff
@@ -103,7 +103,7 @@ impl ResolvedS3Connection {
 
         if let Some(credentials) = &self.credentials {
             let secret_class = &credentials.secret_class;
-            let volume_name = format!("{secret_class}-s3-credentials-{unique_identifier}");
+            let volume_name = format!("{unique_identifier}-{secret_class}-s3-credentials");
```
NickLarsenNZ, line 106: This particular change is breaking, right?

Once ready for review, I think there should be some text in a comment ready for adding to release notes (and the `release-note/action-required` label.

NickLarsenNZ (commented):
Partial review again :)

crates/stackable-operator/src/commons/s3/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    ParseS3Endpoint { source: url::ParseError },
+
+    #[snafu(display("failed to set S3 endpoint scheme '{scheme}' for endpoint '{endpoint}'"))]
+    SetS3EndpointScheme { endpoint: Url, scheme: String },
+
+    #[snafu(display("failed to add S3 credential volumes and volume mounts"))]
+    AddS3CredentialVolumes { source: SecretClassVolumeError },
+
+    #[snafu(display("failed to add S3 TLS client details volumes and volume mounts"))]
+    AddS3TlsClientDetailsVolumes { source: TlsClientDetailsError },
```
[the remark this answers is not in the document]
sbernauer: Only `RetrieveS3Bucket` is unused, removed in 687141c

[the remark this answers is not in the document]
sbernauer: Added some error details in 72417ff

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -39,6 +49,76 @@ impl Deref for Hostname {
     }
 }
 
+/// A validated host (either a [`Hostname`] or IP address) type.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+#[serde(try_from = "String", into = "String")]
+pub enum Host {
+    IpAddress(IpAddr),
+    Hostname(Hostname),
+}
```
Techassi, line 58: note: Currently, our naming is a little wonky. RFC 1123 states that:

> Whenever a user inputs the identity of an Internet host, it SHOULD
> be possible to enter either (1) a host domain name or (2) an IP
> address in dotted-decimal ("#.#.#.#") form.

So a host name is either a valid domain name, which is also enforced by this sentence in the RFC:

> Host software MUST handle host names of up to 63 characters and
> SHOULD handle host names of up to 255 characters.

or an IP address. Therefore, the enum should be defined as:

```suggestion
pub enum Hostname {
    IpAddress(IpAddr),
    Domain(DomainName),
}
```

---

A nice thing we already do correctly is:

> The host SHOULD check the string syntactically for a dotted-decimal
> number before looking it up in the Domain Name System.

Techassi: note: Also on that note, we might want to take a look at the current validation and ensure all regular expressions are nailed down according to the RFCs.

But that is unrelated to this PR, just a general thought I had.

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -39,6 +49,76 @@ impl Deref for Hostname {
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        value.parse()
+    }
+}
+
+impl From<Host> for String {
+    fn from(value: Host) -> Self {
+        format!("{value}")
+    }
+}
```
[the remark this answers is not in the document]
Techassi: note: I'm personally a fan of writing custom `Deserialize` and `Serialize` implementations, but I wouldn't block this PR on it.

Techassi (changes requested):
I didn't look at the S3 stuff super closely (I can if you want @sbernauer). The TLS code looks fine (as it is copy+pasted from the previous location).

All other comments are minor notes / suggestions.

crates/stackable-operator/src/validation.rs
```diff
@@ -90,6 +90,11 @@ pub enum Error {
 
     #[snafu(display("input is {length} bytes long but must be no more than {max_length}"))]
     TooLong { length: usize, max_length: usize },
+
+    #[snafu(display(
+        "input is not a valid host, which needs to be either a hostname or IP address"
+    ))]
+    NotAHost {},
```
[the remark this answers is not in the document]
Techassi: Also, we should maybe include the input which failed to parse in the error.

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -39,6 +49,76 @@ impl Deref for Hostname {
...
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        if let Ok(ip) = value.parse::<IpAddr>() {
+            return Ok(Host::IpAddress(ip));
+        }
+
+        if let Ok(hostname) = value.parse() {
+            return Ok(Host::Hostname(hostname));
+        };
+
+        Err(validation::Error::NotAHost {})
```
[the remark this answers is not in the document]
Techassi: I would really like to see us using `snafu`'s context selector here, but I guess this is a limitation of bundling all validation errors into a single (foreign) enum.

crates/stackable-operator/src/validation.rs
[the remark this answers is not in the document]
sbernauer: Added the input in 4fb9d50

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -1,44 +1,136 @@
...
 
     fn deref(&self) -> &Self::Target {
         &self.0
     }
 }
 
+#[derive(Debug, Snafu)]
+pub enum HostNameParseError {
+    #[snafu(display(
+        "the given hostname '{hostname}' is not a valid hostname, which needs to be either a domain name or IP address"
```
Techassi, line 56: note: Error fields of type `String` should use the debug output, which automatically adds double quotes around it. See [guide](https://docs.stackable.tech/home/stable/contributor/code-style-guide#_error_messages)

```suggestion
        "the given hostname {hostname:?} is not a valid hostname, which needs to be either a domain name or IP address"
```

crates/stackable-operator/CHANGELOG.md
```diff
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- BREAKING: Add `HostName` type and use it within LDAP and OIDC AuthenticationClass as well as S3Connection ([#863]).
+
+### Changed
+
+- BREAKING: TLS verification struct now reside in the `commons::tls_verification` module, instead of being placed below `commons::authentication::tls` ([#863]).
```
Techassi, line 13: suggestion: Slight rewording.

```suggestion
- BREAKING: The TLS verification struct now resides in the `commons::tls_verification` module, instead of being placed below `commons::authentication::tls` ([#863]).
```

crates/stackable-operator/CHANGELOG.md
```diff
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file.
...
+- BREAKING: Rename the `Hostname` type to `DomainName` to be consistent with RFC 1123 ([#863]).
+
+### Fixed
+
+- BREAKING: The fields `bucketName`, `connection` and `host` on `S3BucketSpec`, `InlinedS3BucketSpec` and `S3ConnectionSpec` are now mandatory. Previously operators probably errored out in case they where missing, so this should not affect users, but make the errors much clearer ([#863]).
```
Techassi, line 18: suggestion: Slight rewording.

```suggestion
- BREAKING: The fields `bucketName`, `connection` and `host` on `S3BucketSpec`, `InlinedS3BucketSpec` and `S3ConnectionSpec` are now mandatory. Previously operators errored out in case these fields where missing ([#863]).
```

crates/stackable-operator/CHANGELOG.md
```diff
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file.
...
+
+[#863]: https://github.com/stackabletech/operator-rs/pull/863
+
+## TODO: Create a release before #863, as it's majorly breaking
```
Techassi, line 22: note: This needs to be removed before merging.

Techassi (changes requested):
LGTM, only minor changes.

crates/stackable-operator/src/commons/s3/helpers.rs
NickLarsenNZ: > This change should be operator-internal and not effect the user.

Sounds good to me.

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -1,44 +1,136 @@
...
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use snafu::Snafu;
 
 use crate::validation;
 
-/// A validated hostname type, for use in CRDs.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+/// A validated domain name type conforming to RFC 1123, e.g. an IPv4, but not an IPv6 address.
```
Techassi, line 9: note: This doc comments seems outdated, rewording needed.

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -1,44 +1,136 @@
...
 
+#[derive(Debug, Snafu)]
+pub enum HostNameParseError {
+    #[snafu(display(
+        "the given hostname {hostname:?} is not a valid hostname, which needs to be either a domain name or IP address"
+    ))]
+    InvalidHostname { hostname: String },
+}
+
+/// A validated hostname (either a [`DomainName`] or IP address) type.
```
Techassi, line 61: note: Link to `IpAddr`.

```suggestion
/// A validated hostname, which is either a [`DomainName`] or [`IpAddr`].
```

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -74,3 +166,48 @@ impl Deref for KerberosRealmName {
...
+mod tests {
+    use super::*;
+
+    use rstest::rstest;
+
+    #[rstest]
+    #[case("foo")]
+    #[case("foo.bar")]
+    // This is also a valid domain name
+    #[case("1.2.3.4")]
```
Techassi, line 180: note: Actually, no. This is not a valid domain name. Domain name labels **must** start with one of the 52 alphabetic characters (A-Z and a-z).

Techassi (changes requested):
LGTM overall, just three small comments.

crates/stackable-operator/src/commons/networking.rs
sbernauer: I'm always torn as this is our CRD description, and we should not leak Rust internals into our API. But yeah, anyway

crates/stackable-operator/src/commons/networking.rs
Techassi: See https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1

crates/stackable-operator/src/commons/networking.rs
sbernauer: Ohm I fail to see the outdated parts, can you please make a suggestion?

Techassi (approved):

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -1,44 +1,136 @@
...
+}
+
+impl JsonSchema for HostName {
+    fn schema_name() -> String {
+        "HostName".to_owned()
+    }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        String::json_schema(gen)
+    }
```
nightkr, line 76: This is effectively dropping the validation regex from the composite type.

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -1,44 +1,136 @@
...
+    InvalidHostname { hostname: String },
+}
+
+/// A validated hostname, which is either a [`DomainName`] or [`IpAddr`].
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+#[serde(try_from = "String", into = "String")]
+pub enum HostName {
+    IpAddress(IpAddr),
+    DomainName(DomainName),
+}
```
nightkr, line 67: Do we actually care about distinguishing these cases?

crates/stackable-operator/src/commons/networking.rs
```diff
@@ -1,44 +1,136 @@
...
+        }
+
+        if let Ok(domain_name) = value.parse() {
+            return Ok(HostName::DomainName(domain_name));
+        };
+
+        InvalidHostnameSnafu {
+            hostname: value.to_owned(),
+        }
+        .fail()
```
nightkr, line 94: This is dropping the context for _why_ validation failed.

crates/stackable-operator/src/commons/networking.rs
sbernauer: You're right. So far we don't have any regex on HostName, but instead let IpAddr::parse do it's thing and try to parse it as domain name in case it#s not an IP.
We should be able to come up with an regex that accepts a DomainName, IPv4 or IPv6 address. But do you think it's worth the effort?

crates/stackable-operator/src/commons/networking.rs
sbernauer: I personally modeled it as String, but @Techassi suggested using an enum instead, which sounded fine to me. No preference from me.

crates/stackable-operator/src/commons/networking.rs
nightkr: That's fair.

crates/stackable-operator/src/commons/networking.rs
Techassi: The RFC calls for these two distinct variants of a hostname.

Why should we remove all the value we get from using proper types by just using a plain string? We _could_ simplify the `DomainName` variant to use `String`, but I would definitely keep the enum.

crates/stackable-operator/src/commons/networking.rs
nightkr: That's one way, but it loses the error chain. Another alternative would be to guess based on a cursory inspection.. something like:

```rs
if value.starts_with(char::is_ascii_digit) || value.contains(':') {
    Ok(HostName::IpAddress(value.parse().context(InvalidIpAddressSnafu { hostname: value })?))
} else {
    Ok(HostName::DomainName(value.parse().context(InvalidDomainNameSnafu { hostname: value })?))
}
```

crates/stackable-operator/src/commons/networking.rs
nightkr: Ultimately the question here is the purpose of the type. Detailed types makes sense for inspecting and manipulating the hostname. The original intent for `Hostname` was the much smaller scope of "make sure this opaque looks close enough to a hostname that it won't fuck up the syntax of the config file it's string-templated into".

crates/stackable-operator/src/commons/networking.rs
Techassi: We could also do something like:

```rust
match value.parse::<IpAddr>() {
    Ok(ip_addr) => return Ok(Hostname::IpAddr(ip_addr)),
    Err(ip_err) => match value.parse::<DomainName>() {
        Ok(dn) => return Ok(Hostname::DomainName(dn)),
        Err(dn_err) => // return both errors
    }
}
```

nightkr: Yes that's effectively what @sbernauer suggested. There problem is that there's no way for an `Error` to have multiple `source()`s, so we'd have to pick one primary and render the other one as unawkwardly as we can.

crates/stackable-operator/src/commons/networking.rs
Techassi: > looks close enough

If we want to achieve this, sure. But then why reference RFC 1123 and only vaguely conform to it. If we want to conform to various RFCs, we should do so as closely as possible.

Another region where we _try_ to conform to an RFC (1035), we also completely miss the mark when validating domain labels. So I'm just trying to avoid the same thing happening here (even tho the domain name is not 100% valid).

crates/stackable-operator/src/commons/networking.rs
Techassi: >  there's no way for an Error to have multiple source()s, so we'd have to pick one primary and render the other one as unawkwardly as we can.

That is indeed very true (and unfortunate). The cursed inspection is also kinda _meh_ tho. I don't know if it helps (I assume not), RFC 1123 states that the host name should first tried to be parsed as an IP address. So at least we do the correct thing RFC-wise.

PR #864, opened 2024-09-11
docs(stackable-versioned): Update usage guide

Techassi:

> [!TIP]
> Can be rendered locally using: 
> ```
> cargo doc -p stackable-versioned --document-private-items --all-features
> ```

This updates the usage guide as well as mention that the macro must be placed before all other macros and attributes.

```[tasklist]
# Reviewer
```


crates/stackable-versioned-macros/src/attrs/common/item.rs
```diff
@@ -317,6 +318,10 @@ fn default_default_fn() -> SpannedValue<Path> {
     )
 }
 
+// TODO (@Techassi): Add validation for when from_name AND from_type or both
+// none => is this action needed in the first place?
```
NickLarsenNZ, line 322: I assume this was supposed to be:

```suggestion
// TODO (@Techassi): Add validation for when from_name AND from_type are both
// none => is this action needed in the first place?
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -8,50 +8,95 @@ mod attrs;
...
-/// This macro enables generating versioned structs.
+/// This macro enables generating versioned structs and enums.
 ///
-/// ## Usage Guide
+/// # Usage Guide
 ///
-/// ### Quickstart
+/// In this guide, code blocks usually come in pairs. The first code block
+/// describes how the macro is used. The second expandable block displays the
+/// generated piece of code for explanation purposes. It should be noted, that
```
NickLarsenNZ, line 17: **praise**: The expandable code blocks are a good idea!

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -8,50 +8,95 @@ mod attrs;
...
+/// the exact code can diverge from what is being depicted in this guide. For
+/// example, `#[automatically_derived]` and `#[allow(deprecated)]` are removed
+/// in most examples to reduce visual clutter.
+///
+/// ## Declaring Versions
+///
+/// It is **important** to note that this macro must be placed before any other
+/// (derive) macros and attributes. This ensures that the macros and attributes
+/// are applied to the generated versioned instances of the struct or enum.
```
NickLarsenNZ, line 26:
```suggestion
/// It is **important** to note that this macro must be placed before any other
/// (derive) macros and attributes. Macros supplied before the versioned macro
/// will be erased, because the original struct or enum (container) is erased,
/// and new containers are generated. This ensures that the macros and
/// attributes are applied to the generated versioned instances of the
/// container.
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -8,50 +8,95 @@ mod attrs;
...
+///
+/// Before any of the fields or variants can be versioned, versions need to be
+/// declared at the container level. Each version currently supports two
+/// parameters: `name` and the `deprecated` flag. The `name` must be a valid
+/// (and supported) format - currently, only Kubernetes API versions are
+/// supported. The macro checks each declared version and reports any error
+/// encountered during parsing.
```
NickLarsenNZ, line 33: As per https://github.com/rust-lang/rust/issues/79710#issuecomment-1868261223, we can make this a warning.

```suggestion
/// (and supported) format.
///
/// <div class="warning">
/// Currently, only Kubernetes API versions are
/// supported. The macro checks each declared version and reports any error
/// encountered during parsing.
/// </div>
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -8,50 +8,95 @@ mod attrs;
...
-/// ### Declaring Versions
+/// <details>
+/// <summary>Generated code</summary>
+///
+/// 1. The `#[automatically_derived]` attribute indicates that the following
+///    piece of code is automatically generated by a macro instead of being
+///    handwritten by a developer. This information is used by cargo and rustc.
+/// 2. For each declared version, a new module containing the struct or enum
+///    (container) is generated. This enables you to reference the container by
+///    versions via `v1alpha1::Foo`.
```
NickLarsenNZ, line 51:
```suggestion
/// 2. For each declared version, a new module containing the container is
///    generated. This enables you to reference the container by versions via
///    `v1alpha1::Foo`.
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -65,101 +110,272 @@ mod consts;
...
+///     bar: usize,
+/// }
 /// ```
 ///
-/// ### Field Actions
+/// ## Item Actions
 ///
-/// This library currently supports three different field actions. Fields can
+/// This library currently supports three different item actions. Items can
 /// be added, renamed and deprecated. The macro ensures that these actions
```
NickLarsenNZ, line 121:
```suggestion
/// This crate currently supports three different item actions. Items can
/// be added, changed, and deprecated. The macro ensures that these actions
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -65,101 +110,272 @@ mod consts;
...
 ///
-/// For fields marked as deprecated, two additional rules apply:
+/// For items marked as deprecated, one additional rule applies:
 ///
-/// - Fields must start with the `deprecated_` prefix.
-/// - The deprecation note cannot be empty.
+/// - Fields must start with the `deprecated_` and variants with the
+///   `Deprecated` prefix. This is enforced because Kubernetes doesn't allow
+///   removing fields in CRDs entirely. Instead, they should be marked as
+///   deprecated. By convention this is done with the `deprecated` prefix.
```
NickLarsenNZ, line 138: **thought (non-blocker)**: This should probably be a property of the `k8s` feature, so that developers not generating CRDs can still opt-in to k8s API versioning.

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -65,101 +110,272 @@ mod consts;
...
 /// ```
+/// </details>
 ///
-/// #### Skip [`From`] generation
+/// ## Auto-generated `From` Implementations
 ///
-/// Generation of these [`From`] implementations can be skipped at the container
-/// and version level. This enables customization of the implementations if the
-/// default implementation is not sufficient.
+/// To enable smooth version upgrades of the same container, the macro
```
NickLarsenNZ, line 365: Consider rewording "upgrades of the same container".

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -65,101 +110,272 @@ mod consts;
...
+/// automatically generates `From` implementations. On a high level, code
+/// generated for two versions _a_ and _b_, with _a < b_ looks like this:
+/// `impl From<a> for b`. As you can see, only upgrading is currently supported.
+/// Downgrading, so going from a higher version to a lower one, is not supported
+/// at the moment.
```
NickLarsenNZ, line 370:
```suggestion
/// Downgrading from a higher version to a lower one is not currently supported.
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -65,101 +110,272 @@ mod consts;
...
+///
+/// This automatic generation can be skipped to for example enable a custom
+/// implementation for more complex conversions.
```
NickLarsenNZ, line 373: Consider revising.

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -65,101 +110,272 @@ mod consts;
...
+///
+/// ### Skipping at the Container Level
+///
+/// Disabling this feature on the container level results in no `From`
+/// implementation for all versions.
```
NickLarsenNZ, line 378:
```suggestion
/// Disabling this behavior at the container level results in no `From`
/// implementation for all versions.
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -172,39 +388,72 @@ mod consts;
...
-/// #### Customize Default Function for Added Fields
+/// ### Skipping at the Version Level
 ///
-/// It is possible to customize the default function used in the generated
-/// [`From`] implementation for populating added fields. By default,
-/// [`Default::default()`] is used.
+/// Disabling this feature at the version level results in no `From`
+/// implementation for that particular version. This can be read as "skip
+/// generation for converting _this_ version to the next one". In the example
+/// below no conversion between version `v1beta1` and `v1` is generated.
```
NickLarsenNZ, line 403:
```suggestion
/// Disabling this behavior at the version level results in no `From`
/// implementation for that particular version. This can be read as "skip
/// generation for converting _this_ version to the next one". In the example
/// below no conversion between version `v1beta1` and `v1` is generated.
```

crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -172,39 +388,72 @@ mod consts;
...
-///     42
-/// }
 /// ```
+///
+/// ## Kubernetes-specific Features
+///
+/// This macro also offers support for Kubernetes-specific versioning,
+/// especially for CustomResourceDefinitions, short CRDs. These features are
+/// completely opt-in. You need to enable the `k8s` feature (which enables
+/// optional dependencies) and use the `k8s()` parameter in the macro.
```
NickLarsenNZ, line 427: Note to self, reword "short CRDS" bit.

Techassi: Done in 29aed25

crates/stackable-versioned-macros/src/lib.rs
Techassi: Done in 29aed25

crates/stackable-versioned-macros/src/lib.rs
Techassi: Done in 29aed25

NickLarsenNZ (approved):
LGTM

PR #871, opened 2024-09-20
fix!: Avoid clashing volumes and mounts

sbernauer:
# Description

Discovered in https://github.com/stackabletech/trino-operator/pull/646

```
2024-09-19T14:50:51.706423Z ERROR stackable_operator::logging::controller: Failed to reconcile object controller.name="trinocluster.trino.stackable.tech" error=reconciler for object TrinoCluster.v1alpha1.trino.stackable.tech/trino.kuttl-test-golden-toucan failed error.sources=[failed to apply StatefulSet for role group worker/default of TrinoCluster.v1alpha1.trino.stackable.te
ch/trino.kuttl-test-golden-toucan, failed to apply patch, unable to patch resource "trino-worker-default", ApiError: failed to create typed patch object (kuttl-test-golden-toucan/trino-worker-default; apps/v1, Kind=StatefulSet): errors:                                                                                                                                               
  .spec.template.spec.containers[name="trino"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/s3-credentials-class"]                                                                                                                                                                                                                                               
  .spec.template.spec.containers[name="trino"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/minio-tls-certificates"]                                                                                                                                                                                                                                             
  .spec.template.spec.initContainers[name="prepare"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/s3-credentials-class"]                                                                                                                                                                                                                                         
  .spec.template.spec.initContainers[name="prepare"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/minio-tls-certificates"]                                                                                                                                                                                                                                       
  .spec.template.spec.volumes: duplicate entries for key [name="s3-credentials-class-s3-credentials"]                                                                                                                                                                                                                                                                                      
  .spec.template.spec.volumes: duplicate entries for key [name="minio-tls-certificates-ca-cert"]:  (ErrorResponse { status: "Failure", message: "failed to create typed patch object (kuttl-test-golden-toucan/trino-worker-default; apps/v1, Kind=StatefulSet): errors:\n  .spec.template.spec.containers[name=\"trino\"].volumeMounts: duplicate entries for key [mountPath=\"/stackable/
secrets/s3-credentials-class\"]\n  .spec.template.spec.containers[name=\"trino\"].volumeMounts: duplicate entries for key [mountPath=\"/stackable/secrets/minio-tls-certificates\"]\n  .spec.template.spec.initContainers[name=\"prepare\"].volumeMounts: duplicate entries for key [mountPath=\"/stackable/secrets/s3-credentials-class\"]\n  .spec.template.spec.initContainers[name=\"pr
epare\"].volumeMounts: duplicate entries for key [mountPath=\"/stackable/secrets/minio-tls-certificates\"]\n  .spec.template.spec.volumes: duplicate entries for key [name=\"s3-credentials-class-s3-credentials\"]\n  .spec.template.spec.volumes: duplicate entries for key [name=\"minio-tls-certificates-ca-cert\"]", reason: "", code: 500 }), failed to create typed patch object (ku
ttl-test-golden-toucan/trino-worker-default; apps/v1, Kind=StatefulSet): errors:                                                                                                                                                                                                                                                                                                           
  .spec.template.spec.containers[name="trino"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/s3-credentials-class"]                                                                                                                                                                                                                                               
  .spec.template.spec.containers[name="trino"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/minio-tls-certificates"]                                                                                                                                                                                                                                             
  .spec.template.spec.initContainers[name="prepare"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/s3-credentials-class"]                                                                                                                                                                                                                                         
  .spec.template.spec.initContainers[name="prepare"].volumeMounts: duplicate entries for key [mountPath="/stackable/secrets/minio-tls-certificates"]                                                                                                                                                                                                                                       
  .spec.template.spec.volumes: duplicate entries for key [name="s3-credentials-class-s3-credentials"]                                                                                                                                                                                                                                                                                      
  .spec.template.spec.volumes: duplicate entries for key [name="minio-tls-certificates-ca-cert"]: ]
```

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -35,6 +35,8 @@ pub enum Error {
             the new mount's subPath is {new_sub_path:?}"
     ))]
     ClashingMountPath {
+        // The VolumeMount structs where not added to avoid to many information for the users. Instead we pick the most
+        // relevant information only.
         mount_path: String,
         existing_volume_name: String,
         existing_sub_path: Option<String>,
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -28,8 +28,8 @@ pub enum Error {
             The existing volumeMount is {existing_volume_mount:?}, the new one is {new_volume_mount:?}"
     ))]
     ClashingVolumeMountMountPath {
-        existing_volume_mount: VolumeMount,
-        new_volume_mount: VolumeMount,
+        existing_volume_mount: Box<VolumeMount>,
+        new_volume_mount: Box<VolumeMount>,
     },
 }
 
@@ -539,21 +539,17 @@ mod tests {
             "lengthexceededlengthexceededlengthexceededlengthexceededlengthex";
         assert_eq!(long_container_name.len(), 64); // 63 characters is the limit for container names
         let result = ContainerBuilder::new(long_container_name);
-        match result
-            .err()
-            .expect("Container name exceeding 63 characters should cause an error")
-        {
-            Error::InvalidContainerName {
+        if let Error::InvalidContainerName {
                 container_name,
                 source,
-            } => {
-                assert_eq!(container_name, long_container_name);
-                assert_eq!(
-                    source.to_string(),
-                    "input is 64 bytes long but must be no more than 63"
-                )
-            }
-            _ => {}
+            } = result
+            .err()
+            .expect("Container name exceeding 63 characters should cause an error") {
+            assert_eq!(container_name, long_container_name);
+            assert_eq!(
+                source.to_string(),
+                "input is 64 bytes long but must be no more than 63"
+            )
         }
         // One characters shorter name is valid
         let max_len_container_name: String = long_container_name.chars().skip(1).collect();
@@ -617,17 +613,13 @@ mod tests {
         result: Result<ContainerBuilder, Error>,
         expected_err_contains: &str,
     ) {
-        match result
-            .err()
-            .expect("Container name exceeding 63 characters should cause an error")
-        {
-            Error::InvalidContainerName {
+        if let Error::InvalidContainerName {
                 container_name: _,
                 source,
-            } => {
-                assert!(dbg!(source.to_string()).contains(dbg!(expected_err_contains)));
-            }
-            _ => {}
+            } = result
+            .err()
+            .expect("Container name exceeding 63 characters should cause an error") {
+            assert!(dbg!(source.to_string()).contains(dbg!(expected_err_contains)));
         }
     }
 }
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -218,7 +218,7 @@ impl ContainerBuilder {
     pub fn add_volume_mount_struct(&mut self, volume_mount: VolumeMount) -> Result<&mut Self> {
         if let Some(existing_volume_mount) = self.volume_mounts.get(&volume_mount.mount_path) {
             ensure!(
-                existing_volume_mount.eq(&volume_mount),
+                existing_volume_mount == &volume_mount,
                 ClashingVolumeMountMountPathSnafu {
                     existing_volume_mount: existing_volume_mount.clone(),
                     new_volume_mount: volume_mount,
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -300,6 +300,12 @@ impl ContainerBuilder {
     }
 
     pub fn build(&self) -> Container {
+        let volume_mounts = if self.volume_mounts.is_empty() {
+            None
+        } else {
+            Some(self.volume_mounts.values().cloned().collect())
+        };
+
         Container {
             args: self.args.clone(),
             command: self.command.clone(),
@@ -309,11 +315,7 @@ impl ContainerBuilder {
             resources: self.resources.clone(),
             name: self.name.clone(),
             ports: self.container_ports.clone(),
-            volume_mounts: if self.volume_mounts.is_empty() {
-                None
-            } else {
-                Some(self.volume_mounts.clone().into_values().collect())
-            },
+            volume_mounts,
             readiness_probe: self.readiness_probe.clone(),
             liveness_probe: self.liveness_probe.clone(),
             startup_probe: self.startup_probe.clone(),
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -6,7 +6,7 @@ use k8s_openapi::api::core::v1::{
     LifecycleHandler, ObjectFieldSelector, Probe, ResourceRequirements, SecretKeySelector,
     SecurityContext, VolumeMount,
 };
-use snafu::{ResultExt as _, Snafu};
+use snafu::{ensure, ResultExt as _, Snafu};
 
 use crate::{
     commons::product_image_selection::ResolvedProductImage,
@@ -21,6 +21,15 @@ pub enum Error {
     InvalidContainerName {
         source: validation::Errors,
         container_name: String,
+    },
+
+    #[snafu(display(
+        "The volumeMount is clashing with an already existing volumeMount with the same mountPath but a different content. \
+            The existing volumeMount is {existing_volume_mount:?}, the new one is {new_volume_mount:?}"
+    ))]
+    ClashingVolumeMountMountPath {
+        existing_volume_mount: VolumeMount,
+        new_volume_mount: VolumeMount,
     },
 }
 
@@ -206,22 +215,21 @@ impl ContainerBuilder {
     ///
     /// We could have made this function fallible, but decided not to do so for now, as it would be a bigger breaking
     /// change.
-    pub fn add_volume_mount_struct(&mut self, volume_mount: VolumeMount) -> &mut Self {
+    pub fn add_volume_mount_struct(&mut self, volume_mount: VolumeMount) -> Result<&mut Self> {
         if let Some(existing_volume_mount) = self.volume_mounts.get(&volume_mount.mount_path) {
-            if !existing_volume_mount.eq(&volume_mount) {
-                tracing::error!(
-                    ?existing_volume_mount,
-                    new_volume_mount = ?volume_mount,
-                    "The operator tried to add a volumeMount to Container, which was already added with a different content! \
-                        The new volumeMount will be ignored."
-                );
-            }
+            ensure!(
+                existing_volume_mount.eq(&volume_mount),
+                ClashingVolumeMountMountPathSnafu {
+                    existing_volume_mount: existing_volume_mount.clone(),
+                    new_volume_mount: volume_mount,
+                }
+            );
         } else {
             self.volume_mounts
                 .insert(volume_mount.mount_path.clone(), volume_mount);
         }
 
-        self
+        Ok(self)
     }
 
     /// See [`Self::add_volume_mount_struct`] for details
@@ -229,7 +237,7 @@ impl ContainerBuilder {
         &mut self,
         name: impl Into<String>,
         path: impl Into<String>,
-    ) -> &mut Self {
+    ) -> Result<&mut Self> {
         self.add_volume_mount_struct(VolumeMount {
             name: name.into(),
             mount_path: path.into(),
@@ -241,11 +249,12 @@ impl ContainerBuilder {
     pub fn add_volume_mounts(
         &mut self,
         volume_mounts: impl IntoIterator<Item = VolumeMount>,
-    ) -> &mut Self {
+    ) -> Result<&mut Self> {
         for volume_mount in volume_mounts {
-            self.add_volume_mount_struct(volume_mount);
+            self.add_volume_mount_struct(volume_mount)?;
         }
-        self
+
+        Ok(self)
     }
 
     pub fn readiness_probe(&mut self, probe: Probe) -> &mut Self {
@@ -427,6 +436,7 @@ mod tests {
             .add_env_var_from_config_map("envFromConfigMap", "my-configmap", "my-key")
             .add_env_var_from_secret("envFromSecret", "my-secret", "my-key")
             .add_volume_mount("configmap", "/mount")
+            .unwrap()
             .add_container_port(container_port_name, container_port)
             .resources(resources.clone())
             .add_container_ports(vec![ContainerPortBuilder::new(container_port_1)
@@ -543,6 +553,7 @@ mod tests {
                     "input is 64 bytes long but must be no more than 63"
                 )
             }
+            _ => {}
         }
         // One characters shorter name is valid
         let max_len_container_name: String = long_container_name.chars().skip(1).collect();
@@ -616,6 +627,7 @@ mod tests {
             } => {
                 assert!(dbg!(source.to_string()).contains(dbg!(expected_err_contains)));
             }
+            _ => {}
         }
     }
 }
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -28,11 +28,22 @@ pub enum Error {
 
     #[snafu(display(
         "The volumeMount is clashing with an already existing volumeMount with the same mountPath but a different content. \
-            The existing volumeMount is {existing_volume_mount:?}, the new one is {new_volume_mount:?}"
+            The shared mountPath is {mount_path:?}, \
+            the existing mount's volume name is {existing_volume_name:?}, \
+            the existing mount's subPath is {existing_sub_path:?}, \
+            the existing mount's SubPathExpr is {existing_sub_path_expr:?}, \
+            the new mount's volume name is {new_volume_name:?}, \
+            the new mount's subPath is {new_sub_path:?}, \
+            the new mount's SubPathExpr is {new_sub_path_expr:?}"
     ))]
-    ClashingVolumeMountMountPath {
-        existing_volume_mount: Box<VolumeMount>,
-        new_volume_mount: Box<VolumeMount>,
+    ClashingMountPath {
+        mount_path: String,
+        existing_volume_name: String,
+        existing_sub_path: Option<String>,
+        existing_sub_path_expr: Option<String>,
+        new_volume_name: String,
+        new_sub_path: Option<String>,
+        new_sub_path_expr: Option<String>,
     },
 }
 
@@ -219,9 +230,14 @@ impl ContainerBuilder {
         if let Some(existing_volume_mount) = self.volume_mounts.get(&volume_mount.mount_path) {
             ensure!(
                 existing_volume_mount == &volume_mount,
-                ClashingVolumeMountMountPathSnafu {
-                    existing_volume_mount: existing_volume_mount.clone(),
-                    new_volume_mount: volume_mount,
+                ClashingMountPathSnafu {
+                    mount_path: volume_mount.mount_path,
+                    existing_volume_name: existing_volume_mount.name.clone(),
+                    existing_sub_path: existing_volume_mount.sub_path.clone(),
+                    existing_sub_path_expr: existing_volume_mount.sub_path_expr.clone(),
+                    new_volume_name: volume_mount.name,
+                    new_sub_path: volume_mount.sub_path,
+                    new_sub_path_expr: volume_mount.sub_path_expr,
                 }
             );
         } else {
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -38,7 +38,7 @@ pub enum Error {
 /// This will automatically create the necessary volumes and mounts for each `ConfigMap` which is added.
 ///
 /// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
-/// We are also choosing it over an [`std::collections::BTreeMap`], as it's easier to debug for users, as logically
+/// We are also choosing it over a [`BTreeMap`], because it is easier to debug for users, as logically
 /// grouped volumeMounts (e.g. all volumeMounts related to S3) are near each other in the list instead of "just" being
 /// sorted alphabetically.
 #[derive(Clone, Default)]
@@ -212,9 +212,6 @@ impl ContainerBuilder {
     /// Historically this function unconditionally added volumeMounts, which resulted in invalid
     /// [`k8s_openapi::api::core::v1::PodSpec`]s, as volumeMounts where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`]
     /// two times, resulting in e.g. the s3 credentials being mounted twice as the same volumeMount.
-    ///
-    /// We could have made this function fallible, but decided not to do so for now, as it would be a bigger breaking
-    /// change.
     pub fn add_volume_mount_struct(&mut self, volume_mount: VolumeMount) -> Result<&mut Self> {
         if let Some(existing_volume_mount) = self.volume_mounts.get(&volume_mount.mount_path) {
             ensure!(
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -28,20 +28,15 @@ pub enum Error {
 
     #[snafu(display(
         "The volumeMount is clashing with an already existing volumeMount with the same mountPath but a different content. \
-            The shared mountPath is {mount_path:?}, \
-            the existing mount's volume name is {existing_volume_name:?}, \
-            the existing mount's subPath is {existing_sub_path:?}, \
-            the new mount's volume name is {new_volume_name:?}, \
-            the new mount's subPath is {new_sub_path:?}"
+            The clashing mountPath is {clashing_mount_path:?}, \
+            the existing mount's volume name is {existing_volume_name:?} \
+            and the new mount's volume name is {new_volume_name:?}. \
+            Please have a look at the log/traces for details"
     ))]
     ClashingMountPath {
-        // The VolumeMount structs where not added to avoid to many information for the users. Instead we pick the most
-        // relevant information only.
-        mount_path: String,
+        clashing_mount_path: String,
         existing_volume_name: String,
-        existing_sub_path: Option<String>,
         new_volume_name: String,
-        new_sub_path: Option<String>,
     },
 }
 
@@ -226,14 +221,21 @@ impl ContainerBuilder {
     /// two times, resulting in e.g. the s3 credentials being mounted twice as the same volumeMount.
     fn add_volume_mount_impl(&mut self, volume_mount: VolumeMount) -> Result<&mut Self> {
         if let Some(existing_volume_mount) = self.volume_mounts.get(&volume_mount.mount_path) {
+            // We don't want to include the details in the error message, but instead trace them
+            if existing_volume_mount != &volume_mount {
+                tracing::error!(
+                    clashing_mount_path = &volume_mount.mount_path,
+                    ?existing_volume_mount,
+                    new_volume_mount = ?volume_mount,
+                    "The volumeMount is clashing with an already existing volumeMount with the same mountPath but a different content"
+                );
+            }
             ensure!(
                 existing_volume_mount == &volume_mount,
                 ClashingMountPathSnafu {
-                    mount_path: volume_mount.mount_path,
+                    clashing_mount_path: volume_mount.mount_path,
                     existing_volume_name: existing_volume_mount.name.clone(),
-                    existing_sub_path: existing_volume_mount.sub_path.clone(),
                     new_volume_name: volume_mount.name,
-                    new_sub_path: volume_mount.sub_path,
                 }
             );
         } else {
--- a/crates/stackable-operator/src/builder/pod/container.rs
+++ b/crates/stackable-operator/src/builder/pod/container.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 
+use indexmap::IndexMap;
 use k8s_openapi::api::core::v1::{
     ConfigMapKeySelector, Container, ContainerPort, EnvVar, EnvVarSource, Lifecycle,
     LifecycleHandler, ObjectFieldSelector, Probe, ResourceRequirements, SecretKeySelector,
@@ -26,6 +27,11 @@ pub enum Error {
 /// A builder to build [`Container`] objects.
 ///
 /// This will automatically create the necessary volumes and mounts for each `ConfigMap` which is added.
+///
+/// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
+/// We are also choosing it over an [`std::collections::BTreeMap`], as it's easier to debug for users, as logically
+/// grouped volumeMounts (e.g. all volumeMounts related to S3) are near each other in the list instead of "just" being
+/// sorted alphabetically.
 #[derive(Clone, Default)]
 pub struct ContainerBuilder {
     args: Option<Vec<String>>,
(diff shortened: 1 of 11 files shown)


crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -188,28 +196,55 @@ impl ContainerBuilder {
...
 
+    /// This function only adds the [`VolumeMount`] in case there is no volumeMount with the same mountPath already.
+    /// In case there already was a volumeMount with the same path already, an [`tracing::error`] is raised in case the
+    /// contents of the volumeMounts differ.
+    ///
+    /// Historically this function unconditionally added volumeMounts, which resulted in invalid
+    /// [`k8s_openapi::api::core::v1::PodSpec`]s, as volumeMounts where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`]
+    /// two times, resulting in e.g. the s3 credentials being mounted twice as the same volumeMount.
+    ///
+    /// We could have made this function fallible, but decided not to do so for now, as it would be a bigger breaking
```
NickLarsenNZ, line 207: Can we make an issue out of this and perhaps fix it in the next release?

NickLarsenNZ (approved):
LGTM. I like the doc-comments

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -21,11 +22,25 @@ pub enum Error {
...
 }
 
 /// A builder to build [`Container`] objects.
 ///
 /// This will automatically create the necessary volumes and mounts for each `ConfigMap` which is added.
+///
+/// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
+/// We are also choosing it over an [`std::collections::BTreeMap`], as it's easier to debug for users, as logically
+/// grouped volumeMounts (e.g. all volumeMounts related to S3) are near each other in the list instead of "just" being
+/// sorted alphabetically.
```
NickLarsenNZ, line 43:
```suggestion
/// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
/// We are also choosing it over a [`BTreeMap`], because it is easier to debug for users, as logically
/// grouped volumeMounts (e.g. all volumeMounts related to S3) are near each other in the list instead of "just" being
/// sorted alphabetically.
```

I removed the qualifier for `BTreeMap` to make it consistent with the comment on `pod::PodBuilder`.
If that doesn't work in the docs, the both `BTreeMap` and `IndexMap` should have the qualifiers (on both comments).

crates/stackable-operator/src/builder/pod/container.rs
NickLarsenNZ: This comment is outdated

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -388,6 +436,7 @@ mod tests {
             .add_env_var_from_config_map("envFromConfigMap", "my-configmap", "my-key")
             .add_env_var_from_secret("envFromSecret", "my-secret", "my-key")
             .add_volume_mount("configmap", "/mount")
+            .unwrap()
```
NickLarsenNZ, line 439: We should handle the error here.

If there is good reason not to (eg: proven not to return `Result::Err`), then we should use `expect` with a message:

```suggestion
            .expect("add volume mount")
```

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -490,20 +539,18 @@ mod tests {
...
-                assert_eq!(
-                    source.to_string(),
-                    "input is 64 bytes long but must be no more than 63"
-                )
-            }
+            assert_eq!(container_name, long_container_name);
+            assert_eq!(
+                source.to_string(),
+                "input is 64 bytes long but must be no more than 63"
+            )
```
NickLarsenNZ, line 553: thought (non-blocker): I'm not sure how I feel about asserting error message strings.

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -21,11 +22,25 @@ pub enum Error {
...
+    },
 }
 
 /// A builder to build [`Container`] objects.
 ///
 /// This will automatically create the necessary volumes and mounts for each `ConfigMap` which is added.
+///
+/// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
+/// We are also choosing it over an [`std::collections::BTreeMap`], as it's easier to debug for users, as logically
+/// grouped volumeMounts (e.g. all volumeMounts related to S3) are near each other in the list instead of "just" being
```
razvan, line 42: Didn't a plain Vec fulfill all these criteria already ?

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -50,9 +51,21 @@ pub enum Error {
 
     #[snafu(display("object is missing key {key:?}"))]
     MissingObjectKey { key: &'static str },
+
+    #[snafu(display("The volume {volume_name:?} is clashing with an already existing volume with the same name but a different content. \
+        The existing volume is {existing_volume:?}, the new one is {new_volume:?}"))]
```
NickLarsenNZ, line 56:
```suggestion
    #[snafu(display("The volume {volume_name:?} clashes with an existing volume of the same name but with different content. \
        The existing volume is {existing_volume:?}, the new one is {new_volume:?}"))]
```

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -254,28 +269,54 @@ impl PodBuilder {
...
+    /// This function only adds the [`Volume`] in case there is no volume with the same name already.
+    /// In case there already was a volume with the same name already, an [`tracing::error`] is raised in case the
+    /// contents of the volumes differ.
+    ///
+    /// Historically this function unconditionally added volumes, which resulted in invalid [`PodSpec`]s, as volumes
+    /// where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`] two times,
+    /// resulting in e.g. the s3 credentials being mounted twice as the sake volume.
+    ///
+    /// We could have made this function fallible, but decided not to do so for now, as it would be a bigger breaking
+    /// change.
```
NickLarsenNZ, line 295: The function is now fallible
```suggestion
```

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -188,29 +205,56 @@ impl ContainerBuilder {
...
+    /// This function only adds the [`VolumeMount`] in case there is no volumeMount with the same mountPath already.
+    /// In case there already was a volumeMount with the same path already, an [`tracing::error`] is raised in case the
+    /// contents of the volumeMounts differ.
+    ///
+    /// Historically this function unconditionally added volumeMounts, which resulted in invalid
+    /// [`k8s_openapi::api::core::v1::PodSpec`]s, as volumeMounts where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`]
+    /// two times, resulting in e.g. the s3 credentials being mounted twice as the same volumeMount.
+    ///
+    /// We could have made this function fallible, but decided not to do so for now, as it would be a bigger breaking
+    /// change.
```
NickLarsenNZ, line 217: The function is now fallible.

```suggestion
```

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -50,9 +51,21 @@ pub enum Error {
...
+        existing_volume: Box<Volume>,
+        new_volume: Box<Volume>,
+    },
 }
 
 /// A builder to build [`Pod`] or [`PodTemplateSpec`] objects.
+///
+/// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
+/// We are also choosing it over an [`BTreeMap`], as it's easier to debug for users, as logically grouped volumes
+/// (e.g. all volumes related to S3) are near each other in the list instead of "just" being sorted alphabetically.
```
NickLarsenNZ, line 68:
```suggestion
/// This struct is often times using an [`IndexMap`] to have consistent ordering (so we don't produce reconcile loops).
/// We are also choosing it over a [`BTreeMap`], because it is easier to debug for users, as logically grouped volumes
/// (e.g. all volumes related to S3) are near each other in the list instead of "just" being sorted alphabetically.
```

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -673,6 +720,7 @@ mod tests {
                     .with_config_map("configmap")
                     .build(),
             )
+            .unwrap()
```
NickLarsenNZ, line 723: We should handle the error, or use `expect("...")`

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Yes, but the problem with a Vec is that it contain duplicates (multiple volumes with the same name), which k8s will reject

NickLarsenNZ (dismissed):
Just a few changes needed.

crates/stackable-operator/src/builder/pod/container.rs
razvan: I understand that this is what this PR is fixing.
The question was, what problem does IndexMap solve that the plain Vec and a filter couldn't ?

sbernauer: Ah ok, now I understand.

For me it feels "right" to use a data-structure modeling your data. This way it's 100% guaranteed that there are no duplicates. If we leave it as Vec, chances are, in the future someone might just `.push()` something onto it without knowing about our current problems, thus introducing the bug again. This can not happen with a Map (regardless if BTreeMap or IndexMap). I e.g. ran into this as the list comming in from `add_volumes` itself can already contain duplicates, so we can not simply `extend` that on to the Vec. 

Additionally it is more performant, as the Vec would have an o(n) insert cost, while the IndexMap has o(1) - but I would consider this less important

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Yeah, I tried to avoid importing it behind #[cfg(docs)], but let's keep the rustdocs clear. I will fix any errors popping up

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Thx!

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: It's a test ;) Committed suggestion

crates/stackable-operator/src/builder/pod/mod.rs
sbernauer: Pls heave a look at 53ef278de66ca07c9680810230235ee7a8b260a9

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -188,29 +208,53 @@ impl ContainerBuilder {
         self
     }
 
+    /// This function only adds the [`VolumeMount`] in case there is no volumeMount with the same mountPath already.
+    /// In case there already was a volumeMount with the same path already, an [`tracing::error`] is raised in case the
+    /// contents of the volumeMounts differ.
+    ///
+    /// Historically this function unconditionally added volumeMounts, which resulted in invalid
+    /// [`k8s_openapi::api::core::v1::PodSpec`]s, as volumeMounts where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`]
+    /// two times, resulting in e.g. the s3 credentials being mounted twice as the same volumeMount.
```
Techassi, line 217: suggestion: Rewording

```suggestion
    /// Adds a new [`VolumeMount`] to the container while ensuring that no
    /// duplicate exists.
    ///
    /// This function ensures that no two [`VolumeMount`]s with the same mount
    /// path can be added. An appropriate error is returned when such a duplicate
    /// mount path is encountered.
    ///
    /// ### Note
    ///
    /// Previously, this function unconditionally added [`VolumeMount`]s, which
    /// resulted in invalid [`PodSpec`][1]s.
    ///
    /// [1]: k8s_openapi::api::core::v1::PodSpec
```

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -188,29 +208,53 @@ impl ContainerBuilder {
...
+    pub fn add_volume_mount_struct(&mut self, volume_mount: VolumeMount) -> Result<&mut Self> {
+        if let Some(existing_volume_mount) = self.volume_mounts.get(&volume_mount.mount_path) {
+            ensure!(
+                existing_volume_mount.eq(&volume_mount),
```
Techassi, line 221: question: Any particular reason why we don't use `existing_volume_mount == volume_mount` here?

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -188,29 +208,53 @@ impl ContainerBuilder {
...
     }
 
+    /// This function only adds the [`VolumeMount`] in case there is no volumeMount with the same mountPath already.
+    /// In case there already was a volumeMount with the same path already, an [`tracing::error`] is raised in case the
+    /// contents of the volumeMounts differ.
+    ///
+    /// Historically this function unconditionally added volumeMounts, which resulted in invalid
+    /// [`k8s_openapi::api::core::v1::PodSpec`]s, as volumeMounts where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`]
+    /// two times, resulting in e.g. the s3 credentials being mounted twice as the same volumeMount.
+    pub fn add_volume_mount_struct(&mut self, volume_mount: VolumeMount) -> Result<&mut Self> {
```
Techassi, line 218: question: Does this function need to be public? We already have two functions which should cover all uses cases. I would like to try to keep the API interface as little as possible.

Another option is to deprecate the other two functions and always require a `VolumeMount` to be passed. That the removes a bunch of easy-of-use.

suggestion: If made private, let's rename the function to `add_volume_mount_impl`.

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -265,7 +309,11 @@ impl ContainerBuilder {
             resources: self.resources.clone(),
             name: self.name.clone(),
             ports: self.container_ports.clone(),
-            volume_mounts: self.volume_mounts.clone(),
+            volume_mounts: if self.volume_mounts.is_empty() {
+                None
+            } else {
+                Some(self.volume_mounts.clone().into_values().collect())
+            },
```
Techassi, line 316: note: Inlining this looks kinda convoluted. Please move this out of the struct creation. Also, is there a way to do this with combinators?

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -50,9 +51,21 @@ pub enum Error {
...
     #[snafu(display("object is missing key {key:?}"))]
     MissingObjectKey { key: &'static str },
+
+    #[snafu(display("The volume {volume_name:?} clashes with an existing volume of the same name but with different content. \
+        The existing volume is {existing_volume:?}, the new one is {new_volume:?}"))]
+    ClashingVolumeName {
+        volume_name: String,
+        existing_volume: Box<Volume>,
+        new_volume: Box<Volume>,
+    },
```
Techassi, line 61: note: Same thought from above applies here as well. I think displaying the volume name should be enough.

```suggestion
    #[snafu(display("tried to add duplicate volume {name:?}"))]
    DuplicateVolume {
        name: String,
    },
```

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -254,28 +269,51 @@ impl PodBuilder {
...
-    pub fn add_volumes(&mut self, volumes: Vec<Volume>) -> &mut Self {
-        self.volumes.get_or_insert_with(Vec::new).extend(volumes);
-        self
+    /// This function only adds the [`Volume`] in case there is no volume with the same name already.
+    /// In case there already was a volume with the same name already, an [`tracing::error`] is raised in case the
+    /// contents of the volumes differ.
+    ///
+    /// Historically this function unconditionally added volumes, which resulted in invalid [`PodSpec`]s, as volumes
+    /// where added multiple times - think of Trino using the same [`crate::commons::s3::S3Connection`] two times,
+    /// resulting in e.g. the s3 credentials being mounted twice as the sake volume.
```
Techassi, line 292: suggestion: Rewording

```suggestion
    /// Adds a new [`Volume`] to the pod while ensuring that no duplicate exists.
    ///
    /// This function ensures that no two [`Volume`]s with the same mname can be
    /// added. An appropriate error is returned when such a duplicate name is
    /// encountered.
    ///
    /// ### Note
    ///
    /// Previously, this function unconditionally added [`Volume`]s, which
    /// resulted in invalid [`PodSpec`]s.
```

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -254,28 +269,51 @@ impl PodBuilder {
...
+    pub fn add_volume(&mut self, volume: Volume) -> Result<&mut Self> {
+        if let Some(existing_volume) = self.volumes.get(&volume.name) {
+            ensure!(
+                existing_volume.eq(&volume),
```
Techassi, line 296: question: Same question from above.

crates/stackable-operator/src/builder/pod/mod.rs
```diff
@@ -533,7 +573,11 @@ impl PodBuilder {
             }),
             security_context: self.security_context.clone(),
             tolerations: self.tolerations.clone(),
-            volumes: self.volumes.clone(),
+            volumes: if self.volumes.is_empty() {
+                None
+            } else {
+                Some(self.volumes.clone().into_values().collect())
+            },
```
Techassi, line 580: suggestion: Please move this outside of the struct creation (same as above).

crates/stackable-operator/src/commons/authentication/ldap.rs
```diff
@@ -28,6 +31,14 @@ pub enum Error {
 
     #[snafu(display("failed to add LDAP TLS client details volumes and volume mounts"))]
     AddLdapTlsClientDetailsVolumes { source: TlsClientDetailsError },
+
+    #[snafu(display("failed to add needed volumes"))]
```
Techassi, line 35: suggestion:

```suggestion
    #[snafu(display("failed to add required volumes"))]
```

crates/stackable-operator/src/commons/authentication/ldap.rs
```diff
@@ -28,6 +31,14 @@ pub enum Error {
...
+    AddVolumes { source: builder::pod::Error },
+
+    #[snafu(display("failed to add needed volumeMounts"))]
```
Techassi, line 38: suggestion:

```suggestion
    #[snafu(display("failed to add required volumeMounts"))]
```

crates/stackable-operator/src/commons/s3/mod.rs
```diff
@@ -31,4 +33,12 @@ pub enum S3Error {
 
     #[snafu(display("failed to add S3 TLS client details volumes and volume mounts"))]
     AddS3TlsClientDetailsVolumes { source: TlsClientDetailsError },
+
+    #[snafu(display("failed to add needed volumes"))]
+    AddVolumes { source: builder::pod::Error },
+
+    #[snafu(display("failed to add needed volumeMounts"))]
```
Techassi, line 40: suggestion:

```suggestion
    #[snafu(display("failed to add required volumes"))]
    AddVolumes { source: builder::pod::Error },

    #[snafu(display("failed to add required volumeMounts"))]
```

crates/stackable-operator/src/commons/tls_verification.rs
```diff
@@ -4,17 +4,28 @@ use serde::{Deserialize, Serialize};
...
-#[derive(Debug, PartialEq, Snafu)]
+#[derive(Debug, Snafu)]
 pub enum TlsClientDetailsError {
     #[snafu(display("failed to convert secret class volume into named Kubernetes volume"))]
     SecretClassVolume { source: SecretClassVolumeError },
+
+    #[snafu(display("failed to add needed volumes"))]
+    AddVolumes { source: builder::pod::Error },
+
+    #[snafu(display("failed to add needed volumeMounts"))]
```
Techassi, line 25: suggestion:

```suggestion
    #[snafu(display("failed to add required volumes"))]
    AddVolumes { source: builder::pod::Error },

    #[snafu(display("failed to add required volumeMounts"))]
```

crates/stackable-operator/src/product_logging/framework.rs
```diff
@@ -31,6 +33,19 @@ const SHUTDOWN_FILE: &str = "shutdown";
...
 pub const VECTOR_CONFIG_FILE: &str = "vector.yaml";
 
+#[derive(Debug, Snafu)]
+pub enum LoggingError {
+    #[snafu(display("failed to create container"))]
+    CreateContainer {
+        source: builder::pod::container::Error,
+    },
+
+    #[snafu(display("failed to add needed volumeMounts"))]
```
Techassi, line 43: suggestion:

```suggestion
    #[snafu(display("failed to add required volumeMounts"))]
```

Techassi (changes requested):
A bunch of small comments.

crates/stackable-operator/src/builder/pod/container.rs
```diff
@@ -21,11 +25,25 @@ pub enum Error {
...
     },
+
+    #[snafu(display(
+        "The volumeMount is clashing with an already existing volumeMount with the same mountPath but a different content. \
+            The existing volumeMount is {existing_volume_mount:?}, the new one is {new_volume_mount:?}"
+    ))]
+    ClashingVolumeMountMountPath {
+        existing_volume_mount: Box<VolumeMount>,
+        new_volume_mount: Box<VolumeMount>,
+    },
```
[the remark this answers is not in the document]
sbernauer: We only have access to the volume, *not* the volumeMount, so we don't know the mountPath.
We could only say 'the volume "tls" clashes, bruh', which IMHO is too less information. It's not a strong opinion but I think we would be very happy to know what that volume is (emptyDir, configMap or secret-op volume) in case we get such errors.

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Nope, my fault ;) 1a3f95f

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: I mostly added this to be consistent with PodBuilder and volume addition. Also docs brake, will take care of that later once we have settled on docs.
Anyway, made it private in a814f41cf79fb7ed825a1fade13b2790221ad023

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Moved in b7fcdaa
I could offer the following code, but IMHO it's not really better, WDYT?
```rust
        let volume_mounts = self
            .volume_mounts
            .is_empty()
            .then(|| self.volume_mounts.values().cloned().collect());
```

crates/stackable-operator/src/builder/pod/mod.rs
sbernauer: Moved in b7fcdaa

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Actually it depends on the interpretation of duplicates. For me duplicates are totally fine.
WDYT?
```
    /// Adds a new [`VolumeMount`] to the container while ensuring that no
    /// clashing [`VolumeMount`] exists.
    ///
    /// A clashing [`VolumeMount`] would have same mountPath but a different
    /// content than another [`VolumeMount`]. An appropriate error is returned
    /// when such a clashing mount path is encountered.
    ///
    /// ### Note
    ///
    /// Previously, this function unconditionally added [`VolumeMount`]s, which
    /// resulted in invalid [`PodSpec`][1]s.
    ///
    /// [1]: k8s_openapi::api::core::v1::PodSpec
    ```

File: crates/stackable-operator/src/builder/pod/mod.rs (hunk 16 above)
Comment (sbernauer) [reply crates/stackable-operator/src/builder/pod/mod.rs:292]: We only have access to the volume, not the volumeMount, so we don't know the mountPath.
We could only say 'the volume "tls" clashes, bruh', which IMHO is too less information. It's not a strong opinion but I think we would be very happy to know what that volume is (emptyDir, configMap or secret-op volume) in case we get such errors.

File: crates/stackable-operator/src/builder/pod/mod.rs (hunk 15 above)
Comment (sbernauer) [reply crates/stackable-operator/src/builder/pod/mod.rs:61]: We only have access to the volume, not the volumeMount, so we don't know the mountPath.
We could only say 'the volume "tls" clashes, bruh', which IMHO is too less information. It's not a strong opinion but I think we would be very happy to know what that volume is (emptyDir, configMap or secret-op volume) in case we get such errors.

File: crates/stackable-operator/src/builder/pod/mod.rs (hunk 16 above)
Comment (sbernauer) [reply crates/stackable-operator/src/builder/pod/mod.rs:292]: I'll wait on https://github.com/stackabletech/operator-rs/pull/871#discussion_r1771122689 before changing things

File: crates/stackable-operator/src/builder/pod/container.rs (hunk 24 above)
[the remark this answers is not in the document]
Comment (sbernauer) [reply crates/stackable-operator/src/builder/pod/container.rs:36]: Actually, looking into VolumeMount it also contains
```rust
pub sub_path: Option<String>,
# or 
pub sub_path_expr: Option<String>,
```
which I would prefer to keep (as it might be totally ok as the subpaths differ(?))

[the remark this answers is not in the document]
sbernauer: I mean I'd say the target audience for the error message are we at Stackable (a user should never see this error, it's our fault), so I think better include to much than to less information.

crates/stackable-operator/src/builder/pod/container.rs
Techassi: I feel like the combinators are a little better, but I wouldn't block this PR on it.

crates/stackable-operator/src/builder/pod/container.rs
Techassi: For me, two or more volume mounts are duplicated when the mount paths are identical, because it is effectively the same volume mount. The same mount path can only be referenced at most once per container.

crates/stackable-operator/src/builder/pod/container.rs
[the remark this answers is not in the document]
Techassi: > which I would prefer to keep

Sure, then let's carry individual fields in the error message instead of the whole struct.

> I mean I'd say the target audience for the error message are we at Stackable

Yes, but I would like to prepare stackable-operator and friends to be used outside of Stackable. That's why we also need to focus on what other users might expect from these errors. And reducing the errors in complexity and focusing on the important bits of information is important in that case.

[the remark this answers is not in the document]
sbernauer: Pushed cdf5c36, where I carry individual fields in the error message instead of the whole struct.

crates/stackable-operator/src/builder/pod/mod.rs
sbernauer: We need to box it to avoid stack out of memory errors, so we put the volumes on the heap. IIRC we already need to do that for other error contents as well.
And yeah, Volume is basically a giant enum

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: For me mounts are duplicated when they are the same (so same everything, as `==` would check.) And it's totally fine to have duplicates (as the same ca crt can be added multiple times, e.g. because you have 5 identical S3Connection references), the problems start when we have two different mounts (e.g. pulling in different volumes) sharing the same mountPath.

> Adds a new [`VolumeMount`] to the container while ensuring that no duplicate exists.

I would disagree and say it's totally fine to have duplicates. But it's not okay to have clashing mounts (same path, different content)

crates/stackable-operator/src/builder/pod/container.rs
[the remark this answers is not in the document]
sbernauer: Actually I needed to box the 7 Strings in 8727f9a, which is a bit ugly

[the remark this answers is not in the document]
Techassi: Oh, I was thinking about only using 2 or 3 strings in the message, not all of them. If we _really_ want them all, we could go back to just passing the struct.

I really doubt that we need _all_ fields in the error message. That amount of detail can be provided via tracing/logging if we want to.

So bottom-line: Either use 2-3 strings and optional logging (preferred) or the complete struct.

crates/stackable-operator/src/builder/pod/container.rs
Techassi: So I feel like we are talking about the same thing here. A duplicate mount is a mount which uses the _same_ mount path.

Mounts which are identical (so `==`) are still duplicates, but more specifically copies of each other.

I'm fine with your proposed doc comment if you find the usage of the word "clashing" more suitable then "duplicate".

crates/stackable-operator/src/builder/pod/mod.rs
Techassi: So as I said, I would avoid including the volume in the error. More details can be provided via tracing/logging.

> And yeah, Volume is basically a giant enum

That would actually be fine, as enums will only carry one variant at runtime. But `Volume` is actually a struct with 30+ fields (which are mostly empty, but still). I really don't see the need to include this (twice) in the error message.

crates/stackable-operator/src/builder/pod/container.rs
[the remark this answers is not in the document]
NickLarsenNZ: Agreed.

Only because the user can't possibly fix this problem, so they should get "some" detail on what happened, and then when we receive the issue, we can turn up debug logging to see all the extra detail.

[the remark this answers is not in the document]
sbernauer: I removed the subPathEpr in e40749a, so the enum is small enough again for clippy to stop warning. IMHO we need to keep the subPath, as we are actually using it

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: I think clashing describes it better than duplicate and would pick the proposed doc comment. @NickLarsenNZ please feel free to object

crates/stackable-operator/src/builder/pod/container.rs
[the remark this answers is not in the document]
sbernauer: Implemented in a sync meeting

crates/stackable-operator/src/builder/pod/container.rs
sbernauer: Implemented in a sync meeting

crates/stackable-operator/src/builder/pod/mod.rs
sbernauer: Implemented in a sync meeting

crates/stackable-operator/src/builder/pod/mod.rs
sbernauer: Implemented in a sync meeting

crates/stackable-operator/src/builder/pod/mod.rs
sbernauer: It's in tests

Techassi (approved):

PR #872, opened 2024-09-20
feat(stackable-versioned): Use enum for merged_crd instead of str

Techassi:

This ensures that the user must use a valid version to specify which of the versions of the merged CRDs is stored. Previously, users could provide any `str`, which can lead to runtime errors.

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

### DIFF
--- a/crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
+++ b/crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
@@ -8,12 +8,17 @@ use syn::{parse_quote, DataStruct, Error, Ident};
 use crate::{
     attrs::common::ContainerAttributes,
     codegen::{
-        common::{Container, ContainerInput, ContainerVersion, Item, VersionedContainer},
+        common::{
+            Container, ContainerInput, ContainerVersion, Item, KubernetesOptions, VersionExt,
+            VersionedContainer,
+        },
         vstruct::field::VersionedField,
     },
 };
 
 pub(crate) mod field;
+
+type GenerateVersionReturn = (TokenStream, Option<(TokenStream, (Ident, String))>);
 
 /// Stores individual versions of a single struct. Each version tracks field
 /// actions, which describe if the field was added, renamed or deprecated in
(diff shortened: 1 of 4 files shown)


crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
```diff
@@ -85,24 +90,30 @@ impl Container<DataStruct, VersionedField> for VersionedStruct {
     }
 
     fn generate_tokens(&self) -> TokenStream {
-        let mut kubernetes_crd_fn_calls = TokenStream::new();
-        let mut container_definition = TokenStream::new();
+        let mut tokens = TokenStream::new();
+
+        let mut enum_variants = Vec::new();
```
NickLarsenNZ, line 95: nit: Maybe this should be called `version_enum_variants` to not be confused with a versioned enum.

But that got me thinking, this is in `vstruct`, and I don't see the equivalent code for `venum`. Should this be generic across containers?

NickLarsenNZ (approved):
LGTM, just a query about enum containers and where this code should be.

crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
Techassi: > But that got me thinking, this is in vstruct, and I don't see the equivalent code for venum. Should this be generic across containers?

All Kubernetes (CRD) specific stuff can only be applied to structs. We also already only generate CRD code (derive macro and merged_crd function) for structs. Enums just cannot be used _as_ a CRD, but only as part of it.

NickLarsenNZ: Oh yeah, that makes sense, and I had a strong feeling I was missing something obvious.

NickLarsenNZ: I still think the variable should be renamed for clarity, but that isn't a blocker.

I had an idea about the `Version` enum, it could perhaps be defined in a `k8` (or similar) module. gain, not a blocker, and just adding to past discussion about it.

Techassi: > I still think the variable should be renamed for clarity, but that isn't a blocker.

Sure, I can do that.

> I had an idea about the Version enum, it could perhaps be defined in a k8 (or similar) module. gain, not a blocker, and just adding to past discussion about it.

What about the `Foo` struct used to call `Foo::merged_crd()` on? Should that then also live in the `k8s` module?

NickLarsenNZ: > What about the `Foo` struct used to call `Foo::merged_crd()` on? Should that then also live in the `k8s` module?

Oh good point. It could, but the the struct is not really needed.

We could break the rules, and make a module called `Foo` (capitalised), so it can have both the static `merged_crd()` method, as well as the `Version` struct.

We can then also forgo the naming rules for variants and keep the _familiar_ lowe-case names.

```rust
#[allow(non_snake_case)]
pub mod Foo {
    #[allow(non_camel_case_types)]
    pub enum Version {
        v1alpha1,
        v1alpha2,
    }

    pub fn merged_crd() -> _ { ... }
}

```

Techassi: What do you think about experimenting with this idea in a separate PR and merge the current one as is for now?

The idea seems interesting and I would like to try to make it work so that we can explore how usable it is.

Techassi: Follow-up: #875

PR #875, opened 2024-09-24
refactor(stackable-versioned): Rework version enum for merged_crd

Techassi:
Part of https://github.com/stackabletech/issues/issues/507, follow-up of https://github.com/stackabletech/operator-rs/pull/872#discussion_r1770889930

This PR adjusts the version enum generation used in the `merged_crd` function.

```rust
let merged_crd = Foo::merged_crd(Foo::V1).unwrap();
println!("{}", serde_yaml::to_string(&merged_crd).unwrap());
```

### DIFF
--- a/crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
+++ b/crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
@@ -322,12 +322,12 @@ impl VersionedStruct {
         let enum_ident = &self.idents.kubernetes;
         let enum_vis = &self.visibility;
 
-        let mut enum_variant_matches = TokenStream::new();
+        let mut enum_display_impl_matches = TokenStream::new();
         let mut enum_variant_idents = TokenStream::new();
 
         for (enum_variant_ident, enum_variant_display) in enum_variants {
             enum_variant_idents.extend(quote! {#enum_variant_ident,});
-            enum_variant_matches.extend(quote! {
+            enum_display_impl_matches.extend(quote! {
                 #enum_ident::#enum_variant_ident => f.write_str(#enum_variant_display),
             });
         }
@@ -342,7 +342,7 @@ impl VersionedStruct {
             impl ::std::fmt::Display for #enum_ident {
                 fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::result::Result<(), ::std::fmt::Error> {
                     match self {
-                        #enum_variant_matches
+                        #enum_display_impl_matches
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
```diff
@@ -319,22 +319,40 @@ impl VersionedStruct {
...
 
-        let version_enum_definition = self.generate_kubernetes_version_enum(enum_variants);
+        let mut enum_variant_matches = TokenStream::new();
+        let mut enum_variant_idents = TokenStream::new();
+
+        for (enum_variant_ident, enum_variant_display) in enum_variants {
+            enum_variant_idents.extend(quote! {#enum_variant_ident,});
+            enum_variant_matches.extend(quote! {
+                #enum_ident::#enum_variant_ident => f.write_str(#enum_variant_display),
+            });
```
NickLarsenNZ, line 332: **nit**: These appear to only be for Display impl match arms, so we should name it accordingly.

```suggestion
            enum_display_impl_matches.extend(quote! {
                #enum_ident::#enum_variant_ident => f.write_str(#enum_variant_display),
            });
```

crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
```diff
@@ -319,22 +319,40 @@ impl VersionedStruct {
...
+        }
 
         quote! {
             #[automatically_derived]
-            pub struct #ident;
+            #enum_vis enum #enum_ident {
+                #enum_variant_idents
+            }
```
NickLarsenNZ, line 339: **thought (non-blocker)**: I think ultimately this will move outside of the kubernetes specific stuff. Eg: for semver, there might be `Foo:1_2_3`. But for now, it is fine here.

NickLarsenNZ: Maybe the versions should be available even if the k8s feature is disabled.

NickLarsenNZ (approved):
LGTM. Some suggestions for improvements and future proofing.

crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
Techassi: Yes, makes sense. I will rename locally, because the suggestion doesn't include the `enum_display_impl_matches` declaration.

Techassi: Renamed in c04023a

crates/stackable-versioned-macros/src/codegen/vstruct/mod.rs
Techassi: Yes, I can see that it might be useful to generate the versions even when no K8s specific features are used. Do we want to make the required changes in this PR?

NickLarsenNZ: I'd be ok moving it later.

NickLarsenNZ (approved):
LGTM

PR #881, opened 2024-09-27
test(stackable-versioned): Add snapshot testing

Techassi:
This PR adds snapshot tests to enable macro output testing.

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


crates/stackable-versioned-macros/fixtures/README.md
```diff
@@ -0,0 +1,54 @@
...
+    )]
+    bar: usize,
+    baz: bool,
+}
+```
+
+## Recommended Workflow
+
+First, add new input files (which automatically get picked up by `insta`) to the `fixtures/inputs`
+folder. Make sure the delimiter is placed correctly between the attribute and the container
```
NickLarsenNZ, line 41: Probably no need to add it, but this includes doc-comments.

crates/stackable-versioned-macros/fixtures/README.md
```diff
@@ -0,0 +1,54 @@
...
+definition. Next, generate the snapshot files (initially not accepted) by running
+
+```shell
+cargo insta test -p stackable-versioned-macros
+```
+
+This command will place the new snapshot files in the `fixtures/snapshots` folder. To review them,
```
NickLarsenNZ, line 48: Could be worth mentioning that they are suffixed with `.new`, and should not be committed unless internationally (eg: for collaboration). They should not appear on the `main` branch.

crates/stackable-versioned-macros/fixtures/README.md
```diff
@@ -0,0 +1,54 @@
...
+run the `cargo insta review` command. Now check these files in.
```
NickLarsenNZ, line 49: I thought maybe some elaboration was needed:

```suggestion
run the `cargo insta review` command, then accept or fix the snapshots. Once all are accepted (ie: no `.new` files remaining), check in the files.
```

crates/stackable-versioned-macros/tests/trybuild.rs
```diff
@@ -37,7 +26,6 @@ mod default {
 #[test]
 fn default_macros() {
     let t = trybuild::TestCases::new();
-    t.pass("tests/default/pass/*.rs");
```
NickLarsenNZ, line 40: I'm assuming insta can't snapshots test the failed compilation?

Just checking if trybuild is still useful.

NickLarsenNZ (approved):
LGTM, a couple of suggestions and a question.

crates/stackable-versioned-macros/tests/trybuild.rs
Techassi: I haven't looked into it, but I have a feeling `insta` can't do it. So `trybuild` would still be needed.

crates/stackable-versioned-macros/fixtures/README.md
Techassi: Added in cf1ec21

crates/stackable-versioned-macros/fixtures/README.md
Techassi: Added in cf1ec21

crates/stackable-versioned-macros/tests/trybuild.rs
Techassi: `insta` currently cannot do it. For now, we keep the (failing) `trybuild` test cases.

NickLarsenNZ (approved):
I think it is still a good idea to keep the fail folder to make it obvious to the casual browser (without seeing the logic in trybuild.rs).

NickLarsenNZ (approved):
LGTM

PR #883, opened 2024-10-07
feat!: Add stackable-shared crate

Techassi:
Part of https://github.com/stackabletech/issues/issues/642, prepares for YAML helpers in `stackable-versioned` (see https://github.com/stackabletech/operator-rs/tree/feat/stackable-versioned-yaml-output branch)

This crate initially contains code to be able to use our custom YAML serialization and CRD schema generation across workspace members. Previously, to get access to the specialized helper functions and traits one needed to import the complete stackable-operator crate. That's why this commit moves that piece of code into a shared place, which is way lighter to be imported by other workspace members. It additionally reworks the helpers to be slightly more generic. It still contains the docs URL replacer, which replaces the placeholder with the correct Stackable doc URL in doc comments and thus in OpenAPI schema descriptions.

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


sbernauer:
New day new crate ^^

crates/stackable-shared/src/yaml.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    fn replace(&self, input: &str) -> Result<String> {
+        let docs_version = match self.0 {
+            "0.0.0-dev" => "nightly".to_owned(),
+            ver => {
+                let v = Version::from_str(ver).context(ParseSemanticVersionSnafu { input })?;
+                format!("{major}.{minor}", major = v.major, minor = v.minor)
+            }
+        };
```
sbernauer, line 51: Why not parse the version in DocUrlReplacer::new() and store the Version in DocUrlReplacer?

sbernauer (changes requested):
LGTM otherwise

crates/stackable-shared/src/yaml.rs
Techassi: Oh wow, what an obvious improvement. So thanks for pointing it out :+1:

Techassi: Done in df719d5

sbernauer: Thanks!

sbernauer (approved):

lfrancke:
You added the release note label but no release notes. Is this a user visible change?

sbernauer:
No, this is an internal change (theoretically users of the Rust crates are affected, but not end users)

lfrancke:
In that case I'll remove the label again, thank you.

PR #901, opened 2024-10-24
refactor(stackable-telemetry)!: Improve stackable-telemetry to make it easier to build the TraceGuard

NickLarsenNZ:
# Description

Part of https://github.com/stackabletech/issues/issues/639.

## Problem

It is cumbersome to configured the various subscribers to be enabled at runtime. Using @soenkeliebau's code example:

```rust
// Instead of building a TraceGuard, a mutable builer needs to be assigned so that...
let mut builder = Tracing::builder()
    .service_name("whoyougonnacall")
    .with_console_output("WYGC_CONSOLE", LevelFilter::INFO);

// ... builder methods can be called conditionally:
if enable_trace_exporter().context(ParseConfigSnafu)? {
    builder = builder.with_otlp_trace_exporter("TEST_OTLP_TRACE", LevelFilter::TRACE);
}
if enable_log_exporter().context(ParseConfigSnafu)? {
    builder = builder.with_otlp_log_exporter("TEST_OTLP_LOG", LevelFilter::TRACE);
}

// Then finally the trace TraceGuard is returned.
let _tracing_guard = builder.build().init().context(InitializeTelemetrySnafu)?;
```

The purpose of the builder is to allow overriding of the the defaults, however in the current state, the function parameters for the subscribers (console and otlp exporters) break that, and it will only get messier as config options are added.

## Solution

Add common and subscriber specific configuration through a Settings builder (suggested by @Techassi). The previous example would then look like:

```rust
let _trace_guard = Tracing::builder()
    .service_name("whoyougonnacall")
    .with_console_output(
        Settings::builder()
            .env_var("WYGC_CONSOLE")
            .default_level(LevelFilter::INFO)
            .enabled(true)
            // 👇 this part is new, and has no analogue in the example from the problem statement above.
            .console_log_settings_builder()
            .log_format(Format::Plain)
            // ☝️
            .build(),
    )
    .with_otlp_trace_exporter(
        Settings::builder()
            .env_var("TEST_OTLP_TRACE")
            .default_level(LevelFilter::TRACE)
            .enabled(enable_trace_exporter().context(ParseConfigSnafu)?)
            .build(),
    )
    .with_otlp_log_exporter(
        Settings::builder()
            .env_var("TEST_OTLP_LOG")
            .default_level(LevelFilter::TRACE)
            .enabled(enable_log_exporter().context(ParseConfigSnafu)?)
            .build(),
    )
    .build();
```

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


soenkeliebau:
I like it!

NickLarsenNZ:
> I like it!

It is basically done, there is one thing left to refactor (and enable console log formats), but before merging I will document the pub items (and enable warn on missing docs).

NickLarsenNZ:
This PR was going to include the following items:
- complete docs
- tuples for quicker setup
- console json format output as an option
- 
But they were dropped in order to unblock @xeniape for the RollingFileAppender implementation. I am promising to come back and do those things in a separate PR.

crates/stackable-telemetry/src/tracing/settings/console_log.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.log_format = format;
+        self
+    }
+
+    pub fn build(self) -> ConsoleLogSettings {
+        self.into()
+    }
+}
+
+impl From<ConsoleLogSettingsBuilder> for ConsoleLogSettings {
```
nightkr, line 55: Wouldn't this also be `Build`? At least it seems inconsistent to sometimes use `From` and sometimes `Build`.

crates/stackable-telemetry/src/tracing/settings/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Simplifies access common settings from subscriber specific settings.
+pub trait CommonSettings {
+    /// Access to the [`Settings::environment_variable`] field.
+    fn environment_variable(&self) -> &'static str;
+
+    /// Access to the [`Settings::default_level`] field.
+    fn default_level(&self) -> LevelFilter;
+
+    /// Access to the [`Settings::enabled`] field.
+    fn enabled(&self) -> bool;
```
nightkr, line 23: Would a disabled subscriber even exist beyond the builder stage?

crates/stackable-telemetry/src/tracing/settings/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+pub mod otlp_trace;
+pub use otlp_trace::*;
+
+/// Simplifies access common settings from subscriber specific settings.
+pub trait CommonSettings {
+    /// Access to the [`Settings::environment_variable`] field.
+    fn environment_variable(&self) -> &'static str;
+
+    /// Access to the [`Settings::default_level`] field.
+    fn default_level(&self) -> LevelFilter;
```
nightkr, line 20: This could be isolated from each impl by defining a

```rust
fn common(&self) -> &Settings;
```

and then providing default impls for everything else that delegates to it:

```suggestion
    /// Access to the [`Settings::environment_variable`] field.
    fn environment_variable(&self) -> &'static str {
      self.common().environment_variable
    }

    /// Access to the [`Settings::default_level`] field.
    fn default_level(&self) -> LevelFilter {
      self.common().default_level
    }
```

nightkr (commented):
Not really relevant since it's excluded from this PR, but the tuple variant smells a lot like "`From` magic" that I would rather avoid.

crates/stackable-telemetry/src/tracing/settings/mod.rs
Techassi: Agreed, great idea!

nightkr: Another alternative would be to have each variant provide a `Deref<Target = Settings>`.

Techassi: Oh yeah, also a nice idea. I will first try out the `common()` approach and then I can look into the `Deref` solution.

Techassi: First iteration in 5de6ed4

crates/stackable-telemetry/src/tracing/settings/mod.rs
Techassi: No, but we still need access to the `enabled` field, which is part of the common `Settings` struct.

nightkr: I'm not going to cause a fuss over it right now, but it would be nice to see if we can kill it off eventually.

crates/stackable-telemetry/src/tracing/settings/mod.rs
Techassi: Second iteration in 8f59b15

crates/stackable-telemetry/src/tracing/settings/mod.rs
NickLarsenNZ: I'm unsure what you mean, can you describe a little more concretely?

The actual Tracing Subscriber/Layer is never created/registered if `enabled` is false.

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -14,12 +14,16 @@ use opentelemetry_sdk::{
     trace, Resource,
 };
 use opentelemetry_semantic_conventions::resource;
+use settings::{ConsoleLogSettings, OtlpLogSettings, OtlpTraceSettings};
```
Techassi, line 17: nit: This should be moved down (not an external crate).

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -14,12 +14,16 @@ use opentelemetry_sdk::{
...
 use snafu::{ResultExt as _, Snafu};
-use tracing::{level_filters::LevelFilter, subscriber::SetGlobalDefaultError};
+use tracing::subscriber::SetGlobalDefaultError;
 use tracing_subscriber::{filter::Directive, layer::SubscriberExt, EnvFilter, Layer, Registry};
 
+pub mod settings;
+
 type Result<T, E = Error> = std::result::Result<T, E>;
 
+#[allow(missing_docs)]
```
Techassi, line 26: nit: I think we _currently_ don't need that attribute.

crates/stackable-telemetry/src/tracing/mod.rs
NickLarsenNZ: Done in afc21636bf3ec0324aabd3dcf20b65dfe38de946

crates/stackable-telemetry/src/tracing/mod.rs
NickLarsenNZ: Done in afc21636bf3ec0324aabd3dcf20b65dfe38de946

crates/stackable-telemetry/src/tracing/settings/mod.rs
nightkr: I haven't tested it or thought it through completely, but in short, you could use `bool::then` and let the `with_foo` take an `Option<FooSettings>`. Something like:

```rust
.with_otlp_log_exporter(
  otlp_log_flag.then(|| Settings::builder()
    .blah()
    .build()
  )
)
```

Techassi (approved):

Techassi (approved):

crates/stackable-telemetry/src/tracing/settings/mod.rs
NickLarsenNZ: I have added an item to the parent issue: https://github.com/stackabletech/issues/issues/639

NickLarsenNZ: With the help of @Techassi this has been done in https://github.com/stackabletech/operator-rs/pull/951

PR #909, opened 2024-11-21
Fixes accidentally deleted serviceaccount and rolebinding objects

soenkeliebau:
# Description
Currently it is possible that `servieAccount` and `roleBinding` objects accidentally get deleted if multiple clusters of the same type (airflow, zookeeper, ...) get deployed to the same namespace and one of them gets deleted.

The issue is, that we don't name the serviceaccount and rolebinding objects after the clustername, but they are named the same for all cluster, so bacially just "airflow-rolebinding" and "airflow-serviceaccount"  across the board.
When you new deploy two airflows into the same namespace and delete the wrong one of them again (the wrong one being the last that was deployed) then the rolebinding and serviceaccount are gone as well and the other cluster stops working as well.

Reason for "the last one" being the one that kills it is, because the ownerreferences of the rolebinding and serviceaccount point at this and the delete gets propagated to these two objects that are still needed by the other clusters.

This PR fixes how the names are built for the serviceAccount and roleBinding objects are created  so that they are unique for each cluster.

fixes https://github.com/stackabletech/issues/issues/675

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


crates/stackable-operator/src/commons/rbac.rs
```diff
@@ -30,12 +30,18 @@ pub enum Error {
 /// Build RBAC objects for the product workloads.
 /// The `rbac_prefix` is meant to be the product name, for example: zookeeper, airflow, etc.
 /// and it is a assumed that a ClusterRole named `{rbac_prefix}-clusterrole` exists.
+/// 'rbac_prefix' is not used to build the names of the serviceAccount and roleBinding objects,
+/// as this caused problems with multiple clusters of the same product within the same namespace
+/// see <https://stackable.atlassian.net/browse/SUP-148> for more details.
+/// Instead the names for these objects are created by reading the name from the cluster object
+/// and appending [-rolebinding|-serviceaccount] to create unique names instead of using the
+/// same objects for multiple clusters.
```
nightkr, line 38: Not a blocker and we can always fix it later, but this feels more like a changelog entry or a local comment than a doccomment to me.

crates/stackable-operator/src/commons/rbac.rs
```diff
@@ -30,12 +30,18 @@ pub enum Error {
...
 pub fn build_rbac_resources<T: Clone + Resource<DynamicType = ()>>(
     resource: &T,
     rbac_prefix: &str,
     labels: Labels,
 ) -> Result<(ServiceAccount, RoleBinding)> {
-    let sa_name = service_account_name(rbac_prefix);
+    let sa_name = service_account_name(&resource.name_any());
```
nightkr, line 44: The rolebinding should also cover the old name for at least one release, or we'll have downtime during the switchover.

nightkr: (Since existing pods will have the old ServiceAccount name, which will no longer be bound to any roles after this change goes through.)

crates/stackable-operator/src/commons/rbac.rs
soenkeliebau: that is true indeed

crates/stackable-operator/src/commons/rbac.rs
soenkeliebau: changed it. We probably want to track that on our deprecation board?

nightkr: Absolutely.

crates/stackable-operator/src/commons/rbac.rs
```diff
@@ -28,14 +28,24 @@ pub enum Error {
...
+    // Instead the names for these objects are created by reading the name from the cluster object
+    // and appending [-rolebinding|-serviceaccount] to create unique names instead of using the
+    // same objects for multiple clusters.
+    product_name: &str,
     labels: Labels,
 ) -> Result<(ServiceAccount, RoleBinding)> {
-    let sa_name = service_account_name(rbac_prefix);
+    let sa_name = service_account_name(&resource.name_any());
+    // We add the legacy serviceAccount name to the binding here for at least one
+    // release cycle, so that the switchover during the upgrade can be smoother.
```
nightkr, line 47:
```suggestion
    // We add the legacy serviceAccount name to the binding here for at least one
    // release cycle, so that the switchover during the upgrade can be smoother.
    // To be removed in v24.3+1.
```

crates/stackable-operator/src/commons/rbac.rs
soenkeliebau: We are just discussing if this is really necessary.. since we delete the old serviceAccount pretty much immediately, won't the double roleBinding be ineffective anyway?

nightkr: It's worth testing, but AFAIK deleting the old ServiceAccount will only prevent new Pods from being spawned that use it. Any existing pods would still have a valid authn token for the old SA, and their requests still need to be authorized correctly until they're migrated over to the new SA. We're still talking about a relatively short window, but on the order of minutes, not seconds.

However, it gets slightly worse. StatefulSet upgrades will only progress if the STS replicas are in a healthy state, so services that require ongoing K8s access would effectively get stuck in a broken state until it is taken down manually.

I'm not sure how many of our products actually require K8s access these days, but I suspect at least NiFi would be a likely problem case?

nightkr (dismissed):

siegfriedweber (approved):
LGTM

PR #913, opened 2024-11-27
fix(stackable-versioned): Generate Kubernetes code independent of container order

Techassi:
Part of https://github.com/stackabletech/issues/issues/642, follow-up of https://github.com/stackabletech/operator-rs/pull/912

This fixes the wrong generation of Kubernetes specific code which could be triggered by either

- versioning enums alongside structs in a module or
- versioning containers with and without the `k8s` argument in mixed order


NickLarsenNZ (approved):
LGTM

PR #914, opened 2024-11-27
feat(stackable-versioned)!: Pass through arguments to the kube attribute

NickLarsenNZ:
Part of https://github.com/stackabletech/issues/issues/642.

# Description

- Pass through arguments to the `#[kube]` attribute:
  - `status`.
  - `crates`.
  - `shortname`.
- Add support for overriding `kube::core` and `k8s_openapi` in generated code.
- **BREAKING**: Remove the `CustomResource` derive.
  This is because we can't know whether they will use the one from the crate `::kube::CustomResource`, via a re-export like we do in stackable operators.


Techassi (approved):
The crate overrides need to be improved, but that will be tackled as discussed with @NickLarsenNZ.

PR #919, opened 2024-11-28
refactor(stackable-versioned): Simplify crate overrides

Techassi:

This simplifies the handling and generation of crate overrides.

### DIFF
--- a/crates/stackable-versioned-macros/src/codegen/container/mod.rs
+++ b/crates/stackable-versioned-macros/src/codegen/container/mod.rs
@@ -1,7 +1,9 @@
+use std::ops::Deref;
+
 use darling::{util::IdentString, Result};
 use proc_macro2::TokenStream;
-use quote::quote;
-use syn::{Attribute, Ident, ItemEnum, ItemStruct, Visibility};
+use quote::{quote, ToTokens};
+use syn::{parse_quote, Attribute, Ident, ItemEnum, ItemStruct, Path, Visibility};
 
 use crate::{
     attrs::{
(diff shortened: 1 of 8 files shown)


crates/stackable-versioned-macros/src/codegen/container/mod.rs
```diff
@@ -282,10 +284,136 @@ impl From<KubernetesArguments> for KubernetesOptions {
...
+            crate_overrides.extend(quote! { serde = #serde, });
+        }
+
+        if !crate_overrides.is_empty() {
+            tokens.extend(quote! { , crates(#crate_overrides) });
+        }
+    }
+}
+
+#[derive(Debug)]
```
NickLarsenNZ, line 384: Suggestion: docs

crates/stackable-versioned-macros/src/codegen/container/mod.rs
```diff
@@ -282,10 +284,136 @@ impl From<KubernetesArguments> for KubernetesOptions {
...
+pub(crate) struct Override<T> {
+    is_overridden: bool,
+    inner: T,
+}
+
+impl<T> Override<T> {
+    pub(crate) fn new_default(inner: T) -> Self {
```
NickLarsenNZ, line 391:
```suggestion
    /// Mark a value as a default.
    ///
    /// This is used to indicate that the value is a default and was not overridden.
    pub(crate) fn new_default(inner: T) -> Self {
```

crates/stackable-versioned-macros/src/codegen/container/mod.rs
```diff
@@ -282,10 +284,136 @@ impl From<KubernetesArguments> for KubernetesOptions {
...
+        Override {
+            is_overridden: false,
+            inner,
+        }
+    }
+
+    pub(crate) fn new_custom(inner: T) -> Self {
```
NickLarsenNZ, line 398:
```suggestion
    /// Mark a value as overridden.
    ///
    /// This is used to indicate that the value was overridden and not the default.
    pub(crate) fn new_custom(inner: T) -> Self {
```

NickLarsenNZ (dismissed):
LGTM

NickLarsenNZ (approved):
LGTM

PR #920, opened 2024-11-28
fix(stackable-versioned): Emit correct enum based on kube kind argument

Techassi:

The enum used to generate the merged CRD function now uses the correct ident if customized via `#[kube/k8s(kind = "MyName")]`.


NickLarsenNZ (approved):
LGTM

PR #923, opened 2024-11-29
feat(stackable-versioned): Support visibility for CRD enum

Techassi:

The generated CRD enum now uses the same visibility as the container definition (standalone mode) or the module (nested mode).

### Before

```rust
#[versioned]
pub(crate) struct FooSpec {}

// Expanded to ...

pub enum Foo {}
```

### After

```rust
#[versioned]
pub(crate) struct FooSpec {}

// Expands to ...

pub(crate) enum Foo {}
```


NickLarsenNZ (approved):
LGTM

PR #924, opened 2024-11-29
chore(stackable-versioned): Remove {write,print}_merged_crd function

Techassi:

The dependency on stackable-shared can cause some conflicts in projects which use stackable-versioned. This also de-couples Stackable specific code from the versioning macro, which we want to be usable outside of Stackable as well.

Consumers can still generate the merged CRD and then use functions from the `stackable_shared::CustomResourceExt` trait to print or write the YAML.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3281,7 +3281,6 @@ version = "0.4.1"
 dependencies = [
  "kube",
  "snafu 0.8.5",
- "stackable-shared",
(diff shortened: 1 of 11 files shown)


NickLarsenNZ (changes requested):
The stackable-shared dependency can be removed now.

NickLarsenNZ (approved):
LGTM

PR #931, opened 2024-12-13
feat!: Add jvmArgumentOverrides as well as common code

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/584

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


crates/stackable-operator/src/role_utils.rs
```diff
@@ -144,10 +155,129 @@ pub struct CommonConfiguration<T> {
...
+                let with_anchors = format!("^{without_anchors}$");
+                match Regex::new(&with_anchors) {
+                    Ok(regex) => Some(regex),
+                    Err(error) => {
+                        tracing::warn!(
+                            regex,
+                            regex_without_anchors = without_anchors,
+                            regex_with_anchors = with_anchors,
+                            ?error,
+                            "Could not parse regex from \"jvmArgumentOverrides.removeRegex\", ignoring it"
```
maltesander, line 261: Im not sure its a good idea to ignore the regex instead of erroring?

Is this conscious decision or due to the trait?

sbernauer: This is mostly because the trait is not fallible. We could not use the Trait, but not sure which is better ^^
Any opinions?

We can also bump warn to error maybe

maltesander: Hmm i think i prefer to error out here (with a Result) instead of deploying a cluster that may or may not do what its supposed to?

sbernauer: Updated the PR to not use the Merge trait any more, but a fallible function

maltesander (approved):
Just nitpicking, LGTM otherwise :)

crates/stackable-operator/src/role_utils.rs
```diff
@@ -296,3 +479,163 @@ impl<K: Resource> Display for RoleGroupRef<K> {
...
+                RoleGroup {
+                    config: CommonConfiguration {
+                        product_specific_common_config: role_group,
+                        ..Default::default()
+                    },
+                    replicas: Some(42),
+                },
+            )]),
+            ..Default::default()
+        };
```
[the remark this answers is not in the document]
maltesander: Nice thanks, i would adapt the second test as well to be consistent? I can do it as well.

maltesander (dismissed):
Thanks, LGTM!

crates/stackable-operator/src/role_utils.rs
```diff
@@ -296,3 +479,118 @@ impl<K: Resource> Display for RoleGroupRef<K> {
...
+    #[test]
+    fn test_merge_java_common_config_keep_order() {
+        let operator_generated =
+            JvmArgumentOverrides::new_with_only_additions(["-Xms1m".to_owned()].into());
+
+        let entire_role: Role<(), GenericRoleConfig, JavaCommonConfig> = serde_yaml::from_str(
+            "
+                jvmArgumentOverrides:
+                    add:
+                      - -Xms2m
```
sbernauer, line 573:
```suggestion
                  add:
                    - -Xms2m
```

maltesander (dismissed):
LGTM!

maltesander (approved):
LGTM!

PR #932, opened 2024-12-18
refactor(stackable-telemetry)!: Rename Tracing to Telemetry

Techassi:
This PR renames `Tracing` to `Telemetry` to more appropriately reflect what the type does: constructing subscribers to process different _telemetry_ signals (currently logs and traces).


NickLarsenNZ (changes requested):
Please wait until I'm back. There's no rush on this yet, and I want to know how metrics fit in to this (from memory it is handled a little different, unless we massage the API a bit)

Techassi:
We'll hold off on this change until @NickLarsenNZ and I can take a look at how metrics will be handled and if they can be incorporated into the `Tracing` / `Telemetry` type.

PR #933, opened 2024-12-19
feat: add RollingFileAppender to stackable-telemetry

xeniape:
# Description

Part of https://github.com/stackabletech/issues/issues/639. Adds the RollingFileAppender to the stackable-telemetry crate.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-telemetry/src/tracing/mod.rs
+++ b/crates/stackable-telemetry/src/tracing/mod.rs
@@ -1,5 +1,5 @@
 //! This module contains functionality to initialise tracing Subscribers for
-//! console output, and OpenTelemetry OTLP export for traces and logs.
+//! console output, file output and OpenTelemetry OTLP export for traces and logs.
 //!
 //! It is intended to be used by the Stackable Data Platform operators and
 //! webhooks, but it should be generic enough to be used in any application.
@@ -16,9 +16,10 @@ use opentelemetry_sdk::{
 use opentelemetry_semantic_conventions::resource;
 use snafu::{ResultExt as _, Snafu};
 use tracing::subscriber::SetGlobalDefaultError;
+use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{filter::Directive, layer::SubscriberExt, EnvFilter, Layer, Registry};
 
-use settings::{ConsoleLogSettings, OtlpLogSettings, OtlpTraceSettings};
+use settings::{ConsoleLogSettings, FileLogSettings, OtlpLogSettings, OtlpTraceSettings};
 
 pub mod settings;
 
@@ -151,6 +152,7 @@ pub enum Error {
 pub struct Tracing {
     service_name: &'static str,
     console_log_settings: ConsoleLogSettings,
+    file_log_settings: FileLogSettings,
     otlp_log_settings: OtlpLogSettings,
     otlp_trace_settings: OtlpTraceSettings,
     logger_provider: Option<LoggerProvider>,
@@ -179,6 +181,29 @@ impl Tracing {
             let console_output_layer =
                 tracing_subscriber::fmt::layer().with_filter(env_filter_layer);
             layers.push(console_output_layer.boxed());
+        }
+
+        if self.file_log_settings.enabled {
+            let env_filter_layer = env_filter_builder(
+                self.file_log_settings.common_settings.environment_variable,
+                self.file_log_settings.default_level,
+            );
+
+            let file_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix(self.service_name.to_string())
+                .filename_suffix("tracing-rs.json")
+                .max_log_files(6)
+                .build(&self.file_log_settings.file_log_dir)
+                .expect("failed to initialize rolling file appender");
+
+            layers.push(
+                tracing_subscriber::fmt::layer()
+                    .json()
+                    .with_writer(file_appender)
+                    .with_filter(env_filter_layer)
+                    .boxed(),
+            );
         }
 
         if self.otlp_log_settings.enabled {
@@ -320,12 +345,6 @@ mod builder_state {
     #[derive(Default)]
     pub struct PreServiceName;
 
-    /// The state before the [`EnvFilter`][1] environment variable name is set.
-    ///
-    /// [1]: tracing_subscriber::filter::EnvFilter
-    #[derive(Default)]
-    pub struct PreEnvVar;
-
     /// The state that allows you to configure the supported [`Subscriber`][1]
     /// [`Layer`][2].
     ///
@@ -349,6 +368,7 @@ pub struct TracingBuilder<S: BuilderState> {
     console_log_settings: ConsoleLogSettings,
     otlp_log_settings: OtlpLogSettings,
     otlp_trace_settings: OtlpTraceSettings,
+    file_log_settings: FileLogSettings,
 
     /// Allow the generic to be used (needed for impls).
     _marker: std::marker::PhantomData<S>,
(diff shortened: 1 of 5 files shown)


crates/stackable-telemetry/CHANGELOG.md
```diff
@@ -7,13 +7,15 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Introduce common `Settings` and subscriber specific settings ([#901]).
+- Added support for logging to files ([#933]).
```
Techassi, line 10: suggestion: Use `Add` instead of `Added`.

```suggestion
- Add support for logging to files ([#933]).
```

note (non-blocking): I usually add newer entries in the changelog at the top. Feel free to move the entry or leave it as is.

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -181,6 +183,29 @@ impl Tracing {
...
+                self.file_log_settings.default_level,
+            );
+
+            let file_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix(self.service_name.to_string())
+                .filename_suffix("tracing-rs.json")
+                .max_log_files(6)
+                .build(&self.file_log_settings.file_log_dir)
+                .expect("failed to initialize rolling file appender");
```
Techassi, line 198: suggestion: We should error here instead of panicking, because this function is fallible compared to before, see:

https://github.com/stackabletech/operator-rs/blob/29e920d13d03112c0000714cbda735b242585f5d/crates/stackable-operator/src/logging/mod.rs#L31

crates/stackable-telemetry/src/tracing/settings/file_log.rs
```diff
@@ -0,0 +1,155 @@
...
+impl BuilderState for builder_state::Config {}
+
+/// For building [`FileLogSettings`].
+///
+/// <div class="warning">
+/// Do not use directly, instead use the [`Settings::builder`] associated function.
+/// </div>
+pub struct FileLogSettingsBuilder<S: BuilderState> {
+    pub(crate) common_settings: Settings,
+    pub(crate) file_log_dir: Option<PathBuf>,
```
Techassi, line 86: question: Why is this optional, if we expect thus value to be always set? See:

https://github.com/stackabletech/operator-rs/blob/4d3ebd08da506f101b2479dea74c534e798f4e12/crates/stackable-telemetry/src/tracing/settings/file_log.rs#L112

In my opinion, this shouldn't be optional or should have a default if the user should be able to build the settings without calling `file_log_settings_builder`.

crates/stackable-telemetry/src/tracing/settings/file_log.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Set the directory for log files.
+    ///
+    /// A directory is required for using the File Log subscriber.
+    pub fn with_file_log_dir(self, path: String) -> FileLogSettingsBuilder<builder_state::Config> {
+        FileLogSettingsBuilder {
+            common_settings: self.common_settings,
+            file_log_dir: Some(PathBuf::from(path)),
+            _marker: std::marker::PhantomData,
+        }
+    }
```
Techassi, line 102: suggestion: We should use an actual path here instead of a string.

```suggestion
    pub fn with_file_log_dir<P>(self, path: P) -> FileLogSettingsBuilder<builder_state::Config> where P: AsRef<Path> {
        FileLogSettingsBuilder {
            common_settings: self.common_settings,
            file_log_dir: path.as_ref().into_path_buf(),
            _marker: std::marker::PhantomData,
        }
    }
```

crates/stackable-telemetry/src/tracing/settings/mod.rs
```diff
@@ -105,6 +108,11 @@ impl SettingsBuilder {
         self.into()
     }
 
+    /// Set specific [`FileLogSettings`].
+    pub fn file_log_settings_builder(self) -> FileLogSettingsBuilder<builder_state::PreLogDir> {
```
Techassi, line 112: note: I think it makes sense to require the user to pass the folder path here as it is required. This then also should get rid of the builder states.

The `with_` functions are intended for optional settings (which override some kind of default value).

Techassi (changes requested):
The core change looks good, but I left a few comments / ideas on how we should design the API interface.

crates/stackable-telemetry/src/tracing/settings/file_log.rs
xeniape: I was under the impression while implementing that the From impl was necessary, which complicated the whole thing and this needed to be optional for that. But after clarifying and refactoring that, it's not optional anymore and in general much simpler :smile:

xeniape:
Thanks for all the feedback! Adjusted everything in https://github.com/stackabletech/operator-rs/pull/933/commits/bfe89a7609ae8092a48c018c33639b7ce6ab27f4

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -510,6 +558,14 @@ mod test {
                     .enabled(true)
                     .build(),
             )
+            .with_file_output(
+                Settings::builder()
+                    .with_environment_variable("ABC_FILE")
+                    .with_default_level(LevelFilter::INFO)
+                    .enabled(true)
+                    .file_log_settings_builder(String::from("/abc_file_dir"))
```
Techassi, line 566: question: While this technically works, can we use `PathBuf` here instead of `String`? This helps to show developers the preferred way of using this API.

crates/stackable-telemetry/src/tracing/settings/file_log.rs
```diff
@@ -0,0 +1,70 @@
...
+                default_level: LevelFilter::DEBUG,
+                enabled: true,
+            },
+            file_log_dir: PathBuf::from("/logs"),
+        };
+        let result = Settings::builder()
+            .with_environment_variable("hello")
+            .with_default_level(LevelFilter::DEBUG)
+            .enabled(true)
+            .file_log_settings_builder(String::from("/logs"))
```
Techassi, line 65: note: Again, same as above.

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -181,6 +183,29 @@ impl Tracing {
             layers.push(console_output_layer.boxed());
         }
 
+        if self.file_log_settings.enabled {
+            let env_filter_layer = env_filter_builder(
+                self.file_log_settings.common_settings.environment_variable,
+                self.file_log_settings.default_level,
+            );
```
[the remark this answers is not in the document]
NickLarsenNZ: I had a quick try of this, but it proved more difficult than I expected.
I'll pick this up separately another time.

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -1,5 +1,5 @@
 //! This module contains functionality to initialise tracing Subscribers for
-//! console output, and OpenTelemetry OTLP export for traces and logs.
+//! console output, file output and OpenTelemetry OTLP export for traces and logs.
```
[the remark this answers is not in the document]
NickLarsenNZ: Done in bfe89a7

crates/stackable-telemetry/src/tracing/mod.rs
NickLarsenNZ:
```suggestion
                    .file_log_settings_builder(PathBuf::from("/abc_file_dir"))
```

crates/stackable-telemetry/src/tracing/settings/file_log.rs
NickLarsenNZ:
```suggestion
            .file_log_settings_builder(PathBuf::from("/logs"))
```

crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -446,6 +469,26 @@ impl TracingBuilder<builder_state::Config> {
...
+    pub fn with_file_output(
+        self,
+        file_log_settings: FileLogSettings,
+    ) -> TracingBuilder<builder_state::Config> {
+        TracingBuilder {
+            service_name: self.service_name,
+            console_log_settings: self.console_log_settings,
+            file_log_settings,
+            otlp_log_settings: self.otlp_log_settings,
+            otlp_trace_settings: self.otlp_trace_settings,
```
Techassi, line 491: note: #940 changed the API interface of these functions and as such I suggest to also change it here to be in line.

```suggestion
    pub fn with_file_output(
        self,
        file_log_settings: impl Into<FileLogSettings>,
    ) -> TracingBuilder<builder_state::Config> {
        TracingBuilder {
            service_name: self.service_name,
            console_log_settings: self.console_log_settings,
            file_log_settings: file_log_settings.into(),
            otlp_log_settings: self.otlp_log_settings,
            otlp_trace_settings: self.otlp_trace_settings,
```

PR #939, opened 2025-01-10
feat: improve DNS lookup performance

dervoeti:
# Description

Fixes https://github.com/stackabletech/issues/issues/656

The current [regex for domain names](https://github.com/stackabletech/operator-rs/blob/feat/dns-performance-fqdns/crates/stackable-operator/src/validation.rs#L20-L22) does not permit a trailing dot. Analogous to the [logic in Kubernetes](https://github.com/kubernetes/kubernetes/blob/30de989fb57fb5921a7ae3e3203cf7ecac9cf3f0/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L100), I decided to remove the trailing dot before validation (if the dot is present).
After validation was successful, a trailing dot is now always appended to make the domain name a FQDN. This improves DNS performance, since for FQDNs the "search" domains in resolv.conf are not considered.

# How to test this

### General test setup
Local Kind cluster, add the line `log` (after `errors`) to the CoreDNS ConfigMap to make CoreDNS log all DNS queries, scale CoreDNS Deployment to 1 so we can get all the logs from one CoreDNS Pod, restart the CoreDNS Deployment to reload the config.

Install some operators:
```
stackablectl operator install commons secret listener zookeeper
```

Setup a Zookeeper cluster and a ZNode, wait for all pods to start, check CoreDNS logs

I grepped through the CoreDNS logs for `svc.cluster.local.cluster.local`, which indicates a "search" domain was appended to the DNS query (which degrades performance).

### Using the current dev release, ndots=5 (default value)
Create a ZK cluster with 3 replicas:
```
apiVersion: zookeeper.stackable.tech/v1alpha1
kind: ZookeeperCluster
metadata:
  name: simple-zk
  namespace: default
spec:
  image:
    productVersion: 3.9.2
  servers:
    roleGroups:
      default:
        replicas: 3
```

Create a dummy ZNode:
```
apiVersion: zookeeper.stackable.tech/v1alpha1
kind: ZookeeperZnode
metadata:
  name: dummy-znode
spec:
  clusterRef:
    name: simple-zk
```

Lines like this should appear in the CoreDNS log:
```
[INFO] 10.244.0.45:35325 - 7328 "A IN simple-zk.default.svc.cluster.local.cluster.local. udp 67 false 512" NXDOMAIN qr,aa,rd 160 0.000192228s
[INFO] 10.244.0.45:35325 - 31385 "AAAA IN simple-zk.default.svc.cluster.local.cluster.local. udp 67 false 512" NXDOMAIN qr,aa,rd 160 0.000258389s
[INFO] 10.244.0.45:33465 - 56111 "A IN simple-zk.default.svc.cluster.local.cluster.local. udp 67 false 512" NXDOMAIN qr,aa,rd 160 0.000028455s
[INFO] 10.244.0.45:33465 - 5921 "AAAA IN simple-zk.default.svc.cluster.local.cluster.local. udp 67 false 512" NXDOMAIN qr,aa,rd 160 0.00008136s
```

The lines around these matches show how all search domains are tried and only the final query (for `simple-zk.default.svc.cluster.local.`) succeeds.

### Using the current dev release, PodOverrides with ndots=3
Patch the Zookeeper Operator deployment:
```
kubectl patch deployment -n stackable-operators zookeeper-operator-deployment -p '{"spec":{"template":{"spec":{"dnsConfig":{"options":[{"name":"ndots","value":"3"}]}}}}}'
```
Clean up old ZK resources:
```
kubectl delete zookeeperclusters.zookeeper.stackable.tech simple-zk
kubectl delete pvc data-simple-zk-server-default-{0,1,2}
```

Restart CoreDNS deployment to clear the logs:
```
kubectl rollout restart deployment -n kube-system coredns
```

Recreate ZK cluster with ndots=3:
```
apiVersion: zookeeper.stackable.tech/v1alpha1
kind: ZookeeperCluster
metadata:
  name: simple-zk
  namespace: default
spec:
  image:
    productVersion: 3.9.2
  servers:
    roleGroups:
      default:
        podOverrides:
          dnsConfig:
            options:
              - name: ndots
                value: "3"
        replicas: 3
```

Wait for Pods to start, check CoreDNS log. `svc.cluster.local.cluster.local` should not be present in the log.

### Using this fix, no PodOverrides
Clean up old ZK resources:
```
kubectl delete zookeeperclusters.zookeeper.stackable.tech simple-zk
kubectl delete pvc data-simple-zk-server-default-{0,1,2}
```

Uninstall Zookeeper operator:
```
stackablectl operator uninstall zookeeper
```

Install the Zookeeper operator with the fix from this PR (`make run-dev`).

Restart CoreDNS deployment to clear the logs:
```
kubectl rollout restart deployment -n kube-system coredns
```

Create a normal ZK cluster without the ndots override:
```
apiVersion: zookeeper.stackable.tech/v1alpha1
kind: ZookeeperCluster
metadata:
  name: simple-zk
  namespace: default
spec:
  image:
    productVersion: 3.9.2
  servers:
    roleGroups:
      default:
        replicas: 3
```

Wait for Pods to start, check CoreDNS log. `svc.cluster.local.cluster.local` should not be present in the log.

I additionally tested the fix with ndots set to 7, it still works even though the FQDN contains only 5 dots. I think that's because the search domains are never used when you make a DNS query for an FQDN (with a trailing dot).

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


maltesander:
Tested and works!

Will wait with approval for now until:
- we decided on where to add the changes e.g. DomainName vs FqdnDomainName etc.
- we tested with operator PRs, integrationtests and demos

crates/stackable-operator/src/utils/cluster_info.rs
```diff
@@ -24,16 +24,19 @@ impl KubernetesClusterInfo {
     pub fn new(cluster_info_opts: &KubernetesClusterInfoOpts) -> Self {
         let cluster_domain = match &cluster_info_opts.kubernetes_cluster_domain {
             Some(cluster_domain) => {
+                // Append trailing dot to make the domain name a FQDN to improve DNS lookup performance, see https://github.com/stackabletech/issues/issues/656
+                let cluster_domain = DomainName::from_str(&format!("{}.", cluster_domain))
+                    .expect("cluster_domain should already have been validated");
```
maltesander, line 29: If i provide `foo.bar.` this would fail when reading and verifying the `DomainName`?

maltesander (commented):
LGTM overall!
Changelog entry is missing.
Do we want to mention anything about DNS/ndots etc. here https://docs.stackable.tech/home/stable/kubernetes/#_configuring_the_cluster_domain (or the sublinks)?

sbernauer:
Thanks for working on this!

Actually I now have enough from the DomainName using the RFC_1123_SUBDOMAIN_REGEX :( I would suggest to add a new FQDN_REGEX allowing a trailing dot (as foo. is a valid domain name!) and using that, WDYT?

Regarding the user input I would prefer to not fiddle with the input the user gave us, but document and warn accordingly. Maybe our dotting strategy isn't perfect and the dot actually hurts something? Currently we would have no way of disabling it. But no strong opinion.

I spiked both things in https://github.com/stackabletech/operator-rs/compare/main...feat/dns-performance-fqdns-review, what are your opinions on this?

crates/stackable-operator/src/utils/cluster_info.rs
sbernauer: Yeah, we should probably only append a dot in case it does not already end with a dot

dervoeti:
@sbernauer I like the idea of removing the logic from the `new()` function of `KubernetesClusterInfo` and printing a warning if a trailing dot is not present. That would basically reduce this PR to just appending a dot to `KUBERNETES_CLUSTER_DOMAIN_DEFAULT` :melting_face:
But I think this way it's less error prone **and** we have a way for users to opt out of this behavior: just set `KUBERNETES_CLUSTER_DOMAIN` to `cluster.local`.

Regarding the FQDN regex: We also thought about this, I don't think it's a bad idea but we decided to do it differently to be in line with the way Kubernetes does it: https://github.com/kubernetes/kubernetes/blob/30de989fb57fb5921a7ae3e3203cf7ecac9cf3f0/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L100
But I'm fine with your solution as well. WDYT @maltesander?

crates/stackable-operator/src/utils/cluster_info.rs
dervoeti: Oh, right, that's also the one case I forgot to add to the tests :/
See comments below, if we do it like @sbernauer suggested this is obsolete.
Otherwise we need to fix this (check if trailing dot is already present or always remove it before adding it).

sbernauer:
> Regarding the FQDN regex

Works for me, we can use your existing logic.

> But I think this way it's less error prone and we have a way for users to opt out of this behavior: just set KUBERNETES_CLUSTER_DOMAIN to cluster.local.

I don't understand how they can opt out to be honest. I *think* currently they are unable to set it to `cluster.local`, correct?

maltesander:
To summarize, we can:

1) Just "pass-through" the cluster domain, do not append anything
	+ users can opt out of the dot suffix
	+ default behavior is the same
	- no default performance increase
2) Suffix cluster domain (and default) with dot
    + performance increase by default
    - users cannot opt out
3) Use own FQDN_REGEX 
    + No "." strip / trim / add required, all covered 
    + Users can opt out
    - Would influence everything (e.g. keycloak domain) which is hard to test
      (there probably is some library out there that cannot deal with that...)
    - Even though it seems like a workaround, i like sticking to the implementation
      that kubernetes is doing

Is that correct? I honestly think option 1 is my favortite but this comes at the cost of not having the performance increase by default. WDYT?

crates/stackable-operator/src/validation.rs
```diff
@@ -15,18 +15,23 @@ use const_format::concatcp;
...
     concatcp!(RFC_1123_LABEL_FMT, "(\\.", RFC_1123_LABEL_FMT, ")*");
 const RFC_1123_SUBDOMAIN_ERROR_MSG: &str = "a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character";
-const RFC_1123_LABEL_ERROR_MSG: &str = "a lowercase RFC 1123 label must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character";
 
-// This is a subdomain's max length in DNS (RFC 1123)
-const RFC_1123_SUBDOMAIN_MAX_LENGTH: usize = 253;
-// Minimal length required by RFC 1123 is 63. Up to 255 allowed, unsupported by k8s.
-const RFC_1123_LABEL_MAX_LENGTH: usize = 63;
+const FQDN_MAX_LENGTH: usize = RFC_1123_SUBDOMAIN_MAX_LENGTH;
+/// Same as [`RFC_1123_SUBDOMAIN_FMT`], but allows a trailing slash
```
sbernauer, line 32:
```suggestion
/// Same as [`RFC_1123_SUBDOMAIN_FMT`], but allows a trailing dot
```

sbernauer (dismissed):
Thanks! Code looks good, we have a opt out as well :)

Please add a changelog and make some changes to the https://docs.stackable.tech/home/stable/guides/kubernetes-cluster-domain/ guire

dervoeti:
@maltesander @sbernauer

Updates:
 - Changelog adapted
 - [Docs written](https://github.com/stackabletech/documentation/pull/698)
 - Added explicit tests for a trailing dot in FQDNs
 - I tested different scenarios with Zookeeper and HDFS, including specifying a custom cluster domain (with and without trailing dot). Everything worked as expected, it was possible to opt out via setting `KUBERNETES_CLUSTER_DOMAIN` to `cluster.local` (a warning was printed in that case).

maltesander (commented):
Nice thanks! Checked docs as well.
LGTM. Lets wait for @sbernauer though.

sbernauer (dismissed):
LGTM, thanks!

crates/stackable-operator/src/validation.rs
```diff
@@ -15,18 +15,23 @@ use const_format::concatcp;
...
 const RFC_1123_SUBDOMAIN_ERROR_MSG: &str = "a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character";
-const RFC_1123_LABEL_ERROR_MSG: &str = "a lowercase RFC 1123 label must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character";
 
-// This is a subdomain's max length in DNS (RFC 1123)
-const RFC_1123_SUBDOMAIN_MAX_LENGTH: usize = 253;
-// Minimal length required by RFC 1123 is 63. Up to 255 allowed, unsupported by k8s.
-const RFC_1123_LABEL_MAX_LENGTH: usize = 63;
+const FQDN_MAX_LENGTH: usize = RFC_1123_SUBDOMAIN_MAX_LENGTH;
+/// Same as [`RFC_1123_SUBDOMAIN_FMT`], but allows a trailing dot
+const FQDN_FMT: &str = concatcp!(RFC_1123_SUBDOMAIN_FMT, "\\.?");
```
nightkr, line 33: We shouldn't call "maybe a FQDN, it's in the eye of the beholder" a FQDN. If it doesn't include the dot, we don't know that it's a FQDN.

[RFC 1123](https://www.rfc-editor.org/rfc/rfc1123) doesn't actually mention subdomains at all. [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035.html#page-8) doesn't define FQDN, but defines subdomain as follows:

```ebnf
<subdomain> ::= <label> | <subdomain> "." <label>
```

Starting from that, I would suggest the following terminology:

```ebnf
<fqdn> ::= <subdomain> "."
<domain> ::= <subdomain> | <fqdn>
```

crates/stackable-operator/CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ All notable changes to this project will be documented in this file.
 
 - BREAKING: Bump Rust dependencies to enable Kubernetes 1.32 (via `kube` 0.98.0 and `k8s-openapi`
   0.23.0) ([#867]).
+- BREAKING: Append a dot to the default cluster domain and allow a trailing dot in `DomainName` ([#939]).
```
nightkr, line 23: This should also reflect what it means semantically (the FQDNs), not just the changed tokens.

nightkr (changes requested):

crates/stackable-operator/CHANGELOG.md
dervoeti: Yep, done!

crates/stackable-operator/src/validation.rs
dervoeti: You're right, I now replaced "fqdn" with "domain" in the changed code. We currently don't care about FQDNs explicitly when validating domain names, we just want to additionally allow them.

I was also confused about the RFC references, you're right about RFC 1123 not mentioning subdomains.
I guess these parts of the code were ported from the [Kubernetes source](https://github.com/kubernetes/kubernetes/blob/30de989fb57fb5921a7ae3e3203cf7ecac9cf3f0/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L218-L219), which also references RFC 1123. But to me it looks like RFC 1035 would be the correct one to refer to. I didn't want to change it though.

PR #947, opened 2025-01-21
fix: Revert default cluster domain to 'cluster.local'

dervoeti:
# Description

Revert the change of the default cluster domain introduced by https://github.com/stackabletech/operator-rs/pull/939.

Reasoning: We came across some issues with TLS certificates and Kerberos keytabs and decided to make this behavior (FQDN domains for services/pods) opt-in for now. We still wanted to keep experimental support for it, but only if a user opts in by explicitly setting `KUBERNETES_CLUSTER_DOMAIN` to a domain with a trailing dot.

I tested this change with Zookeeper, both with and without `KUBERNETES_CLUSTER_DOMAIN` set, works as expected.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/src/utils/cluster_info.rs
+++ b/crates/stackable-operator/src/utils/cluster_info.rs
@@ -2,18 +2,18 @@ use std::str::FromStr;
 
 use crate::commons::networking::DomainName;
 
-const KUBERNETES_CLUSTER_DOMAIN_DEFAULT: &str = "cluster.local.";
+const KUBERNETES_CLUSTER_DOMAIN_DEFAULT: &str = "cluster.local";
 
 /// Some information that we know about the Kubernetes cluster.
 #[derive(Debug, Clone)]
 pub struct KubernetesClusterInfo {
-    /// The Kubernetes cluster domain, typically `cluster.local.`.
+    /// The Kubernetes cluster domain, typically `cluster.local`.
     pub cluster_domain: DomainName,
 }
 
 #[derive(clap::Parser, Debug, Default, PartialEq, Eq)]
 pub struct KubernetesClusterInfoOpts {
-    /// Kubernetes cluster domain, usually this is `cluster.local.`.
+    /// Kubernetes cluster domain, usually this is `cluster.local`.
     ///
     /// Please note that we recommend adding a trailing dot (".") to reduce DNS requests, see
     /// <https://github.com/stackabletech/issues/issues/656> for details.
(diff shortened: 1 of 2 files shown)


crates/stackable-operator/src/utils/cluster_info.rs
```diff
@@ -2,18 +2,18 @@ use std::str::FromStr;
...
     pub cluster_domain: DomainName,
 }
 
 #[derive(clap::Parser, Debug, Default, PartialEq, Eq)]
 pub struct KubernetesClusterInfoOpts {
-    /// Kubernetes cluster domain, usually this is `cluster.local.`.
+    /// Kubernetes cluster domain, usually this is `cluster.local`.
     ///
     /// Please note that we recommend adding a trailing dot (".") to reduce DNS requests, see
     /// <https://github.com/stackabletech/issues/issues/656> for details.
```
maltesander, line 19: I think the comment below is not a recommendation anymore as long as we want to mark it as experimental?

dervoeti: Ah, that's right. And we should probably also remove the warning printed by the operator if you specify a custom cluster domain without a trailing dot. I'll fix this.

crates/stackable-operator/src/utils/cluster_info.rs
```diff
@@ -29,9 +25,6 @@ impl KubernetesClusterInfo {
         let cluster_domain = match &cluster_info_opts.kubernetes_cluster_domain {
             Some(cluster_domain) => {
                 tracing::info!(%cluster_domain, "Using configured Kubernetes cluster domain");
-                if !cluster_domain.ends_with('.') {
-                    tracing::warn!(%cluster_domain, "Your configured Kubernetes cluster domain is not fully qualified (it does not end with a dot (\".\")). We recommend adding a trailing dot to reduce DNS requests, see https://github.com/stackabletech/issues/issues/656 for details");
-                }
```
nightkr, line 34: Assuming that we still want to move in this direction _eventually_, do we want to keep this message around in some form? (Maybe commented, maybe gated on a constant, etc)

dervoeti: Not sure, I think I'd remove it for now, we have the code in Git if we need it. For users, my plan was to [add a hint to the docs](https://github.com/stackabletech/documentation/blob/9527b94e584de5e753a28d4b86e6da496d59ba14/modules/guides/pages/kubernetes-cluster-domain.adoc?plain=1#L15).

maltesander (approved):
LGTM!

PR #946, opened 2025-01-20
Remove merge trait and make product_specific_common_config public

maltesander:
# Description

- Remove merge trait from erase()
- make product_specific_common_config public

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

```[tasklist]
# Author
- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

PR #948, opened 2025-01-22
chore: Release stackable-operator 0.84.1

xeniape:
This releases `stackbale-operator` 0.84.1 which includes:

### Fixed

- Remove `Merge` trait bound from `erase` and make `product_specific_common_config` public ([#946]).
- BREAKING: Revert the change of appending a dot to the default cluster domain to make it a FQDN, it is now `cluster.local` again. Users can instead explicitly opt-in to FQDNs via the ENV variable `KUBERNETES_CLUSTER_DOMAIN`. ([#947]).

[#946]: https://github.com/stackabletech/operator-rs/pull/946
[#947]: https://github.com/stackabletech/operator-rs/pull/947

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3199,7 +3199,7 @@ dependencies = [
 
 [[package]]
 name = "stackable-operator"
-version = "0.84.0"
+version = "0.85.0"
 dependencies = [
  "chrono",
  "clap",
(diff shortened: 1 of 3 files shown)


sbernauer (dismissed):
Ideally this would be 0.84.1, but not blocking it on it

sbernauer (approved):
Thanks!

xeniape:
> Ideally this would be 0.84.1, but not blocking it on it

You're absolutely right. Didn't think about it too much but the changes only include fixes :) adjusted it, thanks!

PR #951, opened 2025-01-28
refactor(stackable-telemetry): Support Option<T> in subscriber methods

Techassi:
Part of https://github.com/stackabletech/issues/issues/639

As requested in https://github.com/stackabletech/operator-rs/pull/901#discussion_r1860798808, this PR enables this:

```rust
.with_otlp_log_exporter(
  otlp_log_flag.then(|| Settings::builder()
    .with_environment_variable("ABC_OTLP_LOG")
    .with_default_level(LevelFilter::INFO)
    .build()
  )
)
```


crates/stackable-telemetry/CHANGELOG.md
```diff
@@ -6,16 +6,23 @@ All notable changes to this project will be documented in this file.
...
 - Add support for logging to files ([#933]).
 
 ### Changed
 
-- BREAKING: Renamed `TracingBuilder` methods with long names, and prefix with `with_` ([#901]).
+- BREAKING: Change subscriber settings into an enum to indicate if the subscriber is enabled/disabled ([#951]).
+- BREAKING: Rename `TracingBuilder` methods with long names, and prefix with `with_` ([#901]).
 - BREAKING: Use the new subscriber settings in the `TracingBuilder` ([#901]).
 
+## Removed
```
Techassi, line 19:
```suggestion
### Removed
```

crates/stackable-telemetry/src/tracing/settings/mod.rs
```diff
@@ -16,6 +16,11 @@ pub use otlp_log::*;
 pub mod otlp_trace;
 pub use otlp_trace::*;
 
+pub trait SettingsToggle {
+    fn is_enabled(&self) -> bool;
+    fn is_disabled(&self) -> bool;
```
nightkr, line 21: Do we really need both of these?

crates/stackable-telemetry/src/tracing/settings/console_log.rs
```diff
@@ -1,27 +1,22 @@
...
 /// Configure specific settings for the Console Log subscriber.
 #[derive(Debug, Default, PartialEq)]
-pub struct ConsoleLogSettings {
-    /// Common subscriber settings that apply to the Console Log Subscriber.
-    pub common_settings: Settings,
-
-    /// Console Subscriber log event output format.
-    pub log_format: Format,
-}
+pub enum ConsoleLogSettings {
```
nightkr, line 9: Any reason to implement this toggle separately for each variant, instead of just reusing `Option<ConsoleLogSettings>`?

crates/stackable-telemetry/src/tracing/settings/mod.rs
NickLarsenNZ: technically no, but it does make the `assert!` look nicer (instead of `assert!(!...is_enabled())`.

Technically none of it is required, but it gets uglier to check the variant.

NickLarsenNZ: What I will do is make is a default impl though.

nightkr: Disagree on the nicer, but we could at least provide `is_disabled()` as a default impl (especially since the concrete impls just delegate to it anyway).

nightkr: Ah, didn't see the followup message. :P

crates/stackable-telemetry/src/tracing/settings/console_log.rs
NickLarsenNZ: Yes, because of impls we needed:

```rust
impl<T> From<Option<T>> for BlahSettings
where
    T: Into<BlahSettings>,
{ ... }
```

NickLarsenNZ: Or did you mean providing something like:

```rust
pub struct GenericSettings<T> {
  Disabled,
  Enabled<T>,
}

impl<T, U> From<Option<T>> for GenericSettings<U>
where
    T: Into<U>,
{ ... }
```

nightkr: Ideally I'd avoid the `From` entirely; `with_file_output` takes an `Option<FileLogSettings>` and that's that. But I understand if that's too extreme, or if that ship has sailed already...

As a compromise, we could define our own trait with a blanket impl:

```rs
trait IntoLogSettings<S> {
    fn into_log_settings(self) -> Option<S>;
}
impl<T: IntoLogSettings, S> IntoLogSettings<S> for Option<T> {
    fn into_log_settings(self) -> Option<S> {
        self.and_then(|x| x.into_log_settings())
    }
}
```

NickLarsenNZ: I didn't completely follow what you meant, but I quite like having the `From` so we can cater to a few different styles, so dropping that might undo some of the intended ergonomics.

Is it ok to merge as is, or would you like to append a commit to show what you meant?

nightkr: There are really a few separate concerns here:

1. I dislike how `.into()` becomes this.. "okay, magic happens here to make it work" placeholder that doesn't really _say_ anything about the intention of the code. Sometimes that's fine, but it depends on how obvious that conversion is.
2. `From`/`Into` composes poorly with standard types (like you already noticed around the orphan rules and `Option`)
3. We have a bad habit of overusing `From`/`Into` in contexts where their meaning is really not clear from the context. For example: tell me what `impl From<(&'static str, LevelFilter, bool)> for OtlpLogSettings` means without reading its source code. Is the `str` an environment variable? A target spec? A URL? This shouldn't be an implicit conversion at all, it should be a named and documented method like `Settings::from_env()`.

> Is it ok to merge as is, or would you like to append a commit to show what you meant?

Ignoring 3 for a moment (that's an older thing that isn't really in scope for this PR), I'm envisioning something like https://gist.github.com/nightkr/364ccc31ece663158a4c7e448be11065. I think I'm okay with merging the PR as is for now, and revisiting these problems later.

NickLarsenNZ: 1. I can understand the magic argument, but this is where the function/argument name should be giving enough context. If it can convert, then it is fine.

2. You can say that about any useful external trait.

3. I'm not sure that meaning needs to be clear to be able to use the traits. From/Into are simply conversion traits. As for the example, correct, you cannot tell what it is used for - but then you have to ask, how did you find that in the first place? If it is in the docs, then there should be docs near it (for this case I prefer to type alias so you can actually put docs on the tuple). The tuple is not good on its own, but with the conversion trait, I think it is fine - it doesn't force anyone to use it but leaves it there for people who do understand it and want to use it (remaining flexible of end-developers styles).

I will merge as is, and I'll keep your considerations in mind for future changes.

PR #956, opened 2025-02-03
chore: Add rustfmt config

Techassi:
This PR adds a rustfmt config file and also applies the resulting reformatting. In addition, the pre-commit config now uses a local hook to check formatting of Rust code is correct.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,7 @@ repos:
       - id: rustfmt
         name: rustfmt
         language: system
+        # Pinning to a specific rustc version, so that we get consistent formatting
         entry: cargo +nightly-2025-01-15 fmt --all -- --check
         stages: [pre-commit]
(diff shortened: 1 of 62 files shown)


.pre-commit-config.yaml
```diff
@@ -53,3 +51,10 @@ repos:
         entry: .scripts/verify_crate_versions.sh
         stages: [pre-commit, pre-merge-commit, manual]
         pass_filenames: false
+
+      - id: rustfmt
+        name: rustfmt
+        language: system
+        entry: cargo +nightly-2025-01-15 fmt --all -- --check
```
sbernauer, line 58:
```suggestion
	# Pinning to a specific rustc version, so that we get consistent formatting
        entry: cargo +nightly-2025-01-15 fmt --all -- --check
```

sbernauer (approved):

sbernauer (approved):

PR #967, opened 2025-02-24
feat(stackable-versioned): Add convert_with arg in changed action

Techassi:
This PR adds a new argument `convert_with` to the `changed()` action. It allows users to customize the conversion function used on a field level. This is useful when auto-generated `From` impl between versions should be kept, but individual fields need custom conversions instead of using the default `.into()` call.

**Example**

```rust
#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1beta1")
)]
pub struct Foo {
    #[versioned(changed(
        since = "v1beta1",
        from_type = "u8",
        convert_with = "u8_to_u16"
    ))]
    bar: u16,
}

fn u8_to_u16(old: u8) -> u16 {
    old as u16
}
```

### DIFF
--- a/crates/stackable-versioned-macros/src/codegen/item/field.rs
+++ b/crates/stackable-versioned-macros/src/codegen/item/field.rs
@@ -167,18 +167,29 @@ impl VersionedField {
                         _,
                         ItemStatus::Change {
                             from_ident: old_field_ident,
+                            convert_with,
                             to_ident,
                             ..
                         },
-                    ) => {
-                        quote! {
+                    ) => match convert_with {
+                        // The user specified a custom conversion function which
+                        // will be used here instead of the default .into() call
+                        // which utilizes From impls.
+                        Some(convert_fn) => quote! {
+                            #to_ident: #convert_fn(#from_struct_ident.#old_field_ident),
+                        },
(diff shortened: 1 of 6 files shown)


crates/stackable-versioned-macros/src/codegen/item/field.rs
```diff
@@ -167,18 +167,29 @@ impl VersionedField {
...
-                    }
+                        },
+                    },
                     (old, next) => {
                         let next_field_ident = next.get_ident();
                         let old_field_ident = old.get_ident();
 
+                        // NOTE (@Techassi): Do we really need .into() here. I'm
+                        // currently not sure why it is there and if it is needed
+                        // in some edge cases.
```
NickLarsenNZ, line 192: I believe it was left there for consistency with the other calls.

sbernauer (approved):
Very nice quality of life feature!

crates/stackable-versioned-macros/src/codegen/item/field.rs
Techassi: Alright, I will leave the comment in for now. We can revisit this in the future and remove it or the `.into()` call if we desire.

PR #971, opened 2025-02-27
feat(stackable-versioned): Re-emit and merge modules in versioned modules

Techassi:

This PR allows users to define modules in the `versioned` module, which allows composition of imports which then produces easier to consume imports in downstream code.

**Example**

```rust
// In crate/crd/mod.rs

#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1")
)]
mod versioned {
    mod v1alpha1 {
        pub use crate::crd::oidc::v1alpha1::*
        pub use crate::crd::tls::v1alpha1::*
    }

    struct Foo {
        bar: usize,
    }
}
```

### DIFF
--- a/crates/stackable-versioned-macros/fixtures/inputs/default/submodule.rs
+++ b/crates/stackable-versioned-macros/fixtures/inputs/default/submodule.rs
@@ -1,0 +1,11 @@
+#[versioned(version(name = "v1alpha1"), version(name = "v1"))]
+// ---
+mod versioned {
+    mod v1alpha1 {
+        pub use my::reexport::v1alpha1::*;
+    }
+
+    struct Foo {
+        bar: usize,
+    }
+}
--- a/crates/stackable-versioned-macros/src/codegen/module.rs
+++ b/crates/stackable-versioned-macros/src/codegen/module.rs
@@ -1,53 +1,114 @@
 use std::{collections::HashMap, ops::Not};
 
-use darling::util::IdentString;
+use darling::{util::IdentString, Error, Result};
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{token::Pub, Ident, Visibility};
+use syn::{token::Pub, Ident, Item, ItemMod, ItemUse, Visibility};
 
-use crate::codegen::{container::Container, VersionDefinition};
+use crate::{
+    codegen::{container::Container, VersionDefinition},
+    ModuleAttributes,
+};
 
 pub(crate) type KubernetesItems = (Vec<TokenStream>, Vec<IdentString>, Vec<String>);
-
-pub(crate) struct ModuleInput {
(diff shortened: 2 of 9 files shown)


crates/stackable-versioned-macros/src/codegen/module.rs
```diff
@@ -1,53 +1,114 @@
...
+                }
+                Item::Mod(item_module) => match item_module.content {
+                    Some((_, items)) => {
+                        let imports: Vec<ItemUse> = items
+                            .into_iter()
+                            // We are only interested in use statements. Everything else is ignored.
+                            // NOTE (@Techassi): We might want to error instead.
+                            .filter_map(|item| match item {
+                                Item::Use(item_use) => Some(item_use),
+                                _ => None,
```
NickLarsenNZ, line 80: I think this should be an error, suggesting they do that outside.

crates/stackable-versioned-macros/src/codegen/module.rs
```diff
@@ -1,53 +1,114 @@
...
+                            })
+                            .collect();
+
+                        // NOTE (@Techassi): Do we wanna enforce that modules must use version names?
```
NickLarsenNZ, line 84: I think we should either allow any module names and not have magic, or error out when the magic way isn't used as intended.

There should be no good reason for a user to add an extra module inside, so this is purely sugar to help with imports.

NickLarsenNZ (approved):
LGTM. I'll leave it up to you if you introduce the error handling now or later (which will technically be a breaking change).

crates/stackable-versioned-macros/fixtures/inputs/default/submodule.rs
```diff
@@ -0,0 +1,11 @@
...
+// ---
+mod versioned {
+    mod v1alpha1 {
+        pub use my::reexport::v1alpha1::*;
+    }
+
+    struct Foo {
+        bar: usize,
+    }
+}
```
NickLarsenNZ, line 11: Just having some fun, inventing our own syntax:

```suggestion
#[versioned(version(name = "v1alpha1"), version(name = "v1"))]
// ---
mod versioned {
    imports for v1alpha1 {
        my::reexport::v1alpha1::*;
    }

    struct Foo {
        bar: usize,
    }
}
```

_You can completely ignore this_

crates/stackable-versioned-macros/src/codegen/module.rs
Techassi: Done in 12b9883

crates/stackable-versioned-macros/src/codegen/module.rs
Techassi: Done in 12b9883

NickLarsenNZ (approved):
LGTM

PR #973, opened 2025-02-28
chore: Bump Rust version and workflow actions

Techassi:
Part of https://github.com/stackabletech/operator-rs/issues/970

This PR:

- Bumps to Rust toolchain to 1.84.1
- Bumps various actions used in workflows to their newest version
- Hardens the workflows by applying suggestions from zizmor


.github/workflows/publish-docs.yml
```diff
@@ -15,12 +15,16 @@ on:
 env:
   RUST_TOOLCHAIN_VERSION: "1.82.0"
```
NickLarsenNZ, line 16: Should this also be bumped?

```suggestion
  RUST_TOOLCHAIN_VERSION: "1.84.1"
```

NickLarsenNZ (approved):
LGTM, just one question about the Rust toolchain version.

.github/workflows/publish-docs.yml
Techassi: Good spotting! I will update the issue template to include this.

NickLarsenNZ (approved):
LGTM

PR #976, opened 2025-03-06
refactor: Adjust S3 region field structure

Techassi:
This PR adjusts the region field according to this https://github.com/stackabletech/decisions/issues/46#issuecomment-2703905958

```yaml
region:
  # defaults to us-east-1 to be a non-breaking change
  name: eu-central-1
```

### DIFF
--- a/crates/stackable-operator/src/commons/s3/crd.rs
+++ b/crates/stackable-operator/src/commons/s3/crd.rs
@@ -97,12 +97,20 @@ pub enum S3AccessStyle {
     VirtualHosted,
 }
 
-/// Set a named AWS region, or defer to an auto-discovery mechanism.
-#[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+/// Set a named AWS region.
+#[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Region {
     #[serde(default = "default_region_name")]
     name: String,
+}
+
+impl Default for Region {
+    fn default() -> Self {
+        Self {
+            name: default_region_name(),
+        }
+    }
 }
 
 fn default_region_name() -> String {
--- a/crates/stackable-operator/src/commons/s3/crd.rs
+++ b/crates/stackable-operator/src/commons/s3/crd.rs
@@ -64,14 +64,8 @@ pub struct S3ConnectionSpec {
     /// NOTE: This is not the bucket region, and is used by the AWS SDK to
     /// construct endpoints for various AWS service APIs. It is only useful when
     /// using AWS S3 buckets.
-    ///
-    /// When using AWS S3 buckets, you can configure optimal AWS service API
-    /// connections in the following ways:
-    /// - From **inside** AWS: Use an auto-discovery source (eg: AWS IMDS).
-    /// - From **outside** AWS, or when IMDS is disabled, explicity set the
-    ///   region name nearest to where the client application is running from.
     #[serde(default)]
-    pub region: AwsRegion,
+    pub region: Region,
 
     /// Which access style to use.
     /// Defaults to virtual hosted-style as most of the data products out there.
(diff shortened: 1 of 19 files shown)


crates/stackable-operator/src/commons/s3/crd.rs
```diff
@@ -104,55 +98,13 @@ pub enum S3AccessStyle {
 }
 
 /// Set a named AWS region, or defer to an auto-discovery mechanism.
-#[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
```
sbernauer, line 101: Can you please provide the new `region` CRD snippet? I'm not sure if the defaults are set correctly

Techassi: Your suspicion was right. This indeed does not correctly set the default.

```yaml
region:
  default:
    name: ''
  description: |-
    AWS service API region used by the AWS SDK when using AWS S3 buckets.

    This defaults to `us-east-1` and can be ignored if not using AWS S3 buckets.

    NOTE: This is not the bucket region, and is used by the AWS SDK to construct endpoints for various AWS service APIs. It is only useful when using AWS S3 buckets.
  properties:
    name:
      default: us-east-1
      type: string
  type: object
```

Is there a way to forward this as expected using `schemars`?

Techassi: Nevermind, I was looking at this from the wrong angle. It is fixed in 7627c27.

sbernauer: Thanks!

Techassi:
The CI now also doesn't pass because of RUSTSEC-2025-0009. I already fixed in a different PR: [#977 (commit)](https://github.com/stackabletech/operator-rs/pull/977/commits/45a9707a959dbabb20a9df368efd2e815f144b6f). We can pull this into here instead, because this PR also includes the various changes needed to fix the clippy errors.

crates/stackable-operator/src/commons/s3/crd.rs
```diff
@@ -64,14 +64,8 @@ pub struct S3ConnectionSpec {
...
     /// using AWS S3 buckets.
-    ///
-    /// When using AWS S3 buckets, you can configure optimal AWS service API
-    /// connections in the following ways:
-    /// - From **inside** AWS: Use an auto-discovery source (eg: AWS IMDS).
-    /// - From **outside** AWS, or when IMDS is disabled, explicity set the
-    ///   region name nearest to where the client application is running from.
     #[serde(default)]
-    pub region: AwsRegion,
+    pub region: Region,
```
NickLarsenNZ, line 68: Note: comments need updating since region means everything (both client and bucket region) to AWS :/

crates/stackable-operator/src/commons/s3/crd.rs
```diff
@@ -97,14 +97,22 @@ pub enum S3AccessStyle {
     VirtualHosted,
 }
 
-/// Set a named AWS region, or defer to an auto-discovery mechanism.
-#[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+/// Set a named AWS region.
```
NickLarsenNZ, line 100: This could mean either AWS client _or_ bucket region

NickLarsenNZ (changes requested):
LGM, just some comments need updating now that we know more

crates/stackable-operator/src/commons/s3/crd.rs
NickLarsenNZ: In the context of S3, this will primarily mean the bucket region (the client region uses the same setting, but the bucket regions MUST be set correctly for valid sigv4 headers).

crates/stackable-operator/src/commons/s3/crd.rs
NickLarsenNZ: In the context of S3, this will primarily mean the bucket region (the client region uses the same setting, but the bucket regions MUST be set correctly for valid sigv4 headers).

crates/stackable-operator/src/commons/s3/crd.rs
NickLarsenNZ: Done in 6701b6b720408144b761269fed8eb1cd77ff30d8

NickLarsenNZ (approved):
LGTM

PR #983, opened 2025-03-14
feat: Add Region::is_default_config function

sbernauer:
# Description

Needed for https://github.com/stackabletech/druid-operator/pull/696#issuecomment-2724018547

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-operator/src/commons/s3/crd.rs
+++ b/crates/stackable-operator/src/commons/s3/crd.rs
@@ -102,9 +102,10 @@ pub struct Region {
     pub name: String,
 }
 
-/// Having it as `const &str` as well, so we don't always allocate a [`String`] just for comparisons
-const DEFAULT_REGION_NAME: &str = "us-east-1";
 impl Region {
+    /// Having it as `const &str` as well, so we don't always allocate a [`String`] just for comparisons
+    pub const DEFAULT_REGION_NAME: &str = "us-east-1";
+
     fn default_region_name() -> String {
         DEFAULT_REGION_NAME.to_string()
     }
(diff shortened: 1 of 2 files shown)


crates/stackable-operator/src/commons/s3/crd.rs
```diff
@@ -98,18 +98,31 @@ pub enum S3AccessStyle {
...
 #[serde(rename_all = "camelCase")]
 pub struct Region {
-    #[serde(default = "default_region_name")]
+    #[serde(default = "Region::default_region_name")]
     pub name: String,
 }
 
+/// Having it as `const &str` as well, so we don't always allocate a [`String`] just for comparisons
+const DEFAULT_REGION_NAME: &str = "us-east-1";
+impl Region {
```
NickLarsenNZ, line 107: The const can be moved into the impl, and also made pub in case it is needed (eg: in a warning).

```suggestion
impl Region {
    /// Having it as `const &str` as well, so we don't always allocate a [`String`] just for comparisons
    pub const DEFAULT_REGION_NAME: &str = "us-east-1";

```

Then use it as `Self::DEFAULT_REGION_NAME`

NickLarsenNZ (approved):
LGTM

crates/stackable-operator/src/commons/s3/crd.rs
sbernauer: Makes sense, thanks!

NickLarsenNZ (approved):
LGTM

PR #986, opened 2025-03-19
Add loadBalancerClass field to ListenerClass.spec

nightkr:
# Description

See https://github.com/stackabletech/listener-operator/issues/285, part of https://github.com/stackabletech/listener-operator/pull/288

This PR adds a two new fields to ListenerClass:

```yaml
spec:
  loadBalancerClass: foo
  loadBalancerAllocateNodePorts: true
```

Which are mapped to [`Service.spec.{loadBalancerClass,allocateLoadBalancerNodePorts}`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#servicespec-v1-core), respectively. Currently these are ignored unless `serviceType: LoadBalancer`; in v1alpha2, `serviceType` should be turned into a rich enum instead, and these fields be moved into it, like so:

```yaml
spec:
  serviceType:
    loadBalancer:
      loadBalancerClass: foo
      allocateNodePorts: true
```

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] (Integration-)Test cases added → should be tested in stackabletech/listener-operator#288
- [ ] Documentation added or updated → should be documented in stackabletech/listener-operator#288
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


nightkr:
Moving the CRD changes into RFC

siegfriedweber:
The suggestion looks okay for me. Nevertheless, I would like to present my thoughts that lead me to another suggestion:

* How would the spec look like in v1alpha2 for ClusterIP and NodePort where usually nothing is defined?
  ```yaml
  spec:
    serviceType:
      clusterIp: {}
  ```
  The default use case will look slightly uglier.
* The additional Service properties are sometimes specific to the Service type (e.g. `loadBalancerClass`) and sometimes they are not (e.g. `publishNotReadyAddresses`). Kubernetes decided against a deep structure in the Service spec for whatever reason. Why should we introduce one if users already got used to the flat structure?
* The additional properties "are" not a service type, they are part of the service with a specific type, so `serviceType: <type identifier>` feels more accurate than `serviceType: <structure where the next key is the type identifier>`.
* We prefixed `serviceType` with `service` because it is the type of the Service, not the type of the ListenerClass. It might be consistent, to also prefix the other Service properties, e.g. `serviceLoadBalancerClass`.
* The Service property `allocateLoadBalancerNodePorts` is named `loadBalancerAllocateNodePorts` in the Listener, probably to emphasize that it belongs to the Service type `loadBalancer`. It is easier for the user, if Kubernetes knowledge can just be applied to our platform without translation.
* In v1alpha2, `loadBalancerClass` is called `loadBalancerClass` in the `loadBalancer` structure, but `loadBalancerAllocateNodePorts` is renamed to `allocateNodePorts`. In which cases will the prefix be removed and in which ones not?

The ListenerClass requires Service properties. Why not just add an according structure where the properties can be set in exactly the same way, as they are set in Services:

```yaml
spec:
  serviceProperties:
    type: LoadBalancer:
    loadBalancerClass: manual
    allocateLoadBalancerNodePorts: false
```

If we do not want to introduce a breaking change, we could just keep the field `serviceType`:

```yaml
spec:
  serviceType: LoadBalancer
  serviceProperties:
    loadBalancerClass: manual
    allocateLoadBalancerNodePorts: false
```

nightkr:
> How would the spec look like in v1alpha2 for ClusterIP and NodePort where usually nothing is defined? 

Yeah.

> The additional Service properties are sometimes specific to the Service type (e.g. loadBalancerClass) and sometimes they are not (e.g. publishNotReadyAddresses). Kubernetes decided against a deep structure in the Service spec for whatever reason. Why should we introduce one if users already got used to the flat structure?

Looking back at v1.0.0 (https://github.com/kubernetes/kubernetes/blob/v1.0.0/api/swagger-spec/v1.json#L13468-L13499), they seem to have made the same mistake as us: made it a string enum before they had any per-type options, and then they were unable to change it since it was already a stable API.

They also do some weird pseudo-enumisms (For example: "This field will be wiped when a service is updated to a non 'LoadBalancer' type." for `loadBalancerClass`).

> We prefixed serviceType with service because it is the type of the Service, not the type of the ListenerClass. It might be consistent, to also prefix the other Service properties, e.g. serviceLoadBalancerClass.

In retrospect, I think this might've been a mistake. Creating a backing Service is one way to implement a Listener, but it's not really core to the API contract at all (for example, I could see future ListenerClass types that use Ingress or Gateway instead, or which bypass K8s networking entirely).

> The additional properties "are" not a service type, they are part of the service with a specific type

They are parameters for instantiating that service type, and is similar to how rich enums are handled elsewhere ([`Volume`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#volume-v1-core)). Unless you mean that you object to the word "type", in which case I agree but don't have a better option to offer.

> In v1alpha2, loadBalancerClass is called loadBalancerClass in the loadBalancer structure, but loadBalancerAllocateNodePorts is renamed to allocateNodePorts. In which cases will the prefix be removed and in which ones not?

Mentally, I read `loadBalancerClass` as "load-balancer class" (one concept that _implies_ that it is only active in load-balancer mode), but `loadBalancerAllocateNodePorts` as "when in load-balancer mode, allocate node ports?" (one concept guarded by a conditional). I would drop the conditional when implied by context, but would not drop half the concept name.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Added `ListenerClass.spec.loadBalancerClass` field ([#986]).
```
siegfriedweber, line 9: Should `ListenerClass.spec.loadBalancerAllocateNodePorts` also be mentioned?

siegfriedweber (changes requested):

siegfriedweber (approved):
LGTM

nightkr:
Clippy runs fine locally, the remaining CI failure is "just" the pinning issue that we've already been discussing.

PR #1016, opened 2025-04-22
docs(stackable-telemetry): Add missing env vars to doc comment

Techassi:
Part of https://github.com/stackabletech/issues/issues/639

### DIFF
--- a/crates/stackable-telemetry/src/tracing/mod.rs
+++ b/crates/stackable-telemetry/src/tracing/mod.rs
@@ -1,4 +1,4 @@
-//! This module contains functionality to initialise tracing Subscribers for
+//! This module contains functionality to initialize tracing Subscribers for
 //! console output, file output, and OpenTelemetry OTLP export for traces and logs.
 //!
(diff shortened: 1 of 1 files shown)


crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -222,23 +222,25 @@ pub enum Error {
 /// ### Console logs
```
NickLarsenNZ, line 222: As discussed on the call, we could add a comment mentioning that the arguments in parentheses are only applicable when the clap feature is enabled.

NickLarsenNZ (approved):
Good spotting. One suggestion.

crates/stackable-telemetry/src/tracing/mod.rs
Techassi: Done in 367580ce57a21c2ca45c334caac35ea25f9711c2.

NickLarsenNZ (approved):
LGTM

PR #1021, opened 2025-04-30
chore: Dependecy bumps

sbernauer:
# Description

Leftovers
```
cargo +nightly update -Z unstable-options --breaking --dry-run
   Upgrading const-oid ^0.9.6 -> ^0.10.1
   Upgrading rand_core ^0.6.4 -> ^0.9.3
```

can not be bumped because of transitive dependecy version mismatches.

### DIFF
--- a/crates/stackable-telemetry/CHANGELOG.md
+++ b/crates/stackable-telemetry/CHANGELOG.md
@@ -30,10 +30,12 @@ All notable changes to this project will be documented in this file.
   - `FileLogSettingsBuilder::with_max_log_files` which took a `usize` was renamed to
     `FileLogSettingsBuilder::with_max_files` and now takes an `impl Into<Option<usize>>`
     for improved builder ergonomics.
+- Bump `opentelemetry` and related crates to `0.29.x` and `tracing-opentelemetry` to `0.30.0` ([#1021]).
 
 [#1009]: https://github.com/stackabletech/operator-rs/pull/1009
 [#1010]: https://github.com/stackabletech/operator-rs/pull/1010
 [#1012]: https://github.com/stackabletech/operator-rs/pull/1012
+[#1021]: https://github.com/stackabletech/operator-rs/pull/1021
 
 ## [0.5.0] - 2025-04-08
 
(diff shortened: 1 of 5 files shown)


crates/stackable-versioned-macros/fixtures/README.md
```diff
@@ -1,5 +1,8 @@
 # Snapshot Testing
 
+> [!NOTE]
+> Please have `rust-src` installed, e.g. using `rustup component add rust-src`.
+
 > [!NOTE]
 > Also see the compile-fail tests, described [here](../tests/README.md).
```
Techassi, line 7: suggestion: Combine both notes.

```suggestion
> [!NOTE]
> Please have `rust-src` installed, e.g. using `rustup component add rust-src`.
>
> Also see the compile-fail tests, described [here](../tests/README.md).
```

Techassi (commented):
LGTM in general.

My suggestion will also fix the pre-commit failure. If we decide not to use it, we have the fix the check by updating the markdownlint config.

I would also like to see an entry regarding the opentelemetry bumps in the `stackable-telemetry` changelog.

crates/stackable-telemetry/CHANGELOG.md
```diff
@@ -30,10 +30,12 @@ All notable changes to this project will be documented in this file.
   - `FileLogSettingsBuilder::with_max_log_files` which took a `usize` was renamed to
     `FileLogSettingsBuilder::with_max_files` and now takes an `impl Into<Option<usize>>`
     for improved builder ergonomics.
+- Bump `opentelemetry` and related crates to `0.29.x` and `tracing-opentelemetry` to `0.30.0` ([#1021]).
```
Techassi, line 33: Nice, thank you! However, this needs to be moved into the "Unreleased" section instead.

Techassi (changes requested):

crates/stackable-telemetry/CHANGELOG.md
sbernauer: Good catch, sorry!

Techassi (approved):

PR #1025, opened 2025-05-07
fix(stackable-operator): Re-export versioned error types

Techassi:
This PR adds public re-exports for versioned CRD-specific error types.

> [!NOTE]
> The adjustments to the doc comments might need to be tweaked.


NickLarsenNZ (approved):
LGTM

PR #1026, opened 2025-05-08
fix(stackable-telemetry): Only derive clap::ValueEnum if the clap feature is enabled

NickLarsenNZ:
Resolves:

```
use of unresolved module or unknown crate `clap`
```

image


sbernauer (approved):
That's why I test with no and all features in my private project :)
https://github.com/sbernauer/breakwater/blob/aa7481042af303b8380a41890d6f7cedeb8c6591/.github/workflows/build.yml#L122-L136

sbernauer:
Already raised https://github.com/stackabletech/operator-rs/pull/1027 ;)

PR #1038, opened 2025-05-19
feat(stackable-versioned): Add kube_client to crate overrides

Techassi:
Requested by @sbernauer and needed for the ongoing CRD downgrade work.


sbernauer (approved):
Thanks!

PR #1044, opened 2025-05-21
feat!(stackable-certs): Support adding SAN entries

sbernauer:
# Description

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/crates/stackable-certs/src/ca/mod.rs
+++ b/crates/stackable-certs/src/ca/mod.rs
@@ -1,6 +1,6 @@
 //! Contains types and functions to generate and sign certificate authorities
 //! (CAs).
-use std::str::FromStr;
+use std::{fmt::Debug, str::FromStr};
 
 use const_oid::db::rfc5280::{ID_KP_CLIENT_AUTH, ID_KP_SERVER_AUTH};
 use k8s_openapi::api::core::v1::Secret;
@@ -9,9 +9,10 @@ use snafu::{OptionExt, ResultExt, Snafu};
 use stackable_operator::{client::Client, commons::secret::SecretReference, time::Duration};
 use tracing::{debug, instrument};
 use x509_cert::{
+    Certificate,
     builder::{Builder, CertificateBuilder, Profile},
-    der::{DecodePem, pem::LineEnding, referenced::OwnedToRef},
-    ext::pkix::{AuthorityKeyIdentifier, ExtendedKeyUsage},
+    der::{DecodePem, asn1::Ia5String, pem::LineEnding, referenced::OwnedToRef},
+    ext::pkix::{AuthorityKeyIdentifier, ExtendedKeyUsage, SubjectAltName, name::GeneralName},
     name::Name,
     serial_number::SerialNumber,
     spki::{EncodePublicKey, SubjectPublicKeyInfoOwned},
@@ -66,14 +67,20 @@ pub enum Error {
 
     #[snafu(display("failed to parse AuthorityKeyIdentifier"))]
     ParseAuthorityKeyIdentifier { source: x509_cert::der::Error },
+
+    #[snafu(display("The subject alternative DNS name \"{dns_name}\" is not a Ia5String"))]
+    SaDnsNameNotAIa5String {
+        dns_name: String,
+        source: x509_cert::der::Error,
+    },
 }
 
 /// Custom implementation of [`std::cmp::PartialEq`] because some inner types
 /// don't implement it.
 ///
-/// Note that this implementation is restritced to testing because there is a
+/// Note that this implementation is restricted to testing because there is a
 /// variant that is impossible to compare, and will cause a panic if it is
-/// attemped.
+/// attempted.
 #[cfg(test)]
 impl PartialEq for Error {
     fn eq(&self, other: &Self) -> bool {
@@ -170,7 +177,7 @@ where
     /// These parameters include:
     ///
     /// - a randomly generated serial number
-    /// - a default validity of one hour (see [`DEFAULT_CA_VALIDITY_SECONDS`])
+    /// - a default validity of one hour (see [`DEFAULT_CA_VALIDITY`])
     ///
     /// The CA contains the public half of the provided `signing_key` and is
     /// signed by the private half of said key.
@@ -181,9 +188,8 @@ where
     #[instrument(name = "create_certificate_authority", skip(signing_key_pair))]
     pub fn new(signing_key_pair: S) -> Result<Self> {
         let serial_number = rand::random::<u64>();
-        let validity = Duration::from_secs(DEFAULT_CA_VALIDITY_SECONDS);
 
-        Self::new_with(signing_key_pair, serial_number, validity)
+        Self::new_with(signing_key_pair, serial_number, DEFAULT_CA_VALIDITY)
     }
 
     /// Creates a new CA certificate.
@@ -200,8 +206,8 @@ where
         // We don't allow customization of the CA subject by callers. Every CA
         // created by us should contain the same subject consisting a common set
         // of distinguished names (DNs).
-        let subject = Name::from_str(ROOT_CA_SUBJECT).context(ParseSubjectSnafu {
-            subject: ROOT_CA_SUBJECT,
+        let subject = Name::from_str(SDP_ROOT_CA_SUBJECT).context(ParseSubjectSnafu {
+            subject: SDP_ROOT_CA_SUBJECT,
         })?;
 
         let spki_pem = signing_key_pair
@@ -267,15 +273,16 @@ where
     /// authentication, because they include [`ID_KP_CLIENT_AUTH`] and
     /// [`ID_KP_SERVER_AUTH`] in the extended key usage extension.
     ///
-    /// It is also possible to directly greate RSA or ECDSA-based leaf
+    /// It is also possible to directly create RSA or ECDSA-based leaf
     /// certificates using [`CertificateAuthority::generate_rsa_leaf_certificate`]
     /// and [`CertificateAuthority::generate_ecdsa_leaf_certificate`].
     #[instrument(skip(self, key_pair))]
-    pub fn generate_leaf_certificate<T>(
+    pub fn generate_leaf_certificate<'a, T>(
         &mut self,
         key_pair: T,
         name: &str,
         scope: &str,
+        subject_alterative_dns_names: impl IntoIterator<Item = &'a str> + Debug,
         validity: Duration,
     ) -> Result<CertificatePair<T>>
     where
@@ -301,10 +308,6 @@ where
         let spki = SubjectPublicKeyInfoOwned::from_pem(spki_pem.as_bytes())
             .context(DecodeSpkiFromPemSnafu)?;
 
-        // The leaf certificate can be used for WWW client and server
-        // authentication. This is a base requirement for TLS certs.
-        let eku = ExtendedKeyUsage(vec![ID_KP_CLIENT_AUTH, ID_KP_SERVER_AUTH]);
-
         let signer = self.certificate_pair.key_pair.signing_key();
         let mut builder = CertificateBuilder::new(
             Profile::Leaf {
@@ -325,9 +328,27 @@ where
         )
         .context(CreateCertificateBuilderSnafu)?;
 
-        // Again, add the extension created above.
+        // The leaf certificate can be used for WWW client and server
+        // authentication. This is a base requirement for TLS certs.
         builder
-            .add_extension(&eku)
+            .add_extension(&ExtendedKeyUsage(vec![
+                ID_KP_CLIENT_AUTH,
+                ID_KP_SERVER_AUTH,
+            ]))
+            .context(AddCertificateExtensionSnafu)?;
+
+        let sans = subject_alterative_dns_names
+            .into_iter()
+            .map(|dns_name| {
+                Ok(GeneralName::DnsName(Ia5String::new(dns_name).context(
+                    SaDnsNameNotAIa5StringSnafu {
+                        dns_name: dns_name.to_string(),
+                    },
+                )?))
+            })
+            .collect::<Result<Vec<_>, Error>>()?;
+        builder
+            .add_extension(&SubjectAltName(sans))
             .context(AddCertificateExtensionSnafu)?;
 
         debug!("create and sign leaf certificate");
@@ -344,14 +365,15 @@ where
     /// See [`CertificateAuthority::generate_leaf_certificate`] for more
     /// information.
     #[instrument(skip(self))]
-    pub fn generate_rsa_leaf_certificate(
+    pub fn generate_rsa_leaf_certificate<'a>(
         &mut self,
         name: &str,
         scope: &str,
+        subject_alterative_dns_names: impl IntoIterator<Item = &'a str> + Debug,
         validity: Duration,
     ) -> Result<CertificatePair<rsa::SigningKey>> {
         let key = rsa::SigningKey::new().context(GenerateRsaSigningKeySnafu)?;
-        self.generate_leaf_certificate(key, name, scope, validity)
+        self.generate_leaf_certificate(key, name, scope, subject_alterative_dns_names, validity)
     }
 
     /// Generates an ECDSAasync -based leaf certificate which is signed by this CA.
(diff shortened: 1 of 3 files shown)


crates/stackable-certs/src/ca/ca_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+///     ca::{CertificateAuthority, CertificateAuthorityBuilder},
+/// };
+///
+/// let ca: CertificateAuthority<ecdsa::SigningKey> = CertificateAuthorityBuilder::builder()
+///     .build()
+///     .build_ca()
+///     .expect("failed to build CA");
+/// ```
+#[derive(Builder)]
+pub struct CertificateAuthorityBuilder<'a, SKP>
```
NickLarsenNZ, line 76: I thin this should be:

```suggestion
pub struct CertificateAuthority<'a, SKP>
```

So when we call `build()`, we get back a CA and not a CA Builder.

_Note, the Builder macro would otherwise generate `CertificateAuthorityBuilderBuilder` (ie: it generates the Builder, not us)._

crates/stackable-certs/src/ca/ca_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+///
+/// ```no_run
+/// use stackable_certs::{
+///     keys::ecdsa,
+///     ca::{CertificateAuthority, CertificateAuthorityBuilder},
+/// };
+///
+/// let ca: CertificateAuthority<ecdsa::SigningKey> = CertificateAuthorityBuilder::builder()
+///     .build()
+///     .build_ca()
```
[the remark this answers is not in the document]
NickLarsenNZ: I worked it out below, and recommended renaming the struct.

crates/stackable-certs/src/ca/ca_builder.rs
NickLarsenNZ: To illustrate further:

- `CertificateAuthority::builder() -> CertificateAuthorityBuilder`
- `CertificateAuthorityBuilder::build() ->` a valid `CertificateAuthority`

crates/stackable-certs/src/ca/ca_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+    ParseValidity { source: x509_cert::der::Error },
+
+    #[snafu(display("failed to parse \"{subject}\" as subject"))]
+    ParseSubject {
+        source: x509_cert::der::Error,
+        subject: String,
+    },
+
+    #[snafu(display("failed to create signing key pair"))]
+    CreateSigningKeyPair { source: E },
```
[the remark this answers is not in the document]
NickLarsenNZ: Oh I see, we just don't know the concrete impl because it is a trait's associated type.

crates/stackable-certs/src/ca/ca_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+        serial_number: Option<u64>,
+    ) {
+        assert_eq!(ca_cert.subject, subject.parse().unwrap());
+
+        let not_before = ca_cert.validity.not_before.to_system_time();
+        let not_after = ca_cert.validity.not_after.to_system_time();
+        assert_eq!(
+            not_after
+                .duration_since(not_before)
+                .expect("Failed to calculate duration between notBefore and notAfter"),
```
NickLarsenNZ, line 230: praise: Thanks for using expect to explain why it should be valid in the test :+1:

crates/stackable-certs/src/ca/ca_builder.rs
NickLarsenNZ: Oh, but we already have a struct: `CertificateAuthority<SK>` at https://github.com/stackabletech/operator-rs/blob/88d850d14abe56df13af230c2b74eaa5cb5cfd1e/crates/stackable-certs/src/ca/mod.rs#L19

in my mind, that one should be the one that derives Builder. Otherwise we need to revise what each struct is, because the current `CertificateAuthorityBuilder` is not a builder (per previous message above).

NickLarsenNZ (commented):
Just commenting on a few things. Mostly thoughts.

crates/stackable-certs/src/ca/ca_builder.rs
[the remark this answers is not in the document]
sbernauer: > we just don't know the concrete impl

Exactly, there is no other way to get this work (other than boxing the error). I copied this from some other error enum

crates/stackable-certs/src/ca/ca_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Validity/lifetime of the certificate.
+    ///
+    /// If not specified the default of [`DEFAULT_CA_VALIDITY`] will be used.
+    #[builder(default = DEFAULT_CA_VALIDITY)]
+    validity: Duration,
+
+    /// Serial number of the generated certificate.
+    ///
+    /// If not specified a random serial will be generated.
+    serial_number: Option<u64>,
```
Techassi, line 93: note: This is an internal detail we shouldn't allow being customized by the user of this library. I'm sure there is something like

```suggestion
    /// Serial number of the generated certificate.
    #[builder(skip)]
    serial_number: Option<u64>,
```

crates/stackable-certs/src/ca/ca_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Cryptographic keypair used to sign leaf certificates.
+    ///
+    /// If not specified a random keypair will be generated.
+    signing_key_pair: Option<SKP>,
```
Techassi, line 98: note: I would like to see the same high-level functions to create a RSA or ECDSA based CA.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -1,486 +1,47 @@
...
-    ///
-    /// - a randomly generated serial number
-    /// - a default validity of one hour (see [`DEFAULT_CA_VALIDITY_SECONDS`])
-    ///
-    /// The CA contains the public half of the provided `signing_key` and is
-    /// signed by the private half of said key.
-    ///
-    /// If the default values for the serial number and validity don't satisfy
-    /// the requirements of the caller, use [`CertificateAuthority::new_with`]
-    /// instead.
```
Techassi, line 180: note: Compared to these docs, the current docs lack in depth and technical explanations. I've spent a significant amount of time on these and would like to see them carried over to the new implementation as complete as possible.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -1,486 +1,47 @@
...
-    ///
-    /// The certificate requires a `name` and a `scope`. Both these values
-    /// are part of the certificate subject. The format is: `{name} Certificate
-    /// for {scope}`. These leaf certificates can be used for client/server
-    /// authentication, because they include [`ID_KP_CLIENT_AUTH`] and
-    /// [`ID_KP_SERVER_AUTH`] in the extended key usage extension.
-    ///
-    /// It is also possible to directly greate RSA or ECDSA-based leaf
-    /// certificates using [`CertificateAuthority::generate_rsa_leaf_certificate`]
-    /// and [`CertificateAuthority::generate_ecdsa_leaf_certificate`].
```
Techassi, line 272: note: Again, why is this missing from the new implementation?

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -1,486 +1,47 @@
...
-    /// - a randomly generated serial number
-    /// - a default validity of one hour (see [`DEFAULT_CA_VALIDITY_SECONDS`])
-    ///
-    /// The CA contains the public half of the provided `signing_key` and is
-    /// signed by the private half of said key.
-    ///
-    /// If the default values for the serial number and validity don't satisfy
-    /// the requirements of the caller, use [`CertificateAuthority::new_with`]
-    /// instead.
-    #[instrument(name = "create_certificate_authority", skip(signing_key_pair))]
```
Techassi, line 181: note: The new implementation drops these instrumentations, why?

crates/stackable-certs/src/cert_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// Validity/lifetime of the certificate.
+    ///
+    /// If not specified the default of [`DEFAULT_CERTIFICATE_VALIDITY`] will be used.
+    #[builder(default = DEFAULT_CERTIFICATE_VALIDITY)]
+    validity: Duration,
+
+    /// Serial number of the generated certificate.
+    ///
+    /// If not specified a random serial will be generated.
+    serial_number: Option<u64>,
```
Techassi, line 109: note: Again, I would like this to be an internal details not exposed to the user.

crates/stackable-certs/src/cert_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Cryptographic keypair used to for the certificates.
+    ///
+    /// If not specified a random keypair will be generated.
+    key_pair: Option<KP>,
+
+    /// Mandatorily sign the certificate using the provided [`CertificateAuthority`].
+    signed_by: &'a CertificateAuthority<KP>,
```
Techassi, line 117: note: This should be part of the `.build()` method on the `CertificateBuilder`.

crates/stackable-certs/src/cert_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+        if ca_not_after < cert_not_after {
+            warn!(
+                ca.validity = ?ca_validity,
+                cert.validity = ?validity,
+                ca.not_after = ?ca_not_after,
+                cert.not_after = ?cert_not_after,
+                subject = ?subject,
+                "The lifetime of certificate authority is shorted than the lifetime of the generated certificate",
+            );
+        }
```
Techassi, line 166: praise: Thank you for adding this.

note: This should be a hard error instead.

crates/stackable-certs/src/cert_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+    KP: CertificateKeypair,
+    <KP::SigningKey as signature::Keypair>::VerifyingKey: EncodePublicKey,
+{
+    /// Required subject of the certificate, usually starts with `CN=`.
+    subject: &'a str,
+
+    /// Optional list of subject alternative names (SAN) DNS entries,
+    /// that are added to the certificate.
+    #[builder(default)]
+    subject_alterative_dns_names: &'a [&'a str],
```
Techassi, line 98: suggestion: I would just go with

```suggestion
    /// Optional list of subject alternative names (SAN) DNS entries,
    /// that are added to the certificate.
    #[builder(default)]
    subject_alterative_names: &'a [&'a str],
```

Techassi: Also, there is no value in carrying this as `&str`. They need to be converted into owned values below anyway. The API should reflect that to not hide (possibly) unexpected cloning/allocation.

crates/stackable-certs/src/cert_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+        let sans = self
+            .subject_alterative_dns_names
+            .iter()
+            .map(|dns_name| {
+                Ok(GeneralName::DnsName(Ia5String::new(dns_name).context(
+                    SaDnsNameNotAIa5StringSnafu {
+                        dns_name: dns_name.to_string(),
+                    },
+                )?))
+            })
```
Techassi, line 209: note: Please flatten this a little bit. I'm not a huge fan of these nested code snippets.

Techassi (changes requested):

crates/stackable-certs/src/ca/ca_builder.rs
NickLarsenNZ: Agreed. On that topic, if we want to have proper CAs... then serial should increment.
So, if we restore a CA (eg: on startup, read from a secret to get the CA keypair), we also need information for the certs that have been issued (not by reading available signed certs, since they might be gone already). We basically need an atomic counter.

I'm not 100% certain of the downside if we just always issue the same version though.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -66,14 +67,20 @@ pub enum Error {
 
     #[snafu(display("failed to parse AuthorityKeyIdentifier"))]
     ParseAuthorityKeyIdentifier { source: x509_cert::der::Error },
+
+    #[snafu(display("The subject alternative DNS name \"{dns_name}\" is not a Ia5String"))]
+    SaDnsNameNotAIa5String {
+        dns_name: String,
+        source: x509_cert::der::Error,
+    },
```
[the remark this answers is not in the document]
sbernauer: Chaned to
```rust
    #[snafu(display(
        "failed to parse subject alternative DNS name \"{subject_alternative_dns_name}\" as a Ia5 string"
    ))]
    ParseSubjectAlternativeDnsName {
        subject_alternative_dns_name: String,
        source: x509_cert::der::Error,
    },
```
in 8f86a93

crates/stackable-certs/src/cert_builder.rs
sbernauer: Oh I misread, I though @nightkr requested to support IPs... I added support for that in 8f86a93 :see_no_evil: 

But as this is a builder I prefer to have a function to add DNs names and one to add IP addresses instead of passing a list of enums (which we first need to introduce).

Regarding the allocation, I see the point that this allocates. On the other hand I prefer the simple api, so that not every caller needs to deal with `der::asn1::ia5_string::allocation::Ia5String`

crates/stackable-certs/src/cert_builder.rs
sbernauer: The best I could come up was changing
```rust
        let san_dns = self.subject_alterative_dns_names.iter().map(|dns_name| {
            Ok(GeneralName::DnsName(
                Ia5String::new(dns_name).with_context(|_| ParseSubjectAlternativeDnsNameSnafu {
                    subject_alternative_dns_name: dns_name.to_string(),
                })?,
            ))
        });
```
to
```rust
        let san_dns = self.subject_alterative_dns_names.iter().map(
            |dns_name| -> Result<_, CreateCertificateError<KP::Error>> {
                Ia5String::new(dns_name)
                    .with_context(|_| ParseSubjectAlternativeDnsNameSnafu {
                        subject_alternative_dns_name: dns_name.to_string(),
                    })
                    .map(GeneralName::DnsName)
            },
        );
```
Not sure if that really improves the situation

crates/stackable-certs/src/ca/ca_builder.rs
sbernauer: e9e8bac

crates/stackable-certs/src/cert_builder.rs
sbernauer: Same as https://github.com/stackabletech/operator-rs/pull/1044#discussion_r2108327741

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -200,8 +206,8 @@ where
         // We don't allow customization of the CA subject by callers. Every CA
         // created by us should contain the same subject consisting a common set
         // of distinguished names (DNs).
-        let subject = Name::from_str(ROOT_CA_SUBJECT).context(ParseSubjectSnafu {
-            subject: ROOT_CA_SUBJECT,
+        let subject = Name::from_str(SDP_ROOT_CA_SUBJECT).context(ParseSubjectSnafu {
+            subject: SDP_ROOT_CA_SUBJECT,
```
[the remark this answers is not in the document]
sbernauer: I think I tend to disagree here. My argumentation would be that stackable-certs doesn't know where the subject is coming from. It could be `CN=<pod-name>`, but maybe also a certificate we issue for users (user-provided subject?), so they can connect to e.g. mTLS Kafka.
Better safe than sorry I'd say

crates/stackable-certs/src/ca/mod.rs
sbernauer: I tried carrying over everything that is still relevant: 1322779

crates/stackable-certs/src/cert_builder.rs
nightkr: > Oh I misread, I though @nightkr requested to support IPs... I added support for that in https://github.com/stackabletech/operator-rs/commit/8f86a935a164309ac5b7974045bf9cb623b46ca8 🙈

I was. (At least from the perspective of "if we want to use it for secret-op at some point".)

crates/stackable-certs/src/ca/ca_builder.rs
sbernauer: Added in fd19a03

sbernauer:
We don't urgently need it.
With how we do cert rotation in webhook (e.g. https://github.com/stackabletech/operator-rs/pull/1066) things have changed, the current proposal doesn't work any more.
Closing this, but the code is not gone.

Issue #1032, opened 2025-05-14
`ClusterCondition::last_update_time` is updated on no-ops, causing infinite reconciles (in the worst case)


nightkr:
### Affected version

Yes. (Still an issue on trunk, introduced in #571, rolled out around SDP 23.4.)

### Current and expected behavior

Reconciling a cluster where there nothing has changed should be a no-op.

`ClusterCondition::last_update_time` breaks this expectation since it is set _unconditionally_ to whatever the current time is, rounded to the second (https://github.com/stackabletech/operator-rs/blob/61596d6140a54e57bf71d395172be15631463bca/crates/stackable-operator/src/status/condition/mod.rs#L350-L355). This is registered as another object modification if the new reconcile is not within the same wall-second as the previous one. Depending on how long one reconcile takes, that can cause (up to) an infinite re-reconciliation loop while the object is trying to settle down (which is _likely_ to be an indication that the cluster is struggling to begin with!).

### Possible solution

1. Drop `last_update_time` completely (for compat: either stub it out or make it equivalent to `last_transition_time`)
2. Take the value from whenever the data source for the condition was updated, rather than the current wall time (if it makes sense/is possible for that condition)

### Additional context

Discovered by @siegfriedweber, discussed at https://stackable-workspace.slack.com/archives/C02FZ581UCD/p1747230004370629

### Would you like to work on fixing this bug?

None

maltesander:
The approach back then was to follow the OpenShift `ClusterOperatorStatusCondition`, see https://github.com/openshift/api/blob/b1bcdbc3/config/v1/types_cluster_operator.go#L101.

There, the `last_updated_time` does not even appear so i am not sure why it was introduced here, as this would always only be the last timestamp the operator reconciled, which does not provide much value.

Suggestion is using Solution 1 and just drop it.

PR #1054, opened 2025-06-12
Remove last_ update_time from ClusterCondition

maltesander:
# Description

fixes https://github.com/stackabletech/operator-rs/issues/1032

Zookeeper status: 
```
status:
  conditions:
  - lastTransitionTime: "2025-06-12T10:58:13Z"
    message: All StatefulSet have the requested amount of ready replicas.
    status: "True"
    type: Available
  - message: The cluster is reconciled normally.
    status: "False"
    type: ReconciliationPaused
  - message: The cluster is running.
    status: "False"
    type: Stopped
  discoveryHash: "5807873964697571649"
```

# Release Note
```
The `lastUpdatedTime` was removed from the `ClusterStatus` of each operator. Users previously relying on that value should migrate to `lastTransitionTime`.
```

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Removed
 
-- Removed `last_update_time` from CRD ClusterCondition status ([#1054]).
+- BREAKING: Removed `last_update_time` from CRD ClusterCondition status ([#1054]).
 
 [#1049]: https://github.com/stackabletech/operator-rs/pull/1049
 [#1054]: https://github.com/stackabletech/operator-rs/pull/1054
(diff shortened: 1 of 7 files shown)


crates/stackable-operator/CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ All notable changes to this project will be documented in this file.
 
 - Update `kube` to `1.1.0` ([#1049]).
 
+### Removed
+
+- Removed `last_update_time` from CRD ClusterCondition status ([#1054]).
```
Techassi: note: This should be marked as breaking.

sbernauer: +1 and get a platform release note please :)

maltesander: I tested it and dont feel its breaking. Applying the new CRD and installing the new operator version simply removes the `last_updated_time` from the CRD.

Techassi: It might not be breaking for us, but some users might use this value. As such, a removal is a breaking change for them.

maltesander: added the breaking hint https://github.com/stackabletech/operator-rs/pull/1054/commits/244bfbbd1be6e6dd2bcd38454246da9ebcf0177e

PR #1055, opened 2025-06-13
refactor(stackable-telemetry): Use semantic convention constants

Techassi:
This PR finally switches to the `opentelemetry-semantic-conventions` crate to use the well-defined constants instead of hard-coded strings.

This PR defines constants in addition to upstream ones to be able to use constants across the board instead of mix-and-matching constants and hard-coded strings.


crates/stackable-telemetry/src/instrumentation/axum/mod.rs
```diff
@@ -462,8 +480,8 @@ impl SpanExt for Span {
 
         if new_span_context != current_span_context {
             tracing::trace!(
-                opentelemetry.trace_id.from = ?current_span_context.trace_id(),
-                opentelemetry.trace_id.to = ?new_span_context.trace_id(),
+                { OTEL_TRACE_ID_FROM } = ?current_span_context.trace_id(),
+                { OTEL_TRACE_ID_TO } = ?new_span_context.trace_id(),
```
NickLarsenNZ, line 484: idea (non-blocker): We probably want to link the spans. I can't recall if that is done by Span ID or Trace ID.

See: https://docs.rs/opentelemetry/latest/opentelemetry/trace/trait.Span.html#tymethod.add_link

NickLarsenNZ (approved):
LGTM

crates/stackable-telemetry/src/instrumentation/axum/mod.rs
Techassi: I agree. However, this should be done in a separate PR.

PR #1056, opened 2025-06-13
feat(stackable-versioned)!: Add conversion tracking

Techassi:

> [!CAUTION]
> It should be noted, that only simple additions of fields can be tracked across conversions. Complex action combinations and other types of changes **cannot** be tracked. Support for this will be added in the future.

## Summary

- BREAKING: The macro can only be applied to modules now.
- BREAKING: Crate overrides are now done at the module level, not the container level.
- BREAKING: Top-level CRD specs now use `crd(...)` instead of `k8s(...)`.
- BREAKING: The (optional) status needs to implement `Default`.
- The conversion tracking is behind the `experimental_conversion_tracking` flag.
- A new `nested` flag was added to mark individual fields as nested (indicating that the sub struct also contains changes which need to be tracked).

```rust
#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1alpha2"),
    version(name = "v1beta1"),
    version(name = "v2"),
    version(name = "v3"),
    options(k8s(experimental_conversion_tracking))
)]
pub mod versioned {
    #[versioned(crd(group = "test.stackable.tech", status = "PersonStatus"))]
    #[derive(Clone, Debug, CustomResource, Deserialize, JsonSchema, Serialize)]
    #[serde(rename_all = "camelCase")]
    pub struct PersonSpec {
        username: String,

        // In v1alpha2 first and last name have been added
        #[versioned(added(since = "v1alpha2"))]
        first_name: String,

        #[versioned(added(since = "v1alpha2"))]
        last_name: String,

        // We started out with a enum. As we *need* to provide a default, we have a Unknown variant.
        // Afterwards we figured let's be more flexible and accept any arbitrary String.
        #[versioned(added(since = "v2"), changed(since = "v3", from_type = "Gender"))]
        gender: String,

        #[versioned(nested)]
        socials: Socials,
    }

    // Ideally, we want this to be versioned independently from the CRD spec above. The addition
    // of the `mastodon` field would then require a change on the `socials` field above. But as
    // mentioned, currently we don't support individual versions nor complex action + nested
    // combinations.
    // This still demonstrates that tracking works in nested sub structs on a basic level (for now).
    #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
    pub struct Socials {
        email: String,

        #[versioned(added(since = "v1beta1"))]
        mastodon: String,
    }
}

#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
pub struct PersonStatus {
    pub alive: bool,
}
```

## Learnings from this PR

- We need per-container versions. The current mechanism doesn't make sense for more complex CRDs with multiple sub structs (at different versions).
- We need fallible conversions.
- We need to support to track other types of changes to be able to use this on real CRDs.
- We need to integrate with Serde (for renames, flattening, untagged).


crates/stackable-operator/Cargo.toml
```diff
@@ -16,8 +16,8 @@ versioned = []
 
 [dependencies]
 stackable-telemetry = { path = "../stackable-telemetry", features = ["clap"] }
-stackable-versioned = { path = "../stackable-versioned", features = ["k8s"] }
 stackable-operator-derive = { path = "../stackable-operator-derive" }
+stackable-versioned = { path = "../stackable-versioned" }
```
sbernauer: nit: Ideally we sort them alphabetical

crates/stackable-versioned-macros/src/attrs/item/field.rs
```diff
@@ -36,6 +36,11 @@ pub struct FieldAttributes {
     // FromMeta.
     /// The original attributes for the field.
     pub attrs: Vec<Attribute>,
+
+    /// Indicates that this field's type is a nested sub struct. The indicator
+    /// is needed to let the macro know to generate conversion code with support
+    /// for tracking across struct boundaries.
+    pub nested: Flag,
```
NickLarsenNZ, line 43: This could be hidden behind the `experimental_conversion_tracking` feature flag.

Or, as you mentioned, we could throw an error when it is used without the feature enabled (I actually like this idea better as it helps with discovering features).

crates/stackable-versioned-macros/src/codegen/container/struct/conversion.rs
```diff
@@ -0,0 +1,833 @@
...
+        Some(json_paths)
+    }
+
+    pub(super) fn needs_tracking(&self, version: &VersionDefinition) -> bool {
+        self.fields.iter().any(|f| {
+            f.changes.as_ref().is_some_and(|c| {
+                c.value_is(&version.inner, |s| {
+                    // For now, only added fields need to be tracked. In the future, removals and
+                    // type changes also need to be tracked
+                    matches!(s, ItemStatus::Addition { .. })
```
sbernauer: nit: Please use a match here, so that we can add a TODO for type changes. Also, the compiler will force us to think about new added actions

crates/stackable-versioned/src/lib.rs
```diff
@@ -9,11 +9,170 @@
...
+    }
+}
+
+#[derive(Debug, Snafu)]
+#[snafu(display("unknown API version {api_version:?}"))]
+pub struct UnknownDesiredApiVersionError {
+    pub api_version: String,
+}
+
+pub fn jthong_path(parent: &str, child: &str) -> String {
```
sbernauer, line 173: I think we should stick to "real" JSONPath syntax here. We currently only need `$.foo.bar` and that's it, *but* we really should stick to the syntax, so that we can non-breaking swap out the field with a real JSONPath crate

crates/stackable-versioned/src/lib.rs
```diff
@@ -9,11 +9,170 @@
...
+// NOTE (@Techassi): This struct represents a rough first draft of how tracking values across
+// CRD versions can be achieved. It might change down the line.
+// FIXME (@Techassi): Ideally we don't serialize empty maps. Further, we shouldn't serialize the
+// changedValues field in the status, if there it is empty. This currently "pollutes" the status
+// with empty JSON objects.
+/// Contains changed values during upgrades and downgrades of CRDs.
+#[derive(Clone, Debug, Default, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+pub struct ChangedValues {
+    /// List of values needed when downgrading to a particular version.
+    pub downgrades: HashMap<String, Vec<ChangedValue>>,
```
sbernauer: note: We need to think if we should use BTreeMap instead. HashMap *might* cause infinite reconcile loops

crates/stackable-versioned/src/lib.rs
```diff
@@ -9,11 +9,170 @@
...
+    // TODO (@Techassi): Add a version indicator here if we ever decide to change the tracking
+    // mechanism.
+}
+
+/// Contains a changed value for a single field of the CRD.
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ChangedValue {
+    /// The name of the field of the custom resource this value is for.
+    pub field_name: String,
```
sbernauer: we can probably find a better name for this, e.g. `json_path`

crates/stackable-versioned/tests/person.rs
```diff
@@ -1,50 +1,99 @@
...
-pub struct PersonSpec {
-    username: String,
+pub mod versioned {
+    #[versioned(crd(group = "test.stackable.tech"))]
+    #[derive(
+        Clone,
+        Debug,
+        Eq,
+        Hash,
+        Ord,
```
sbernauer: nit(non-blocking): I think we can remove some unneeded/wrong derives

crates/stackable-versioned/tests/person.rs
```diff
@@ -1,50 +1,99 @@
...
+        PartialEq,
+        PartialOrd,
+        CustomResource,
+        Deserialize,
+        JsonSchema,
+        Serialize,
+    )]
+    #[serde(rename_all = "camelCase")]
+    pub struct PersonSpec {
```
sbernauer, line 54: Please add a PersonStatus. Can be simple. The tests and assertion should also check the status

crates/stackable-versioned/tests/snapshots/conversions__pass@persons_to_v3.json.snap
```diff
@@ -35,7 +45,17 @@ input_file: crates/stackable-versioned/tests/inputs/conversions/pass/persons_to_
           "firstName": "Sebastian",
```
sbernauer: nit: Some name leftovers

crates/stackable-versioned/tests/person.rs
sbernauer, line 1: Currently with this diff we get a lot of compilation failures:
```diff
diff --git a/crates/stackable-versioned/tests/person.rs b/crates/stackable-versioned/tests/person.rs
index 34ee280..25db241 100644
--- a/crates/stackable-versioned/tests/person.rs
+++ b/crates/stackable-versioned/tests/person.rs
@@ -81,7 +81,7 @@ pub mod versioned {
         )]
         gender: String,
 
-        #[versioned(nested)]
+        #[versioned(nested, added(since = "v1alpha2", default = "default_socials"))]
         socials: Socials,
     }
 
@@ -110,6 +110,12 @@ fn default_gender() -> Gender {
     Gender::Unknown
 }
 
+fn default_socials() -> v1alpha2::Socials {
+    v1alpha2::Socials {
+        email: "".to_string(),
+    }
+}
+
 impl From<Gender> for String {
     fn from(value: Gender) -> Self {
         match value {
```

crates/stackable-versioned/tests/person.rs
Techassi: Done, removed in 968a52c.

crates/stackable-versioned/tests/person.rs
Techassi: Done, added in 968a52c.

crates/stackable-versioned/src/lib.rs
Techassi: I agree, Kubernetes also uses `jsonPath` (`json_path` in Rust) for similar fields, see [additional printer columns](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns) for example.

crates/stackable-versioned/src/lib.rs
Techassi: Yeah, that makes sense. I will adopt the code.

Techassi: Done in 20427af7.

crates/stackable-versioned/src/lib.rs
Techassi: Changed in 790bfb5f.

crates/stackable-versioned/src/lib.rs
Techassi: Changed in db85e294.

crates/stackable-versioned-macros/src/codegen/container/struct/conversion.rs
Techassi: Changed in 88f1df8b.

crates/stackable-operator/Cargo.toml
Techassi: Done in 6cf798ef.

crates/stackable-versioned/tests/snapshots/conversions__pass@persons_to_v3.json.snap
Techassi: Done in 6a9005f6.

crates/stackable-versioned-macros/src/attrs/item/field.rs
Techassi: I've instead opted to fall back to implement normal conversion functions (a simple `From` impl) when the `nested` flag is set, but the `experimental_conversion_tracking` flag is not. We can however make this a hard error instead.

crates/stackable-versioned/tests/person.rs
sbernauer: The following patch fixes at least the &str vs String mismatch problem
```diff
diff --git a/crates/stackable-versioned-macros/src/codegen/item/field.rs b/crates/stackable-versioned-macros/src/codegen/item/field.rs
index 9a91dd5..478b243 100644
--- a/crates/stackable-versioned-macros/src/codegen/item/field.rs
+++ b/crates/stackable-versioned-macros/src/codegen/item/field.rs
@@ -203,7 +203,7 @@ impl VersionedField {
                                     );
 
                                     Some(quote! {
-                                        #to_ident: #from_struct_ident.#from_ident.tracking_into(status, #json_path_ident),
+                                        #to_ident: #from_struct_ident.#from_ident.tracking_into(status, &#json_path_ident),
                                     })
                                 } else {
                                     Some(quote! {
@@ -224,7 +224,7 @@ impl VersionedField {
                                     );
 
                                     Some(quote! {
-                                        #from_ident: #from_struct_ident.#to_ident.tracking_into(status, #json_path_ident),
+                                        #from_ident: #from_struct_ident.#to_ident.tracking_into(status, &#json_path_ident),
                                     })
                                 } else {
                                     Some(quote! {
@@ -250,7 +250,7 @@ impl VersionedField {
                                     );
 
                                     Some(quote! {
-                                        #next_field_ident: #from_struct_ident.#old_field_ident.tracking_into(status, #json_path_ident),
+                                        #next_field_ident: #from_struct_ident.#old_field_ident.tracking_into(status, &#json_path_ident),
                                     })
                                 } else {
                                     Some(quote! {
```

sbernauer (approved):
LGTM % remaining comment and conflicts

crates/stackable-versioned/tests/person.rs
Techassi: Yeah, that's the only thing I would fix for now. Support for complex action + nested combinations should come in with a separate PR.

I will apply your patch asap.

Techassi: Applied in 65a768c6.

crates/stackable-versioned-macros/src/attrs/item/field.rs
Techassi: Changed to a hard error after discussing this with @sbernauer in 710d69be.

Techassi:
Ah! This was allowed previously. I will update the test.

sbernauer (dismissed):

sbernauer (approved):

PR #1057, opened 2025-06-17
feat!(stackable-certs): Support adding SAN entries

sbernauer:
# Description

As https://github.com/stackabletech/operator-rs/pull/1044 didn't move along, splitting the relevant change out to unblock CRD versioning.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-certs/CHANGELOG.md
+++ b/crates/stackable-certs/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 
 - BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
   now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
-  this is needed for basically all modern TLS certificate validations when used with HTTPS.
+  this is needed when the HTTPS server is accessible on multiple DNS names and/or IPs.
   Pass an empty list (`[]`) to keep the existing behavior ([#1057]).
 - BREAKING: The constant `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
   Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#1057]).
--- a/crates/stackable-certs/CHANGELOG.md
+++ b/crates/stackable-certs/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
+  now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
+  this is needed for basically all modern TLS certificate validations when used with HTTPS.
+  Pass an empty list (`[]`) to keep the existing behavior ([#1044]).
+- BREAKING: The constant `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
+  Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#1044]).
+- Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1044]).
+
 ## [0.3.1] - 2024-07-10
 
 ### Changed
@@ -11,6 +21,7 @@ All notable changes to this project will be documented in this file.
 - Bump rust-toolchain to 1.79.0 ([#822]).
 
 [#822]: https://github.com/stackabletech/operator-rs/pull/822
+[#1044]: https://github.com/stackabletech/operator-rs/pull/1044
 
 ## [0.3.0] - 2024-05-08
 
--- a/crates/stackable-certs/CHANGELOG.md
+++ b/crates/stackable-certs/CHANGELOG.md
@@ -6,14 +6,21 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
-  now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
-  this is needed when the HTTPS server is accessible on multiple DNS names and/or IPs.
-  Pass an empty list (`[]`) to keep the existing behavior ([#1057]).
+- Add the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1057]).
+
+### Changed
+
+- BREAKING: The functions `generate_leaf_certificate`, `generate_rsa_leaf_certificate` and
+  `generate_ecdsa_leaf_certificate` of `CertificateAuthority` accept an additional parameter
+  `subject_alterative_dns_names` ([#1057]).
+  - The passed SANs are added to the generated certificate, this is needed when the HTTPS server is
+    accessible on multiple DNS names and/or IPs.
+  - Pass an empty list (`[]`) to keep the existing behavior.
 - BREAKING: Constants have been renamed/retyped ([#1057]):
   - `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
   - `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT`.
-- Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1057]).
+
+[#1057]: https://github.com/stackabletech/operator-rs/pull/1057
 
 ## [0.3.1] - 2024-07-10
 
@@ -22,7 +29,6 @@ All notable changes to this project will be documented in this file.
 - Bump rust-toolchain to 1.79.0 ([#822]).
 
 [#822]: https://github.com/stackabletech/operator-rs/pull/822
-[#1057]: https://github.com/stackabletech/operator-rs/pull/1057
 
 ## [0.3.0] - 2024-05-08
 
--- a/crates/stackable-certs/CHANGELOG.md
+++ b/crates/stackable-certs/CHANGELOG.md
@@ -9,10 +9,10 @@ All notable changes to this project will be documented in this file.
 - BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
   now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
   this is needed for basically all modern TLS certificate validations when used with HTTPS.
-  Pass an empty list (`[]`) to keep the existing behavior ([#XXXX]).
+  Pass an empty list (`[]`) to keep the existing behavior ([#1057]).
 - BREAKING: The constant `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
-  Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#XXXX]).
-- Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#XXXX]).
+  Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#1057]).
+- Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1057]).
 
 ## [0.3.1] - 2024-07-10
 
@@ -21,6 +21,7 @@ All notable changes to this project will be documented in this file.
 - Bump rust-toolchain to 1.79.0 ([#822]).
 
 [#822]: https://github.com/stackabletech/operator-rs/pull/822
+[#1057]: https://github.com/stackabletech/operator-rs/pull/1057
 
 ## [0.3.0] - 2024-05-08
 
--- a/crates/stackable-certs/CHANGELOG.md
+++ b/crates/stackable-certs/CHANGELOG.md
@@ -10,8 +10,9 @@ All notable changes to this project will be documented in this file.
   now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
   this is needed when the HTTPS server is accessible on multiple DNS names and/or IPs.
   Pass an empty list (`[]`) to keep the existing behavior ([#1057]).
-- BREAKING: The constant `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
-  Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#1057]).
+- BREAKING: Constants have been renamed/retyped ([#1057]):
+  - `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
+  - `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT`.
 - Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1057]).
 
 ## [0.3.1] - 2024-07-10
--- a/crates/stackable-certs/CHANGELOG.md
+++ b/crates/stackable-certs/CHANGELOG.md
@@ -9,10 +9,10 @@ All notable changes to this project will be documented in this file.
 - BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
   now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
   this is needed for basically all modern TLS certificate validations when used with HTTPS.
-  Pass an empty list (`[]`) to keep the existing behavior ([#1044]).
+  Pass an empty list (`[]`) to keep the existing behavior ([#XXXX]).
 - BREAKING: The constant `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
-  Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#1044]).
-- Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1044]).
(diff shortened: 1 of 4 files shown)


crates/stackable-certs/src/ca/mod.rs
```diff
@@ -325,9 +330,27 @@ where
...
+            .context(AddCertificateExtensionSnafu)?;
+
+        let sans = subject_alterative_dns_names
+            .into_iter()
+            .map(|dns_name| {
+                Ok(GeneralName::DnsName(Ia5String::new(dns_name).context(
+                    ParseSubjectAlternativeDnsNameSnafu {
+                        subject_alternative_dns_name: dns_name.to_string(),
+                    },
+                )?))
```
NickLarsenNZ: I find these a little bit of a handful and would tend to prefer this:

```suggestion
                let ia5_dns_name = Ia5String::new(dns_name).context(
                    ParseSubjectAlternativeDnsNameSnafu {
                        subject_alternative_dns_name: dns_name.to_string(),
                    },
                )?;
                Ok(GeneralName::DnsName(ia5_dns_name))
```

_You're welcome to ignore this suggestion_

NickLarsenNZ: Even better if we had a TryFrom impl, we could then do:

```rust
        let sans = subject_alterative_dns_names
            .into_iter()
            .map(GeneralName::try_from)
            .collect::<Result<Vec<_>, Error>>()?;
```

crates/stackable-certs/CHANGELOG.md
```diff
@@ -4,13 +4,24 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
+  now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
+  this is needed for basically all modern TLS certificate validations when used with HTTPS.
```
NickLarsenNZ: I disagree that it is needed.

It is only needed when you want to use the certificate with multiple names (in addition to the commonName). So while it is common, it isn't correct to say that it is _needed_ due to _modern_ validation impls.

_Unless I've been hiding under a rock and you have seen problems validating on commonName alone?_

```suggestion
  this is often needed.
```

or

```suggestion
  this is needed when the HTTPS server is accessible on multiple DNS names and/or IPs.
```

crates/stackable-certs/CHANGELOG.md
```diff
@@ -4,13 +4,24 @@ All notable changes to this project will be documented in this file.
...
+  Pass an empty list (`[]`) to keep the existing behavior ([#1057]).
+- BREAKING: The constant `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
+  Also, the constant `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT` ([#1057]).
```
NickLarsenNZ:
```suggestion
- BREAKING: Constants have been renamed/retyped ([#1057]):
  - `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
  - `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT`.
```

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -325,9 +330,27 @@ where
...
         .context(CreateCertificateBuilderSnafu)?;
 
-        // Again, add the extension created above.
+        // The leaf certificate can be used for WWW client and server
+        // authentication. This is a base requirement for TLS certs.
         builder
-            .add_extension(&eku)
+            .add_extension(&ExtendedKeyUsage(vec![
+                ID_KP_CLIENT_AUTH,
+                ID_KP_SERVER_AUTH,
```
NickLarsenNZ, line 337: I know it was existing behaviour, so no change required now, but I don't think we should  make any certificate usable as a Client Cert. It should be intentional.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -325,9 +330,27 @@ where
...
+            .map(|dns_name| {
+                Ok(GeneralName::DnsName(Ia5String::new(dns_name).context(
+                    ParseSubjectAlternativeDnsNameSnafu {
+                        subject_alternative_dns_name: dns_name.to_string(),
+                    },
+                )?))
+            })
+            .collect::<Result<Vec<_>, Error>>()?;
+        builder
+            .add_extension(&SubjectAltName(sans))
```
NickLarsenNZ, line 352: ~Just checking what happens if there are no sans (note to self, see if tests show it).~

Nice, there is a test for the empty sans.

NickLarsenNZ (changes requested):
Just some changes to the changelog, and the rest (I think) are only suggestions.

crates/stackable-certs/src/ca/mod.rs
sbernauer: I'm going to pick your suggestion. I think `TryFrom` is a beast on it's own

crates/stackable-certs/CHANGELOG.md
sbernauer: You are right, thanks for the suggestion!

crates/stackable-certs/src/ca/mod.rs
sbernauer: Sounds like a good idea for the builder pattern PR (#1044)!

crates/stackable-certs/src/ca/mod.rs
NickLarsenNZ: 67cba1800f53c40d7b26e08aa2c5f91d5f55bd96

sbernauer:
Thanks for the review!

Techassi (changes requested):
I didn't take a close look at this, but I will soon.

Blocking it for now.

crates/stackable-certs/CHANGELOG.md
```diff
@@ -4,13 +4,25 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- BREAKING: `CertificateAuthority::generate_leaf_certificate` (and `generate_rsa_leaf_certificate` and `generate_ecdsa_leaf_certificate`)
+  now take an additional parameter `subject_alterative_dns_names`. The passed SANs are added to the generated certificate,
+  this is needed when the HTTPS server is accessible on multiple DNS names and/or IPs.
+  Pass an empty list (`[]`) to keep the existing behavior ([#1057]).
```
Techassi: note: This is a change, not an addition. As such, this should live in the "Changed" section.

suggestion: Slight reword.

```suggestion
- BREAKING: The functions `generate_leaf_certificate`, `generate_rsa_leaf_certificate` and
  `generate_ecdsa_leaf_certificate` of `CertificateAuthority` accept an additional parameter
  `subject_alterative_dns_names` ([#1057]).
  - The passed SANs are added to the generated certificate, this is needed when the HTTPS server is
    accessible on multiple DNS names and/or IPs.
  - Pass an empty list (`[]`) to keep the existing behavior.
```

crates/stackable-certs/CHANGELOG.md
```diff
@@ -4,13 +4,25 @@ All notable changes to this project will be documented in this file.
...
+- BREAKING: Constants have been renamed/retyped ([#1057]):
+  - `DEFAULT_CA_VALIDITY_SECONDS` has been renamed to `DEFAULT_CA_VALIDITY` and now is of type `stackable_operator::time::Duration`.
+  - `ROOT_CA_SUBJECT` has been renamed to `SDP_ROOT_CA_SUBJECT`.
```
Techassi, line 21: note: This also a change and should be moved.

crates/stackable-certs/CHANGELOG.md
```diff
@@ -4,13 +4,25 @@ All notable changes to this project will be documented in this file.
...
+- Added the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1057]).
```
Techassi: suggestion: Slight reword.

```suggestion
- Add the function `CertificateAuthority::ca_cert` to easily get the CA `Certificate` ([#1057]).
```

crates/stackable-certs/CHANGELOG.md
```diff
@@ -4,13 +4,25 @@ All notable changes to this project will be documented in this file.
...
+
 ## [0.3.1] - 2024-07-10
 
 ### Changed
 
 - Bump rust-toolchain to 1.79.0 ([#822]).
 
 [#822]: https://github.com/stackabletech/operator-rs/pull/822
+[#1057]: https://github.com/stackabletech/operator-rs/pull/1057
```
Techassi: note: This link reference is in the wrong place.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -66,14 +67,22 @@ pub enum Error {
 
     #[snafu(display("failed to parse AuthorityKeyIdentifier"))]
     ParseAuthorityKeyIdentifier { source: x509_cert::der::Error },
+
+    #[snafu(display(
+        "failed to parse subject alternative DNS name \"{subject_alternative_dns_name}\" as a Ia5 string"
+    ))]
```
Techassi: note: Get rid of the escaped quotes.

```suggestion
    #[snafu(display(
        "failed to parse subject alternative DNS name {subject_alternative_dns_name:?} as a Ia5 string"
    ))]
```

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -200,8 +208,8 @@ where
         // We don't allow customization of the CA subject by callers. Every CA
         // created by us should contain the same subject consisting a common set
         // of distinguished names (DNs).
-        let subject = Name::from_str(ROOT_CA_SUBJECT).context(ParseSubjectSnafu {
-            subject: ROOT_CA_SUBJECT,
+        let subject = Name::from_str(SDP_ROOT_CA_SUBJECT).context(ParseSubjectSnafu {
+            subject: SDP_ROOT_CA_SUBJECT,
```
Techassi: note: This CA subject must always parse. As such, we should use `expect` here.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -468,19 +498,57 @@ fn format_leaf_certificate_subject(name: &str, scope: &str) -> Result<Name> {
 
 #[cfg(test)]
 mod tests {
+    use const_oid::ObjectIdentifier;
+
     use super::*;
 
+    const TEST_CERT_LIFETIME: Duration = Duration::from_hours_unchecked(1);
+    const TEST_SAN: &str = "airflow-0.airflow.default.svc.cluster.local";
```
Techassi: question: Can we use a non-product specific test value here? This comes to mind:

```suggestion
    const TEST_SAN: &str = "example-0.example.default.svc.cluster.local";
```

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -468,19 +498,57 @@ fn format_leaf_certificate_subject(name: &str, scope: &str) -> Result<Name> {
...
+
     #[tokio::test]
     async fn rsa_key_generation() {
         let mut ca = CertificateAuthority::new_rsa().unwrap();
-        ca.generate_rsa_leaf_certificate("Airflow", "pod", Duration::from_secs(3600))
-            .unwrap();
+        let cert = ca
+            .generate_rsa_leaf_certificate("Airflow", "pod", [TEST_SAN], TEST_CERT_LIFETIME)
```
Techassi: note: In a similar fashion, we should replace "Airflow" with "Example" here as well.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -468,19 +498,57 @@ fn format_leaf_certificate_subject(name: &str, scope: &str) -> Result<Name> {
...
+            .expect("RSA certificate generation failed");
```
Techassi: praise: Nicely done!

note: The message should instead read _why_ this will never fail.

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -468,19 +498,57 @@ fn format_leaf_certificate_subject(name: &str, scope: &str) -> Result<Name> {
...
+    fn assert_cert_attributes(cert: &Certificate) {
+        let cert = &cert.tbs_certificate;
+        // Test subject
+        assert_eq!(
+            cert.subject,
+            Name::from_str("CN=Airflow Certificate for pod").unwrap()
+        );
+
+        // Test SAN extension is present
+        let extensions = cert.extensions.as_ref().expect("cert had no extension");
```
Techassi: note: This should state _why_ it should never fail. Something like this comes to mind:

```suggestion
        let extensions = cert.extensions.as_ref().expect("cert must have extensions");
```

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -468,19 +498,57 @@ fn format_leaf_certificate_subject(name: &str, scope: &str) -> Result<Name> {
...
+                .any(|ext| ext.extn_id == ObjectIdentifier::new_unwrap("2.5.29.17"))
+        );
+
+        // Test lifetime
+        let not_before = cert.validity.not_before.to_system_time();
+        let not_after = cert.validity.not_after.to_system_time();
+        assert_eq!(
+            not_after
+                .duration_since(not_before)
+                .expect("Failed to calculate duration between notBefore and notAfter"),
```
Techassi: note: This should state _why_ it should never fail.

Techassi (changes requested):

crates/stackable-certs/CHANGELOG.md
sbernauer: 49faa95

crates/stackable-certs/CHANGELOG.md
sbernauer: 49faa95

crates/stackable-certs/CHANGELOG.md
sbernauer: 49faa95

crates/stackable-certs/CHANGELOG.md
sbernauer: 49faa95

crates/stackable-certs/src/ca/mod.rs
sbernauer: 3962a16

crates/stackable-certs/src/ca/mod.rs
sbernauer: I don't know why https://github.com/stackabletech/operator-rs/blob/34588a2bed6a85701db6a25fab65648c8c65878c/crates/stackable-certs/src/keys/ecdsa.rs#L33 is fallible.

I'm basically treating this as error handling similar to snafu, where panicing is allowed. I have no clue why this could fail and I'm not comfortable making a claim why this would never fail for all future version that might come

crates/stackable-certs/src/ca/mod.rs
sbernauer: a81cf5b

crates/stackable-certs/src/ca/mod.rs
sbernauer: I personally prefer having "real" names for easier understanding.
The people reading this are developers of our operators and I think concepts are easier to understand with actual, recognizable values

Techassi: > The people reading this are developers of our operators and I think concepts are easier to understand with actual, recognizable values

They might be, this might also get used outside of Stackable and we should strive to make all of our crates usable outside of Stackable. There might not be many external users _now_, but there hopefully are a bunch in the future.

I won't block this PR on this, but we should definitely revisit this in the future if we leave it as is.

crates/stackable-certs/src/ca/mod.rs
Techassi: > I have no clue why this could fail and I'm not comfortable making a claim why this would never fail for all future version that might come

Then we should make it seem like we know by using `expect`. Using `unwrap` would be a better fit then.

But we could always say that signing key generation must not fail to be able to generate a leaf certificate.

crates/stackable-certs/src/ca/mod.rs
Techassi: Can we please lowercase the start of the message?

crates/stackable-certs/src/ca/mod.rs
NickLarsenNZ: ~hmm, I think how it was was good.~

I think we should add a Snafu variant.

The error can happen at runtime and is machine dependent (eg: could fail when there isn't enough entropy). I'm sure we can handle that gracefully.

---

> Using `unwrap` would be a better fit then.

I disagree, we lose the developers reasoning of why it should not fail.

crates/stackable-certs/src/ca/mod.rs
NickLarsenNZ: **non-blocker**:

I also think we should use example placeholders instead of a particular product name (same goes with test product versions unless behavior changes of course).

But perhaps instead of `example`, it can be `product` or `product-name`?

```suggestion
    const TEST_SAN: &str = "product-0.product.default.svc.cluster.local";
```

<details><summary>To elaborate on test versions...</summary>

In our operator tests we often have real product versions, and during the release preparation when searching for usages of removed/deprecated versions, there is often a lot of code changes - or the dilemma of whether to leave them as is.

So IMO, best to just use obvious placeholders like `0.0.0` or `000` (trino) - unless of course the tests check for things that differ between product versions.

</details>

crates/stackable-certs/src/ca/mod.rs
Techassi: > I think we should add a Snafu variant.

This particular line of code is part of a unit test. This would go against how we deal with errors in tests.

NickLarsenNZ: Oh right, I didn't catch that.

Then yes, we use expect, but that raises a good point.
Some cases it is simple to see how something can fail, and other times there can be many reasons, or it comes from deep within (eg: here, with the possible rng failure).

Perhaps here, we can just write something that gives reasonable assurance that this is expected to succeed, and hint to what might have gone wrong if it fails.

```suggestion
            .expect("Must be able to generate an RSA certificate");
```

or with the hint:

```suggestion
            .expect("Must be able to generate an RSA certificate. Perhaps there was an RNG failure");
```

We need to consider two things in the case of unit tests:
- We are showing that ordinarily, this should succeed
- If it fails, how can we assist the developer in tracking it down (was it their change, or was it some underlying issue)

crates/stackable-certs/src/ca/mod.rs
sbernauer: 69dced8

crates/stackable-certs/src/ca/mod.rs
sbernauer: Renamed Airflow with Product in 98acd6c, hope that works

crates/stackable-certs/src/ca/mod.rs
```diff
@@ -468,19 +497,59 @@ fn format_leaf_certificate_subject(name: &str, scope: &str) -> Result<Name> {
...
     }
 
     #[tokio::test]
     async fn ecdsa_key_generation() {
         let mut ca = CertificateAuthority::new_ecdsa().unwrap();
-        ca.generate_ecdsa_leaf_certificate("Airflow", "pod", Duration::from_secs(3600))
-            .unwrap();
+        let cert = ca
+            .generate_ecdsa_leaf_certificate("Product", "pod", [TEST_SAN], TEST_CERT_LIFETIME)
+            .expect("ecdsa certificate generation failed");
```
Techassi: note: We should use the same error message as above.

```suggestion
            .expect(
                "Must be able to generate an ECDSA certificate. Perhaps there was an RNG failure",
            );
```

Techassi (changes requested):
LGTM, one final comment.

PR #1059, opened 2025-06-18
fix(stackable-versioned): Emit an error when invalid items are added to the versioned module

NickLarsenNZ:
This improves DX by emitting an error when invalid items are added to the versioned module:

image

### DIFF
--- a/crates/stackable-versioned-macros/src/codegen/module.rs
+++ b/crates/stackable-versioned-macros/src/codegen/module.rs
@@ -108,7 +108,14 @@ impl Module {
                         ),
                     }
                 }
-                _ => continue,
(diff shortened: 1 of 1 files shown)


maltesander (approved):
LGTM!

sbernauer:
Nice, thanks!
A changelog entry would be nice, especially as this is technically a breaking change

NickLarsenNZ:
> Nice, thanks! A changelog entry would be nice, especially as this is technically a breaking change

Ah yeah, I missed the changelog, but I don't see how it is a breaking change.

It _was_ broken (because any methods defined were thrown away, and therefore unusable). Now we get a warning.

sbernauer:
exactly, previously compiling code is now not compiling any more.
Many thanks, approved!

PR #1050, opened 2025-05-27
feat(stackable-versioned)!: Integrate with ConversionReviews

sbernauer:
This PR adds utilities to convert a CRD from one version into a different version. This is tightly integrated with CRD conversions, which send `ConversionReview`s to registered webhooks and return that review to the K8s apiserver.

## TLDR

- BREAKING: The `#[versioned(k8s(skip(merged_crd)))]` flag has been removed
- BREAKING: The version enum used in `merged_crd` is now suffixed with `Version`
- Add new `try_convert` function to convert objects received via a ConversionReview
- Add new `enable_tracing` option to `#[versioned(k8s(options(...)))]`

## Detailed Overview

There are multiple different parts to make this whole process work. All of these parts are automatically generated by the macro and for the user it is a single function call: `Foo::try_convert`. The individual parts are explained below.

> [!NOTE]
> The CRD spec currently does not contain any fields. During the development of the conversion mechanism, this is emitted to more easily focus on the newly generated code. It will eventually contain fields to fully test the implementation.

```rust
#[versioned(
    version(name = "v1alpha1"),
    version(name = "v1beta1"),
    version(name = "v1"),
    k8s(group = "foo.stackable.tech")
)]
#[derive(Clone, Debug, CustomResource, Deserialize, Serialize, JsonSchema)]
pub struct FooSpec {}
```

### Updated Version Enum

The generated code tries to abstract away as much of the underlying code as possible. For this use-case, a top-level version enum exists. Before this PR, it only provided an easy way to produce a single, merged CRD based on all defined CRD versions via `Foo::merged_crd(Foo::V1)`. The variants described the available versions and forced users to use well-defined versions instead of `&str` which can be anything (eg. an invalid version).

This PR changes the inner workings of this enum. Instead of only providing plain variants without any data, each variant now holds the appropriate version of the custom resource.

```rust
// Before
pub enum Foo {
    V1Alpha1,
    V1Beta1,
    V1,
}

// After
pub enum Foo {
    V1Alpha1(v1alpha1::Foo),
    V1Beta1(v1beta1::Foo),
    V1(v1::Foo),
}
```

This change enables us to directly deserialize an object of a custom resource into the enum. This then enables us to generate further (in addition to `merged_crd`) top-level abstractions on the enum.

With this change it is not possible to use `Self` to specify the stored apiversion via `Foo::merged_crd()`. Instead, a new enum for this purpose only is generated.

```rust
pub enum FooVersion {
    V1Alpha1,
    V1Beta1,
[...]

PR #1061, opened 2025-06-23
fix(stackable-versioned): Validate kind, desired and current version

sbernauer:
# Motivation

#1056 will also fix at least some of the bugs, but I would prefer to fix the bugs fast and have tests.
I'm happy that my code gets overwritten by #1056 as longs as the tests pass.
By fixing the problems we can continue on working an the actual conversion HTTPS webhook stuff.

# Description

Fix some problems in the (brand new) conversions. And most important: Re-add tests that temporarily have been removed from https://github.com/stackabletech/operator-rs/pull/1050

* check if we know about the desired version, error if not. Previously we would noop for unknown desired versions.
* k8s sends us `test.stackable.tech/v1alpha1` as apiVersion, *not* `v1alpha1`
* Check that the sent k8s kind is the one we expect

PS: The diff is so big because lots of tests (snapshots) changed. Actual code change is only a small portion.


crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,37 @@ impl Struct {
...
+                fn from_json_value(object: #serde_json_path::Value) -> ::std::result::Result<Self, #parse_object_error> {
+                    let object_kind = object
+                        .get("kind")
+                        .ok_or_else(|| #parse_object_error::FieldMissing{ field: "kind".to_string() })?
+                        .as_str()
+                        .ok_or_else(|| #parse_object_error::FieldNotStr{ field: "kind".to_string() })?;
+
+                    // Note(@sbernauer): One could argue we don't need to check the kind, but this
+                    // is a nice sanity check. If for *some* reason a unexpected kind ends up at a
+                    // conversion, the problem might be very hard to spot without this.
```
NickLarsenNZ: I think this should be rewritten to only explain why we do the check and that there is no good reason to remove it. Eg:

```suggestion
                    // Note(@sbernauer): The kind must be checked here, because it is
                    // possible for the wrong object to be deserialized.
                    // Checking here stops us assuming the kind is correct and
                    // accidentally updating upgrade/downgrade information in the
                    // status in a later step.
```

NickLarsenNZ: While we check the _kind_, we should also check the _group_ matches.

That might take a bit more effort since we'd need to parse it from the `apiVersion` (could split_once on `/`). Do we assert that all versioned things have a group?

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -429,43 +466,22 @@ impl Struct {
             )
                 -> ::std::result::Result<::std::vec::Vec<#serde_json_path::Value>, #convert_object_error>
             {
+                let desired_api_version = #version_enum_ident::try_from_api_version(desired_api_version)?;
+                // let desired_api_version: #version_enum_ident = desired_api_version.try_into()?;
```
NickLarsenNZ: I guess this can go now:

```suggestion
```

sbernauer: Sure thing! I didn't push the change :facepalm: 95a10ae

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: > While we check the kind, we should also check the group matches.

Regarding the group, if you look at this
```json
{
    "kind": "ConversionReview",
    "apiVersion": "apiextensions.k8s.io/v1",
    "request": {
        "uid": "c4e55572-ee1f-4e94-9097-28936985d45f",
        "desiredAPIVersion": "test.stackable.tech/v1alpha1",
        "objects": [
            {
                "apiVersion": "test.stackable.tech/v1alpha1",
                "kind": "Person",
                "metadata": {
                    "name": "sbernauer",
                    "namespace": "default",
                    "uid": "d41e2019-5de3-409c-a7b2-0799ecb95e4b"
                },
                "spec": {
                    "username": "sbernauer"
                }
            },
        ]
    }
}
```

I can only find the group in apiVersions. And these are already checked (both current and desired). This PR changes that we only accept test.stackable.tech/v1alpha1 and not v1alpha1

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,39 @@ impl Struct {
...
                 {
-                    #kube_core_path::crd::merge_crds(vec![#(#crd_fns),*], stored_apiversion.as_str())
+                    #kube_core_path::crd::merge_crds(vec![#(#crd_fns),*], stored_apiversion.as_version_str())
                 }
 
                 #convert_method
 
-                fn from_json_value(value: #serde_json_path::Value) -> ::std::result::Result<Self, #parse_object_error> {
-                    let api_version = value
+                fn from_json_value(object: #serde_json_path::Value) -> ::std::result::Result<Self, #parse_object_error> {
```
Techassi: note: I think we should stick with the parameter name `value` here. In my mind, this can only be called an object _after_ it has been parsed as such. Also, the function name suggests that this parameter should be called `value`.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,39 @@ impl Struct {
...
+                    // possible for the wrong object to be deserialized.
+                    // Checking here stops us assuming the kind is correct and
+                    // accidentally updating upgrade/downgrade information in the
+                    // status in a later step.
+                    if object_kind != stringify!(#enum_ident) {
+                        return Err(#parse_object_error::UnexpectedObjectKind{
+                            kind: object_kind.to_string(),
+                            supported_kind: stringify!(#enum_ident).to_string(),
+                        });
+                    }
```
Techassi: sugestion: Directly use a string instead of relying on `stringify!`.

```suggestion
                    // NOTE(@sbernauer): The kind must be checked here, because it is possible for
                    // the wrong object to be deserialized. Checking here stops us assuming the kind
                    // is correct and accidentally updating upgrade/downgrade information in the
                    // status in a later step.
                    if object_kind != #enum_ident_string {
                        return Err(#parse_object_error::UnexpectedKind{
                            kind: object_kind.to_owned(),
                            expected: #enum_ident_string.to_owned(),
                        });
                    }
```

This needs `enum_ident_string`, which can be declared in line 131 like:

```rust
let enum_ident_string = &self.common.idents.kubernetes.to_string();
```

crates/stackable-versioned/src/k8s.rs
```diff
@@ -47,17 +47,25 @@ fn raw_object_schema(_: &mut schemars::r#gen::SchemaGenerator) -> Schema {
...
     #[snafu(display("failed to deserialize object from JSON"))]
     Deserialize { source: serde_json::Error },
+
+    #[snafu(display(
+        "unexpected object kind {kind:?}. The conversion can only convert objects of kind {supported_kind:?}"
+    ))]
+    UnexpectedObjectKind {
+        kind: String,
+        supported_kind: String,
+    },
```
Techassi: suggestion: Rename the error variant, the second parameter and slightly adjust the error message.

The error is called `ParseObjectError` and as such it is implied we are talking about errors related to object parsing. As such, I dropped the term "Object" from the variant.

I would avoid mentioning the term "conversion" in the error message, as this code can also be used outside of conversions.

```suggestion
    #[snafu(display("unexpected object kind {kind:?}, expected {expected:?}"))]
    UnexpectedKind { kind: String, expected: String },
```

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,39 @@ impl Struct {
...
                 #convert_method
 
-                fn from_json_value(value: #serde_json_path::Value) -> ::std::result::Result<Self, #parse_object_error> {
-                    let api_version = value
+                fn from_json_value(object: #serde_json_path::Value) -> ::std::result::Result<Self, #parse_object_error> {
+                    let object_kind = object
+                        .get("kind")
+                        .ok_or_else(|| #parse_object_error::FieldMissing{ field: "kind".to_string() })?
+                        .as_str()
+                        .ok_or_else(|| #parse_object_error::FieldNotStr{ field: "kind".to_string() })?;
```
Techassi: suggestion: Use `to_owned` to better indicate what is done here.

```suggestion
                    let object_kind = object
                        .get("kind")
                        .ok_or_else(|| #parse_object_error::FieldMissing{ field: "kind".to_owned() })?
                        .as_str()
                        .ok_or_else(|| #parse_object_error::FieldNotStr{ field: "kind".to_owned() })?;
```

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,39 @@ impl Struct {
...
+                        return Err(#parse_object_error::UnexpectedObjectKind{
+                            kind: object_kind.to_string(),
+                            supported_kind: stringify!(#enum_ident).to_string(),
+                        });
+                    }
+
+                    let api_version = object
                         .get("apiVersion")
-                        .ok_or_else(|| #parse_object_error::FieldNotPresent)?
+                        .ok_or_else(|| #parse_object_error::FieldMissing{ field: "apiVersion".to_string() })?
```
Techassi: suggestion: Use `to_owned` to better indicate what is done here.

```suggestion
                        .ok_or_else(|| #parse_object_error::FieldMissing{ field: "apiVersion".to_owned() })?
```

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,39 @@ impl Struct {
...
                         .as_str()
-                        .ok_or_else(|| #parse_object_error::FieldNotStr)?;
+                        .ok_or_else(|| #parse_object_error::FieldNotStr{ field: "apiVersion".to_string() })?;
```
Techassi: suggestion: Use `to_owned` to better indicate what is done here.

```suggestion
                        .ok_or_else(|| #parse_object_error::FieldNotStr{ field: "apiVersion".to_owned() })?;
```

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -173,21 +174,39 @@ impl Struct {
...
 
                     let object = match api_version {
-                        #(#api_versions | #variant_strings => {
-                            let object = #serde_json_path::from_value(value)
+                        #(#api_versions => {
+                            let object = #serde_json_path::from_value(object)
```
Techassi: note: As stated above, I prefer `value` here. The `serde_json::from_value` function again suggests that the parameter should be named `value`.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -241,17 +268,33 @@ impl Struct {
...
+                }
+
+                pub fn try_from_api_version(api_version: &str) -> Result<Self, #convert_object_error> {
+                    match api_version {
+                        #(#api_versions => Ok(Self::#variant_idents)),*,
+                        _ => Err(#convert_object_error::DesiredApiVersionUnknown {
+                            unknown_desired_api_version: api_version.to_string(),
+                        }),
+                    }
+                }
```
Techassi: suggestion: Remove the `try_` prefix and use separate error type.

The function name doesn't need to indicate this function is fallible, because it already indicates that via the function signature (it returns a Result). As such, the `try_` prefix can be removed.

Also, the returned error seems to indicate that this function is only used in the context of a conversion, which doesn't have to be the case. I instead suggest designing the errors like this:

```rust
#[derive(Debug, Snafu)]
pub enum ConvertObjectError {
    // ...

    #[snafu(display("unknown desired API version"))]
    UnknownDesiredApiVersion { source: ParseVersionError },
}

#[derive(Debug, Snafu)]
#[snafu(display("failed to parse {input:?} as version))]
pub struct ParseVersionError {
    input: String,
}
```

The suggested changes result in:

```suggestion
                pub fn from_api_version(api_version: &str) -> Result<Self, ParseVersionError> {
                    match api_version {
                        #(#api_versions => Ok(Self::#variant_idents)),*,
                        _ => Err(ParseVersionError {
                            input: api_version.to_owned(),
                        }),
                    }
                }
```

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -434,13 +478,11 @@ impl Struct {
...
-                        // If no match arm matches, this is a noop. This is the case if the desired
-                        // version matches the current object api version.
-                        // NOTE (@Techassi): I'm curious if this will ever happen? In theory the K8s
-                        // apiserver should never send such a conversion review.
-                        _ => converted_objects.push(object),
+                    match (current_object, &desired_api_version) {
+                        #(#conversion_match_arms,)*
+                        // We explicitly list the remaining no-op cases, so the compiler ensures we
+                        // did not miss a conversion.
+                        #(#noop_match_arms,)*
```
Techassi: note: I'd like to avoid generating a bunch of match arms which should basically never be hit. The only benefit of this is compiler checking, which is basically already covered by our snapshot tests.

If we don't generate _all_ non-noop combinations, the compiler error doesn't help. The snapshot tests are there to ensure this never happens.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -470,7 +514,10 @@ impl Struct {
                 let current_object_version_string = &start.inner.to_string();
 
                 let desired_object_version = path.last().expect("the path always contains at least one element");
-                let desired_object_version_string = desired_object_version.inner.to_string();
+                let desired_object_api_version_string = format!(
+                    "{group}/{desired_object_version}",
+                    desired_object_version= desired_object_version.inner
+                );
```
Techassi:
```suggestion
                let desired_object_api_version_string = format!(
                    "{group}/{desired_object_version}",
                    desired_object_version = desired_object_version.inner
                );
```

crates/stackable-versioned/src/k8s.rs
```diff
@@ -47,17 +47,25 @@ fn raw_object_schema(_: &mut schemars::r#gen::SchemaGenerator) -> Schema {
...
 #[derive(Debug, Snafu)]
 pub enum ParseObjectError {
-    #[snafu(display(r#"failed to find "apiVersion" field"#))]
-    FieldNotPresent,
+    #[snafu(display("the field {field:?} is missing"))]
+    FieldMissing { field: String },
 
-    #[snafu(display(r#"the "apiVersion" field must be a string"#))]
-    FieldNotStr,
+    #[snafu(display(r#"the field {field:?} must be a string"#))]
```
Techassi: note: As already done above, this raw string can be removed.

```suggestion
    #[snafu(display("the field {field:?} must be a string"))]
```

crates/stackable-versioned/src/lib.rs
```diff
@@ -17,3 +17,6 @@ pub use stackable_versioned_macros::versioned;
 mod k8s;
 #[cfg(feature = "k8s")]
 pub use k8s::*;
+
+#[cfg(test)]
+mod tests;
```
Techassi: note: I would like to avoid this. Instead, we want automatic support for integration tests by moving the "tests" folder one layer up (out of "src").

crates/stackable-versioned/src/tests/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+fn run_for_file(path: &Path) -> (ConversionReview, ConversionReview) {
+    let request: ConversionReview =
+        serde_json::from_reader(File::open(path).expect("failed to open test file"))
+            .expect("failed to parse ConversionReview from test file");
+    let response = Person::try_convert(request.clone());
+
+    (request, response)
+}
```
Techassi: note: Similar to `stackable-versioned-macros`, I would like this function to live in a `test_utils` module (which is gated behind `#[cfg(test)]`.

I suggest to adjust the name to better indicate what this function does and adding a dedicated error enum similar to `expand_from_file`.

```suggestion
fn convert_via_file(path: &Path) -> (ConversionReview, ConversionReview) {
    let request: ConversionReview =
        serde_json::from_reader(File::open(path).expect("failed to open test file"))
            .expect("failed to parse ConversionReview from test file");
    let response = Person::try_convert(request.clone());

    (request, response)
}
```

_Ideally, this function would be generic over the kind we want to convert. This is however currently not possible but worth a thought in the future._

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -434,13 +478,11 @@ impl Struct {
...
                         .map_err(|source| #convert_object_error::Parse { source })?;
 
-                    match (current_object, desired_api_version) {
-                        #(#match_arms,)*
-                        // If no match arm matches, this is a noop. This is the case if the desired
-                        // version matches the current object api version.
-                        // NOTE (@Techassi): I'm curious if this will ever happen? In theory the K8s
-                        // apiserver should never send such a conversion review.
-                        _ => converted_objects.push(object),
+                    match (current_object, &desired_api_version) {
```
Techassi: note: Using a reference here seems a little weird. We could instead add `#[derive(Copy, Clone)]` to the version enum, to be able to cheaply copy the enum, which is also suggested by the [docs](https://doc.rust-lang.org/stable/std/marker/trait.Copy.html#when-should-my-type-be-copy).

Would be interesting to see if there are any performance differences. [This StackOverflow post](https://stackoverflow.com/a/44695401) seems to suggest that copying is smaller in memory but there will be likely no performance difference.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: See 58b330b

crates/stackable-versioned/src/k8s.rs
sbernauer: d5a3109

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: error added in 4364c53

sbernauer: As for the name I mimicked https://doc.rust-lang.org/std/convert/trait.TryFrom.html. I didn't want to commit to the trait, as there might be cases we need to parse the versions (without the group).
`try_from_api_version` sounds like a logical evolution of `try_from` to me

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: I already had all of this thoughts and though about renaming the function to smh like `try_from_json_object) but didn't want to get into bikeshedding. But here we are :)

This function expects you to pass a k8s object to it, not an arbitrary JSON value such as f32 or String. I think we should make this clear in the name, I would therefore propose to name it  `fn try_from_json_object(object: #serde_json_path::Value)` or `fn try_from_json_k8s_object(k8s_object: #serde_json_path::Value)` (the try_ prefix is already under discussion in https://github.com/stackabletech/operator-rs/pull/1061#discussion_r2163235155)

crates/stackable-versioned/src/lib.rs
sbernauer: Moved in 8f77619

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
Techassi: Could you please also change `to_string` to `to_owned`?

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
Techassi: > error added in [4364c53](https://github.com/stackabletech/operator-rs/commit/4364c539840a0ccee37c1c47501b4378ec4ba861)

Could you please also change `to_string` to `to_owned`?

> As for the name I mimicked [doc.rust-lang.org/std/convert/trait.TryFrom.html](https://doc.rust-lang.org/std/convert/trait.TryFrom.html). I didn't want to commit to the trait, as there might be cases we need to parse the versions (without the group).
try_from_api_version sounds like a logical evolution of try_from to me

The `TryFrom` trait is meant to be used for fallible conversion. The `try_` prefix is used here to differentiate between this trait and the non-fallible `From` trait.

We are however not dealing with a conversion here. We are instead dealing with parsing (a string to an enum variant). In this case, the `FromStr` trait would be the go-to. This trait however can only be used if there is a **single** way to parse the string.

Therefore we have the following options:

- Support parsing from both version and apiVersion - the `FromStr` trait cannot be used, stick with `from_version` and `from_api_version`.
- Support parsing only from apiVersion - the `FromStr` trait can be used

crates/stackable-versioned/src/lib.rs
Techassi: suggestion: I would additionally drop the `fixtures` part from the folder structure. Then we have the same structure as in `stackable-versioned-macros`.

suggestion: I would also rename the `tests/integration_test.rs` file to `tests/conversion.rs` because we are testing the conversion. There might be other kind of _integration tests_ in the future.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
Techassi: I would be fine with `fn try_from_json_object(object_value: #serde_json_path::Value)` and internally clearly separate between the "raw" object and the successfully deserialized "object" we return in the end.

In my opinion the `try_` prefix doesn't provide a whole lot here. The function clearly indicates that it is fallible by returning a `Result`.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: cb3dbfe

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
NickLarsenNZ: > I can only find the group in apiVersions

Correct, it is the first element when split by `/`.

> This PR changes that we only accept test.stackable.tech/v1alpha1 and not v1alpha1

Sounds like it is safe.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: That sounds like a good compromise! 119a845

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: See https://github.com/stackabletech/operator-rs/pull/1061#discussion_r2163982049

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: Yeah, you are right => 83e2438
The to_owned was already done in bulk with a different commit

crates/stackable-versioned/src/lib.rs
sbernauer: works for me

7400d39 now has files such as `crates/stackable-versioned/tests/inputs/conversions/fail/wrong_kind.json`

crates/stackable-versioned/src/tests/mod.rs
sbernauer: I tried putting it in `crates/stackable-versioned/src/test_utils.rs`, the problem is that we don't know Person in there...
And currently we don't have a trait for the `T::try_convert(request.clone())` function. So I'm missing how we can move the `convert_via_file` function

sbernauer: renamed in b7080ac

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: 3f9e21f

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: > The only benefit of this is compiler checking, which is basically already covered by our snapshot tests.

That' the benefit I'm after. We have the downside of having like 20 lines of Rust code, but the upside that the compiler ensures that we explicitly reasoned about every version combination. I personally preferred the way I did it originally, rendering all permutations in a single loop ;) But I'm also ok with this solution.

I'm not sure if I would trust the snapshot tests too much. The current main branch relies on a "catch all" branch, which introduced the "unknown version bug". I prefer better safe than sorry.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
Techassi: Ah sorry, I meant `from_json_object`. Otherwise the linked commit above is fine. I'm also happy to change it myself.

crates/stackable-versioned/src/tests/mod.rs
Techassi: > And currently we don't have a trait for the T::try_convert(request.clone()) function. So I'm missing how we can move the convert_via_file function

Yeah that's a bummer, but a thing we can improve on in the future. I noted it down in the [tracking issue](https://github.com/stackabletech/issues/issues/642).

crates/stackable-versioned/src/lib.rs
Techassi: I like the path `.../inputs/conversions/fail/...`.

I think that we should update the name of the test file to `conversions.rs` as well.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
sbernauer: 90884aa

crates/stackable-versioned/src/lib.rs
sbernauer: 426e1b7

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
Techassi: > I personally preferred the way I did it originally, rendering all permutations in a single loop ;)

I explicitly only generate the non-noop permutations, because all other noops can be handled via a catch all.

> which introduced the "unknown version bug"

This bug was not introduced by having a catch-all branch, but instead how we handled the desired api version. This PR now correctly catches invalid desired and current api versions. As such, I don't see the need to match every noop, because it is now impossible to get passed validation/parsing with an invalid version.

sbernauer: pros:
* The compiler ensures that we explicitly reasoned about every version combination. Either by providing a conversion or using NOOP.

cons:
* Introduce ~30 lines of code

neutral:
* We have tests for `conversion_paths`, this is just a second safeguard.

sbernauer: A small improvement I did is to only generate a single match statement in 0bf0039
It now looks like
```rust
for object in objects {
    let current_object = Self::from_json_object(object.clone())
        .map_err(|source| ::stackable_versioned::ConvertObjectError::Parse { source })?;
    match (current_object, desired_api_version) {
        (Self::V1Alpha1(__sv_person), PersonVersion::V1Alpha2) => {
        (Self::V1Alpha1(__sv_person), PersonVersion::V1Beta1) => {
        (Self::V1Alpha1(__sv_person), PersonVersion::V2) => {
        (Self::V1Alpha1(__sv_person), PersonVersion::V3) => {
        (Self::V1Alpha2(__sv_person), PersonVersion::V1Alpha1) => {
        (Self::V1Alpha2(__sv_person), PersonVersion::V1Beta1) => {
        (Self::V1Alpha2(__sv_person), PersonVersion::V2) => {
        (Self::V1Alpha2(__sv_person), PersonVersion::V3) => {
        (Self::V1Beta1(__sv_person), PersonVersion::V1Alpha1) => {
        (Self::V1Beta1(__sv_person), PersonVersion::V1Alpha2) => {
        (Self::V1Beta1(__sv_person), PersonVersion::V2) => {
        (Self::V1Beta1(__sv_person), PersonVersion::V3) => {
        (Self::V2(__sv_person), PersonVersion::V1Alpha1) => {
        (Self::V2(__sv_person), PersonVersion::V1Alpha2) => {
        (Self::V2(__sv_person), PersonVersion::V1Beta1) => {
        (Self::V2(__sv_person), PersonVersion::V3) => {
        (Self::V3(__sv_person), PersonVersion::V1Alpha1) => {
        (Self::V3(__sv_person), PersonVersion::V1Alpha2) => {
        (Self::V3(__sv_person), PersonVersion::V1Beta1) => {
        (Self::V3(__sv_person), PersonVersion::V2) => {
        // NOOP branches
        (Self::V1Alpha1(_), PersonVersion::V1Alpha1)
        | (Self::V1Alpha2(_), PersonVersion::V1Alpha2)
        | (Self::V1Beta1(_), PersonVersion::V1Beta1)
        | (Self::V2(_), PersonVersion::V2)
        | (Self::V3(_), PersonVersion::V3) => converted_objects.push(object),
    }
}
```

sbernauer: Ok, we are now down to four lines of code the safety check costs us:
```rust
let version_idents = versions.iter().map(|v| &v.idents.variant);
```
and
```rust
#(
    (Self::#version_idents(_), #version_enum_ident::#version_idents)
)|* => converted_objects.push(object)
```

I think the 4 lines are worth the double safety but yeah

sbernauer: I just want this PR merged => ab02a5e

Techassi (approved):

PR #1065, opened 2025-06-26
fix(stackable-versioned): Emit correct match arms for version enum

Techassi:
This is rather curious because the previous generated code should result in compiler errors. The expanded code does not error, but manually copy+pasting the generated code reveals the compiler error as expected:

```
error[E0170]: pattern binding `V1Alpha1` is named the same as one of the variants of the type `FooVersion`
   --> crates/stackable-versioned-macros/tests/test.rs:324:13
    |
324 |             V1Alpha1 => "v1alpha1",
    |             ^^^^^^^^ help: to match on the variant, qualify the path: `FooVersion::V1Alpha1`
    |
    = note: `#[deny(bindings_with_variant_name)]` on by default
``` 

The incorrect match resulted in `as_str` to always return the earliest version (usually `v1alpha1`). This resulted in the wrong version to be marked as stored when calling `Foo::merged_crd`.

### DIFF
--- a/crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
+++ b/crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
@@ -286,13 +286,13 @@ impl Struct {
 
                 pub fn as_api_version_str(&self) -> &str {
                     match self {
-                        #(#variant_idents => #api_versions),*
+                        #(#enum_ident::#variant_idents => #api_versions),*
                     }
                 }
 
                 pub fn from_api_version(api_version: &str) -> Result<Self, #unknown_desired_api_version_error> {
                     match api_version {
-                        #(#api_versions => Ok(Self::#variant_idents)),*,
+                        #(#api_versions => Ok(#enum_ident::#variant_idents)),*,
                         _ => Err(#unknown_desired_api_version_error {
                             api_version: api_version.to_owned(),
                         }),
--- a/crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
+++ b/crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
@@ -280,7 +280,7 @@ impl Struct {
             impl #enum_ident {
                 pub fn as_version_str(&self) -> &str {
                     match self {
-                        #(#variant_idents => #variant_strings),*
+                        #(#enum_ident::#variant_idents => #variant_strings),*
                     }
                 }
 
--- a/crates/stackable-versioned/tests/merged_crd.rs
+++ b/crates/stackable-versioned/tests/merged_crd.rs
@@ -1,0 +1,22 @@
+use crate::person::{Person, PersonVersion};
+
+mod person;
+
+#[test]
+fn stored_apiversion() {
+    let merged_crd = Person::merged_crd(PersonVersion::V2).expect("the CRDs must be mergeable");
+
+    let all_storage_fields_correct =
+        merged_crd
+            .spec
+            .versions
+            .iter()
+            .enumerate()
+            .all(
+                |(idx, crd)| {
+                    if idx == 0 { crd.storage } else { !crd.storage }
+                },
+            );
+
+    assert!(all_storage_fields_correct);
+}
--- a/crates/stackable-versioned/tests/merged_crd.rs
+++ b/crates/stackable-versioned/tests/merged_crd.rs
@@ -4,37 +4,19 @@ mod person;
 
 #[test]
 fn stored_apiversion() {
-    let desired_stored_apiversion = PersonVersion::V2;
+    let stored_apiversion = PersonVersion::V2;
 
-    let merged_crd =
-        Person::merged_crd(desired_stored_apiversion).expect("the CRDs must be mergeable");
+    let merged_crd = Person::merged_crd(stored_apiversion).expect("the CRDs must be mergeable");
 
-    // First, we ensure that all storage fields have the correct value.
-    let all_storage_fields_correct =
-        merged_crd
-            .spec
-            .versions
-            .iter()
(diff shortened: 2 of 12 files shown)


crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
```diff
@@ -280,19 +280,19 @@ impl Struct {
...
                     match self {
-                        #(#variant_idents => #api_versions),*
+                        #(#enum_ident::#variant_idents => #api_versions),*
                     }
                 }
 
                 pub fn from_api_version(api_version: &str) -> Result<Self, #unknown_desired_api_version_error> {
                     match api_version {
-                        #(#api_versions => Ok(Self::#variant_idents)),*,
+                        #(#api_versions => Ok(#enum_ident::#variant_idents)),*,
```
sbernauer, line 295: Any particular reason you prefer `#enum_ident` over `Self`?

crates/stackable-versioned/tests/merged_crd.rs
```diff
@@ -0,0 +1,22 @@
...
+fn stored_apiversion() {
+    let merged_crd = Person::merged_crd(PersonVersion::V2).expect("the CRDs must be mergeable");
+
+    let all_storage_fields_correct =
+        merged_crd
+            .spec
+            .versions
+            .iter()
+            .enumerate()
+            .all(
```
sbernauer: I'm a bit confused. This seems to check that the first version in the list is storage and the rest are !storage.
Shouldn't we check that *the one with name=v2* is the stored version, regardless of the position in the list?

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
Techassi: I used the Rust book as an orientation. They always use the enum name instead of `Self`. This is also an unwritten rule for idiomatic Rust.

See [book](https://doc.rust-lang.org/stable/book/ch06-02-match.html).

crates/stackable-versioned/tests/merged_crd.rs
Techassi: I would say no. The upstream `merge_crds` function always uses the stored version as the root version and therefore puts it at the top. In combination with the second assertion, we make sure that the first version (which is the stored version) is the one we expect.

crates/stackable-versioned-macros/src/codegen/container/struct/k8s.rs
NickLarsenNZ: I think the concrete ident should be used in generated code anyway.
But in person written code, I often use Self (mainly so that renames are easier to deal with).

sbernauer: Ok, works for me, thx

crates/stackable-versioned/tests/merged_crd.rs
sbernauer: Ok, I see (the second test is new).
Could we than please swap the test order and add the comment you just wrote, as it clarified things for me?

sbernauer: Although this seems like an implementation detail of kube to me. I guess k8s itself does not enforce any particular order

Techassi: Sure, I can swap them and add a comment.

Techassi: > Although this seems like an implementation detail of kube to me. I guess k8s itself does not enforce any particular order

No, as far as I know, this is kube-specific. I'm also happy to drop that part of the test, because as you said, it is an implementation detail.

sbernauer: I was thinking of something like
```rust
    let stored_apiversion = PersonVersion::V2;

    let merged_crd =
        Person::merged_crd(stored_apiversion).expect("the Person CRDs must be mergeable");
    let versions = merged_crd.spec.versions;

    // Ensure all versions are present
    assert_eq!(
        versions
            .iter()
            .map(|crd| crd.name.as_str())
            .collect::<HashSet<_>>(),
        HashSet::from(["v1alpha1", "v1alpha2", "v1beta1", "v2", "v3"])
    );

    // There must only be a single storage version: the configured one
    assert_eq!(
        versions
            .iter()
            .filter(|crd| crd.storage)
            .map(|crd| crd.name.as_str())
            .collect::<Vec<_>>(),
        [stored_apiversion.as_version_str()]
    );
```
What do you think about that one? This does not use lube-specifics and should be stable forever (tm)

Techassi: I feel like this over complicates the test and also tests conditions which `kube` should ensure (that all versions are present and only one is marked `storage: true`).

Our current test ensures that _our_ code does the correct thing.

sbernauer: I prefer more tests than less but ok.
Can we please at least test we have only a single stored version? This could be done using the following and should be similarly short
```rust
    // There must only be a single storage version: the configured one
    assert_eq!(
        merged_crd
            .spec
            .versions
            .iter()
            .filter(|crd| crd.storage)
            .map(|crd| crd.name.as_str())
            .collect::<Vec<_>>(),
        [stored_apiversion.as_version_str()]
    );
```

sbernauer: @Techassi suggested that we might add the testcases upstream. I wont block the PR on this

Techassi: I'll raise a PR upstream asap to add tests which assert that the merged CRD contains the right number of versions and that only a single version is marked with `storage: true`.

PR #1066, opened 2025-06-30
feat(webhook): Add functioning conversion webhook with cert rotation

sbernauer:
# Description

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/cli.rs
+++ b/crates/stackable-operator/src/cli.rs
@@ -163,7 +163,7 @@ pub enum Command<Run: Args = ProductOperatorRun> {
 /// Can be embedded into an extended argument set:
 ///
 /// ```rust
-/// # use stackable_operator::cli::{Command, ProductOperatorRun, ProductConfigPath};
+/// # use stackable_operator::cli::{Command, OperatorEnvironmentOpts, ProductOperatorRun, ProductConfigPath};
 /// use clap::Parser;
 /// use stackable_operator::namespace::WatchNamespace;
 /// use stackable_telemetry::tracing::TelemetryOptions;
@@ -176,7 +176,20 @@ pub enum Command<Run: Args = ProductOperatorRun> {
 ///     common: ProductOperatorRun,
 /// }
 ///
-/// let opts = Command::<Run>::parse_from(["foobar-operator", "run", "--name", "foo", "--product-config", "bar", "--watch-namespace", "foobar"]);
+/// let opts = Command::<Run>::parse_from([
+///     "foobar-operator",
+///     "run",
+///     "--name",
+///     "foo",
+///     "--product-config",
+///     "bar",
+///     "--watch-namespace",
+///     "foobar",
+///     "--operator-namespace",
+///     "stackable-operators",
+///     "--operator-service-name",
+///     "foo-operator",
+/// ]);
 /// assert_eq!(opts, Command::Run(Run {
 ///     name: "foo".to_string(),
 ///     common: ProductOperatorRun {
@@ -184,6 +197,10 @@ pub enum Command<Run: Args = ProductOperatorRun> {
 ///         watch_namespace: WatchNamespace::One("foobar".to_string()),
 ///         telemetry_arguments: TelemetryOptions::default(),
 ///         cluster_info_opts: Default::default(),
+///         operator_environment: OperatorEnvironmentOpts {
+///             operator_namespace: "stackable-operators".to_string(),
+///             operator_service_name: "foo-operator".to_string(),
+///         },
 ///     },
 /// }));
 /// ```
@@ -215,6 +232,9 @@ pub struct ProductOperatorRun {
     /// Provides a specific namespace to watch (instead of watching all namespaces)
     #[arg(long, env, default_value = "")]
     pub watch_namespace: WatchNamespace,
+
+    #[command(flatten)]
+    pub operator_environment: OperatorEnvironmentOpts,
 
     #[command(flatten)]
     pub telemetry_arguments: TelemetryOptions,
@@ -278,6 +298,18 @@ impl ProductConfigPath {
     }
 }
 
+#[derive(clap::Parser, Debug, PartialEq, Eq)]
+pub struct OperatorEnvironmentOpts {
+    /// The namespace the operator is running in, usually `stackable-operators`.
+    #[arg(long, env)]
+    pub operator_namespace: String,
+
+    /// The name of the service the operator is reachable at, usually
+    /// something like `<product>-operator`.
+    #[arg(long, env)]
+    pub operator_service_name: String,
+}
+
 #[cfg(test)]
 mod tests {
     use std::{env, fs::File};
@@ -292,6 +324,8 @@ mod tests {
     const DEPLOY_FILE_PATH: &str = "deploy_config_spec_properties.yaml";
     const DEFAULT_FILE_PATH: &str = "default_file_path_properties.yaml";
     const WATCH_NAMESPACE: &str = "WATCH_NAMESPACE";
+    const OPERATOR_NAMESPACE: &str = "OPERATOR_NAMESPACE";
+    const OPERATOR_SERVICE_NAME: &str = "OPERATOR_SERVICE_NAME";
 
     #[test]
     fn verify_cli() {
@@ -388,6 +422,10 @@ mod tests {
             "bar",
             "--watch-namespace",
             "foo",
+            "--operator-namespace",
+            "stackable-operators",
+            "--operator-service-name",
+            "foo-operator",
         ]);
         assert_eq!(
             opts,
@@ -396,11 +434,23 @@ mod tests {
                 watch_namespace: WatchNamespace::One("foo".to_string()),
                 cluster_info_opts: Default::default(),
                 telemetry_arguments: Default::default(),
+                operator_environment: OperatorEnvironmentOpts {
+                    operator_namespace: "stackable-operators".to_string(),
+                    operator_service_name: "foo-operator".to_string(),
+                }
             }
         );
 
         // no cli / no env
-        let opts = ProductOperatorRun::parse_from(["run", "--product-config", "bar"]);
+        let opts = ProductOperatorRun::parse_from([
+            "run",
+            "--product-config",
+            "bar",
+            "--operator-namespace",
+            "stackable-operators",
+            "--operator-service-name",
+            "foo-operator",
+        ]);
         assert_eq!(
             opts,
             ProductOperatorRun {
@@ -408,11 +458,18 @@ mod tests {
                 watch_namespace: WatchNamespace::All,
                 cluster_info_opts: Default::default(),
                 telemetry_arguments: Default::default(),
+                operator_environment: OperatorEnvironmentOpts {
+                    operator_namespace: "stackable-operators".to_string(),
+                    operator_service_name: "foo-operator".to_string(),
+                }
             }
         );
 
         // env with namespace
         unsafe { env::set_var(WATCH_NAMESPACE, "foo") };
+        unsafe { env::set_var(OPERATOR_SERVICE_NAME, "foo-operator") };
+        unsafe { env::set_var(OPERATOR_NAMESPACE, "stackable-operators") };
+
         let opts = ProductOperatorRun::parse_from(["run", "--product-config", "bar"]);
         assert_eq!(
             opts,
@@ -421,6 +478,10 @@ mod tests {
                 watch_namespace: WatchNamespace::One("foo".to_string()),
                 cluster_info_opts: Default::default(),
                 telemetry_arguments: Default::default(),
+                operator_environment: OperatorEnvironmentOpts {
+                    operator_namespace: "stackable-operators".to_string(),
+                    operator_service_name: "foo-operator".to_string(),
+                }
             }
         );
     }
--- a/crates/stackable-webhook/src/lib.rs
+++ b/crates/stackable-webhook/src/lib.rs
@@ -56,6 +56,9 @@ pub trait WebhookHandler<Req, Res> {
     fn call(self, req: Req) -> Res;
 }
 
+/// A result type alias with the library-level [`Error`] type as the default error type.
+pub type Result<T, E = WebhookError> = std::result::Result<T, E>;
+
 #[derive(Debug, Snafu)]
 pub enum WebhookError {
     #[snafu(display("failed to create TLS server"))]
@@ -120,7 +123,7 @@ impl WebhookServer {
         router: Router,
         options: Options,
         subject_alterative_dns_names: Vec<String>,
-    ) -> Result<(Self, mpsc::Receiver<Certificate>), WebhookError> {
+    ) -> Result<(Self, mpsc::Receiver<Certificate>)> {
         tracing::trace!("create new webhook server");
 
         // TODO (@Techassi): Make opt-in configurable from the outside
@@ -154,7 +157,7 @@ impl WebhookServer {
     /// Runs the Webhook server and sets up signal handlers for shutting down.
     ///
     /// This does not implement graceful shutdown of the underlying server.
-    pub async fn run(self) -> Result<(), WebhookError> {
+    pub async fn run(self) -> Result<()> {
         let future_server = self.run_server();
         let future_signal = async {
             let mut sigint = signal(SignalKind::interrupt()).expect("create SIGINT listener");
@@ -187,7 +190,7 @@ impl WebhookServer {
 
     /// Runs the webhook server by creating a TCP listener and binding it to
     /// the specified socket address.
-    async fn run_server(self) -> Result<(), WebhookError> {
+    async fn run_server(self) -> Result<()> {
         tracing::debug!("run webhook server");
 
         self.tls_server.run().await.context(RunTlsServerSnafu)
--- a/crates/stackable-webhook/src/lib.rs
+++ b/crates/stackable-webhook/src/lib.rs
@@ -60,10 +60,10 @@ pub trait WebhookHandler<Req, Res> {
 #[derive(Debug, Snafu)]
 pub enum WebhookError {
     #[snafu(display("failed to create TLS server"))]
-    CreateTlsServer { source: tls::Error },
+    CreateTlsServer { source: tls::TlsServerError },
 
     #[snafu(display("failed to run TLS server"))]
-    RunTlsServer { source: tls::Error },
+    RunTlsServer { source: tls::TlsServerError },
 }
 
 /// A ready-to-use webhook server.
@@ -123,8 +123,7 @@ impl WebhookServer {
         router: Router,
         options: Options,
         subject_alterative_dns_names: Vec<String>,
-        cert_tx: mpsc::Sender<Certificate>,
-    ) -> Result<Self, WebhookError> {
+    ) -> Result<(Self, mpsc::Receiver<Certificate>), WebhookError> {
         tracing::trace!("create new webhook server");
 
         // TODO (@Techassi): Make opt-in configurable from the outside
@@ -147,16 +146,12 @@ impl WebhookServer {
             .route("/health", get(|| async { "ok" }));
 
         tracing::debug!("create TLS server");
-        let tls_server = TlsServer::new(
-            options.socket_addr,
-            router,
-            subject_alterative_dns_names,
-            cert_tx,
-        )
-        .await
-        .context(CreateTlsServerSnafu)?;
+        let (tls_server, cert_rx) =
+            TlsServer::new(options.socket_addr, router, subject_alterative_dns_names)
+                .await
+                .context(CreateTlsServerSnafu)?;
 
-        Ok(Self { tls_server })
+        Ok((Self { tls_server }, cert_rx))
     }
 
     /// Runs the Webhook server and sets up signal handlers for shutting down.
--- a/crates/stackable-webhook/src/lib.rs
+++ b/crates/stackable-webhook/src/lib.rs
@@ -56,7 +56,7 @@ pub trait WebhookHandler<Req, Res> {
     fn call(self, req: Req) -> Res;
 }
 
-/// A result type alias with the library-level [`Error`] type as the default error type.
+/// A result type alias with the [`WebhookError`] type as the default error type.
 pub type Result<T, E = WebhookError> = std::result::Result<T, E>;
 
 #[derive(Debug, Snafu)]
--- a/crates/stackable-webhook/src/lib.rs
+++ b/crates/stackable-webhook/src/lib.rs
@@ -1,6 +1,6 @@
 //! Utility types and functions to easily create ready-to-use webhook servers
 //! which can handle different tasks, for example CRD conversions. All webhook
-//! servers use HTTPS by defaultThis library is fully compatible with the
+//! servers use HTTPS by default. This library is fully compatible with the
 //! [`tracing`] crate and emits debug level tracing data.
 //!
 //! Most users will only use the top-level exported generic [`WebhookServer`]
@@ -10,25 +10,30 @@
 //! ```
 //! use stackable_webhook::{WebhookServer, Options};
 //! use axum::Router;
+//! use tokio::sync::mpsc;
 //!
+//! let (cert_tx, _cert_rx) = mpsc::channel(1);
 //! let router = Router::new();
-//! let server = WebhookServer::new(router, Options::default());
+//! let server = WebhookServer::new(router, Options::default(), vec![], cert_tx);
 //! ```
 //!
 //! For some usages, complete end-to-end [`WebhookServer`] implementations
-//! exist. One such implementation is the [`ConversionWebhookServer`][1]. The
-//! only required parameters are a conversion handler function and [`Options`].
+//! exist. One such implementation is the [`ConversionWebhookServer`][1].
 //!
 //! This library additionally also exposes lower-level structs and functions to
-//! enable complete controll over these details if needed.
+//! enable complete control over these details if needed.
 //!
 //! [1]: crate::servers::ConversionWebhookServer
 use axum::{Router, routing::get};
 use futures_util::{FutureExt as _, pin_mut, select};
 use snafu::{ResultExt, Snafu};
 use stackable_telemetry::AxumTraceLayer;
-use tokio::signal::unix::{SignalKind, signal};
+use tokio::{
+    signal::unix::{SignalKind, signal},
+    sync::mpsc,
+};
 use tower::ServiceBuilder;
+use x509_cert::Certificate;
 
 // use tower_http::trace::TraceLayer;
 use crate::tls::TlsServer;
@@ -41,10 +46,6 @@ pub mod tls;
 // Selected re-exports
 pub use crate::options::Options;
 
-/// A result type alias with the library-level [`Error`] type as teh default
-/// error type.
-pub type Result<T, E = Error> = std::result::Result<T, E>;
-
 /// A generic webhook handler receiving a request and sending back a response.
 ///
 /// This trait is not intended to be implemented by external crates and this
@@ -52,24 +53,12 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 /// implementation is part of the [`ConversionWebhookServer`][1].
 ///
 /// [1]: crate::servers::ConversionWebhookServer
-pub(crate) trait WebhookHandler<Req, Res> {
+pub trait WebhookHandler<Req, Res> {
     fn call(self, req: Req) -> Res;
 }
 
-/// A generic webhook handler receiving a request and state and sending back
-/// a response.
-///
-/// This trait is not intended to be  implemented by external crates and this
-/// library provides various ready-to-use implementations for it. One such an
-/// implementation is part of the [`ConversionWebhookServer`][1].
-///
-/// [1]: crate::servers::ConversionWebhookServer
-pub(crate) trait StatefulWebhookHandler<Req, Res, S> {
-    fn call(self, req: Req, state: S) -> Res;
-}
-
 #[derive(Debug, Snafu)]
-pub enum Error {
+pub enum WebhookError {
     #[snafu(display("failed to create TLS server"))]
     CreateTlsServer { source: tls::Error },
 
@@ -88,8 +77,7 @@ pub enum Error {
 ///
 /// [1]: crate::servers::ConversionWebhookServer
 pub struct WebhookServer {
-    options: Options,
-    router: Router,
+    tls_server: TlsServer,
 }
 
 impl WebhookServer {
@@ -108,9 +96,11 @@ impl WebhookServer {
     /// ```
     /// use stackable_webhook::{WebhookServer, Options};
     /// use axum::Router;
+    /// use tokio::sync::mpsc;
     ///
+    /// let (cert_tx, _cert_rx) = mpsc::channel(1);
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, Options::default());
+    /// let server = WebhookServer::new(router, Options::default(), vec![], cert_tx);
     /// ```
     ///
     /// ### Example with Custom Options
@@ -118,23 +108,61 @@ impl WebhookServer {
     /// ```
     /// use stackable_webhook::{WebhookServer, Options};
     /// use axum::Router;
+    /// use tokio::sync::mpsc;
     ///
+    /// let (cert_tx, _cert_rx) = mpsc::channel(1);
     /// let options = Options::builder()
     ///     .bind_address([127, 0, 0, 1], 8080)
     ///     .build();
+    /// let sans = vec!["my-san-entry".to_string()];
     ///
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, options);
+    /// let server = WebhookServer::new(router, options, sans, cert_tx);
     /// ```
-    pub fn new(router: Router, options: Options) -> Self {
+    pub async fn new(
+        router: Router,
+        options: Options,
+        subject_alterative_dns_names: Vec<String>,
+        cert_tx: mpsc::Sender<Certificate>,
+    ) -> Result<Self, WebhookError> {
         tracing::trace!("create new webhook server");
-        Self { options, router }
+
+        // TODO (@Techassi): Make opt-in configurable from the outside
+        // Create an OpenTelemetry tracing layer
+        tracing::trace!("create tracing service (layer)");
+        let trace_layer = AxumTraceLayer::new().with_opt_in();
+
+        // Use a service builder to provide multiple layers at once. Recommended
+        // by the Axum project.
+        //
+        // See https://docs.rs/axum/latest/axum/middleware/index.html#applying-multiple-middleware
+        // TODO (@NickLarsenNZ): rename this server_builder and keep it specific to tracing, since it's placement in the chain is important
+        let service_builder = ServiceBuilder::new().layer(trace_layer);
+
+        // Create the root router and merge the provided router into it.
+        tracing::debug!("create core router and merge provided router");
+        let router = router
+            .layer(service_builder)
+            // The health route is below the AxumTraceLayer so as not to be instrumented
+            .route("/health", get(|| async { "ok" }));
+
+        tracing::debug!("create TLS server");
+        let tls_server = TlsServer::new(
+            options.socket_addr,
+            router,
+            subject_alterative_dns_names,
+            cert_tx,
+        )
+        .await
+        .context(CreateTlsServerSnafu)?;
+
+        Ok(Self { tls_server })
     }
 
     /// Runs the Webhook server and sets up signal handlers for shutting down.
     ///
     /// This does not implement graceful shutdown of the underlying server.
-    pub async fn run(self) -> Result<()> {
+    pub async fn run(self) -> Result<(), WebhookError> {
         let future_server = self.run_server();
         let future_signal = async {
             let mut sigint = signal(SignalKind::interrupt()).expect("create SIGINT listener");
@@ -167,36 +195,9 @@ impl WebhookServer {
 
     /// Runs the webhook server by creating a TCP listener and binding it to
     /// the specified socket address.
-    async fn run_server(self) -> Result<()> {
+    async fn run_server(self) -> Result<(), WebhookError> {
         tracing::debug!("run webhook server");
 
-        // TODO (@Techassi): Make opt-in configurable from the outside
-        // Create an OpenTelemetry tracing layer
-        tracing::trace!("create tracing service (layer)");
-        let trace_layer = AxumTraceLayer::new().with_opt_in();
-
-        // Use a service builder to provide multiple layers at once. Recommended
-        // by the Axum project.
-        //
-        // See https://docs.rs/axum/latest/axum/middleware/index.html#applying-multiple-middleware
-        // TODO (@NickLarsenNZ): rename this server_builder and keep it specific to tracing, since it's placement in the chain is important
-        let service_builder = ServiceBuilder::new().layer(trace_layer);
-
-        // Create the root router and merge the provided router into it.
-        tracing::debug!("create core router and merge provided router");
-        let router = self
-            .router
-            .layer(service_builder)
-            // The health route is below the AxumTraceLayer so as not to be instrumented
-            .route("/health", get(|| async { "ok" }));
-
-        // Create server for TLS termination
-        tracing::debug!("create TLS server");
-        let tls_server = TlsServer::new(self.options.socket_addr, router)
-            .await
-            .context(CreateTlsServerSnafu)?;
-
-        tracing::info!("running TLS server");
-        tls_server.run().await.context(RunTlsServerSnafu)
+        self.tls_server.run().await.context(RunTlsServerSnafu)
     }
 }
--- a/crates/stackable-webhook/src/lib.rs
+++ b/crates/stackable-webhook/src/lib.rs
@@ -12,9 +12,8 @@
 //! use axum::Router;
 //! use tokio::sync::mpsc;
 //!
-//! let (cert_tx, _cert_rx) = mpsc::channel(1);
 //! let router = Router::new();
-//! let server = WebhookServer::new(router, Options::default(), vec![], cert_tx);
+//! let server = WebhookServer::new(router, Options::default(), vec![]);
 //! ```
 //!
 //! For some usages, complete end-to-end [`WebhookServer`] implementations
@@ -98,9 +97,8 @@ impl WebhookServer {
     /// use axum::Router;
     /// use tokio::sync::mpsc;
     ///
-    /// let (cert_tx, _cert_rx) = mpsc::channel(1);
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, Options::default(), vec![], cert_tx);
+    /// let server = WebhookServer::new(router, Options::default(), vec![]);
     /// ```
     ///
     /// ### Example with Custom Options
@@ -110,14 +108,13 @@ impl WebhookServer {
     /// use axum::Router;
     /// use tokio::sync::mpsc;
     ///
-    /// let (cert_tx, _cert_rx) = mpsc::channel(1);
     /// let options = Options::builder()
     ///     .bind_address([127, 0, 0, 1], 8080)
     ///     .build();
     /// let sans = vec!["my-san-entry".to_string()];
     ///
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, options, sans, cert_tx);
+    /// let server = WebhookServer::new(router, options, sans);
     /// ```
     pub async fn new(
         router: Router,
--- a/crates/stackable-webhook/src/servers/conversion.rs
+++ b/crates/stackable-webhook/src/servers/conversion.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, fmt::Debug};
+use std::fmt::Debug;
 
 use axum::{Json, Router, routing::post};
 use k8s_openapi::{
@@ -74,7 +74,7 @@ pub struct ConversionWebhookServer {
     cert_rx: mpsc::Receiver<Certificate>,
     client: Client,
     field_manager: String,
-    crds: HashMap<String, CustomResourceDefinition>,
+    crds: Vec<CustomResourceDefinition>,
     operator_environment: OperatorEnvironmentOpts,
 }
 
@@ -106,7 +106,7 @@ impl ConversionWebhookServer {
     /// # async fn test() {
     /// let crds_and_handlers = [
     ///     (
-    ///         S3Connection::merged_crd(S3ConnectionVersion::V1Alpha1).unwrap(),
+    ///         S3Connection::merged_crd(S3ConnectionVersion::V1Alpha1).expect("failed to merge S3Connection CRD"),
     ///         S3Connection::try_convert as fn(ConversionReview) -> ConversionReview,
     ///     ),
     /// ];
@@ -151,7 +151,7 @@ impl ConversionWebhookServer {
         let field_manager: String = field_manager.into();
 
         let mut router = Router::new();
-        let mut crds = HashMap::new();
+        let mut crds = Vec::new();
         for (crd, handler) in crds_and_handlers {
             let crd_name = crd.name_any();
             let handler_fn = |Json(review): Json<ConversionReview>| async {
@@ -160,7 +160,7 @@ impl ConversionWebhookServer {
             };
 
             router = router.route(&format!("/convert/{crd_name}"), post(handler_fn));
-            crds.insert(crd_name, crd);
+            crds.push(crd);
         }
 
         // This is how Kubernetes calls us, so it decides about the naming.
@@ -238,7 +238,7 @@ impl ConversionWebhookServer {
         mut cert_rx: mpsc::Receiver<Certificate>,
         client: &Client,
         field_manager: &str,
-        crds: &HashMap<String, CustomResourceDefinition>,
+        crds: &[CustomResourceDefinition],
         operator_environment: &OperatorEnvironmentOpts,
     ) -> Result<(), ConversionWebhookError> {
         while let Some(current_cert) = cert_rx.recv().await {
@@ -259,18 +259,21 @@ impl ConversionWebhookServer {
     async fn reconcile_crds(
         client: &Client,
         field_manager: &str,
-        crds: &HashMap<String, CustomResourceDefinition>,
+        crds: &[CustomResourceDefinition],
         operator_environment: &OperatorEnvironmentOpts,
         current_cert: &Certificate,
     ) -> Result<(), ConversionWebhookError> {
-        tracing::info!(kinds = ?crds.keys(), "Reconciling CRDs");
+        tracing::info!(
+            crds = ?crds.iter().map(CustomResourceDefinition::name_any).collect::<Vec<_>>(),
+            "Reconciling CRDs"
+        );
         let ca_bundle = current_cert
             .to_pem(LineEnding::LF)
             .context(ConvertCaToPemSnafu)?;
 
         let crd_api: Api<CustomResourceDefinition> = Api::all(client.clone());
-        for (kind, crd) in crds {
-            let mut crd = crd.clone();
+        for mut crd in crds.iter().cloned() {
+            let crd_name = crd.name_any();
 
             crd.spec.conversion = Some(CustomResourceConversion {
                 strategy: "Webhook".to_string(),
@@ -283,7 +286,7 @@ impl ConversionWebhookServer {
                         service: Some(ServiceReference {
                             name: operator_environment.operator_service_name.clone(),
                             namespace: operator_environment.operator_namespace.clone(),
-                            path: Some(format!("/convert/{kind}")),
+                            path: Some(format!("/convert/{crd_name}")),
                             port: Some(DEFAULT_HTTPS_PORT.into()),
                         }),
                         ca_bundle: Some(ByteString(ca_bundle.as_bytes().to_vec())),
@@ -292,8 +295,6 @@ impl ConversionWebhookServer {
                 }),
             });
 
-            // TODO: Move this into function and do a more clever update mechanism
-            let crd_name = crd.name_any();
             let patch = Patch::Apply(&crd);
             let patch_params = PatchParams::apply(field_manager);
             crd_api
@@ -302,7 +303,6 @@ impl ConversionWebhookServer {
                 .with_context(|_| UpdateCRDSnafu {
                     crd_name: crd_name.to_string(),
                 })?;
-            tracing::info!(crd.name = crd_name, "Reconciled CRD");
         }
         Ok(())
     }
--- a/crates/stackable-webhook/src/servers/conversion.rs
+++ b/crates/stackable-webhook/src/servers/conversion.rs
@@ -285,11 +285,7 @@ impl ConversionWebhookServer {
                             name: operator_environment.operator_service_name.clone(),
                             namespace: operator_environment.operator_namespace.clone(),
                             path: Some(format!("/convert/{kind}")),
-                            port: Some(
-                                DEFAULT_HTTPS_PORT
-                                    .try_into()
-                                    .expect("DEFAULT_HTTPS_PORT must be convertible into i32"),
-                            ),
+                            port: Some(DEFAULT_HTTPS_PORT.into()),
                         }),
                         ca_bundle: Some(ByteString(ca_bundle.as_bytes().to_vec())),
                         url: None,
--- a/crates/stackable-webhook/src/servers/conversion.rs
+++ b/crates/stackable-webhook/src/servers/conversion.rs
@@ -1,14 +1,55 @@
-use std::fmt::Debug;
+use std::{collections::HashMap, fmt::Debug};
 
-use axum::{Json, Router, extract::State, routing::post};
+use axum::{Json, Router, routing::post};
+use k8s_openapi::{
+    ByteString,
+    apiextensions_apiserver::pkg::apis::apiextensions::v1::{
+        CustomResourceConversion, CustomResourceDefinition, ServiceReference, WebhookClientConfig,
+        WebhookConversion,
+    },
+};
 // Re-export this type because users of the conversion webhook server require
 // this type to write the handler function. Instead of importing this type from
 // kube directly, consumers can use this type instead. This also eliminates
 // keeping the kube dependency version in sync between here and the operator.
 pub use kube::core::conversion::ConversionReview;
+use kube::{
+    Api, Client, ResourceExt,
+    api::{Patch, PatchParams},
+};
+use snafu::{OptionExt, ResultExt, Snafu};
+use stackable_operator::cli::OperatorEnvironmentOpts;
+use tokio::sync::mpsc;
 use tracing::instrument;
+use x509_cert::{
+    Certificate,
+    der::{EncodePem, pem::LineEnding},
+};
 
-use crate::{StatefulWebhookHandler, WebhookHandler, WebhookServer, options::Options};
+use crate::{
+    WebhookError, WebhookHandler, WebhookServer, constants::DEFAULT_HTTPS_PORT, options::Options,
+};
+
+#[derive(Debug, Snafu)]
+pub enum ConversionWebhookError {
+    #[snafu(display("failed to create webhook server"))]
+    CreateWebhookServer { source: WebhookError },
+
+    #[snafu(display("failed to run webhook server"))]
+    RunWebhookServer { source: WebhookError },
+
+    #[snafu(display("failed to receive certificate from channel"))]
+    ReceiverCertificateFromChannel,
+
+    #[snafu(display("failed to convert CA certificate into PEM format"))]
+    ConvertCaToPem { source: x509_cert::der::Error },
+
+    #[snafu(display("failed to update CRD {crd_name:?}"))]
+    UpdateCRD {
+        source: stackable_operator::kube::Error,
+        crd_name: String,
+    },
+}
 
 impl<F> WebhookHandler<ConversionReview, ConversionReview> for F
 where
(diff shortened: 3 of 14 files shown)


crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -25,130 +22,89 @@ use tokio_rustls::{
...
 use tracing::{Instrument, Span, field::Empty, instrument};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
+use x509_cert::Certificate;
+
+mod cert_resolver;
 
-pub type Result<T, E = Error> = std::result::Result<T, E>;
+pub const WEBHOOK_CA_LIFETIME: Duration = Duration::from_minutes_unchecked(3);
+pub const WEBHOOK_CERTIFICATE_LIFETIME: Duration = Duration::from_minutes_unchecked(2);
+pub const WEBHOOK_CERTIFICATE_ROTATION_INTERVAL: Duration = Duration::from_minutes_unchecked(1);
```
sbernauer: reminder to bump these before merging. Currently they are so low for easy testing

crates/stackable-operator/src/cli.rs
```diff
@@ -216,6 +233,9 @@ pub struct ProductOperatorRun {
     #[arg(long, env, default_value = "")]
     pub watch_namespace: WatchNamespace,
 
+    #[command(flatten)]
+    pub operator_environment: OperatorEnvironmentOpts,
```
Techassi: note: Please rename to `OperatorEnvironmentOptions`.

crates/stackable-operator/src/cli.rs
```diff
@@ -216,6 +233,9 @@ pub struct ProductOperatorRun {
...
+
     #[command(flatten)]
     pub telemetry_arguments: TelemetryOptions,
```
Techassi: note(non-blocking): I know this is unrelated, but can we simplify this to `telemetry` to be inline with `operator_environment`. This would be a breaking change for downstream operators, but the addition of `operator_environment` already is breaking anyways.

note(non-blocking): Could we also do the same for `cluster_info_opts`. The `_opts` suffix seems redundant. We could also rename the struct to `ClusterInfoOptions`, because Kubernetes is implied (by context) and I prefer `Options` over `Opts`.

crates/stackable-operator/src/cli.rs
```diff
@@ -278,6 +298,18 @@ impl ProductConfigPath {
     }
 }
 
+#[derive(clap::Parser, Debug, PartialEq, Eq)]
+pub struct OperatorEnvironmentOpts {
+    /// The namespace the operator is running in, usually `stackable-operators`.
```
Techassi, line 308: suggestion: We should mention that setting the operator namespace is preferred to be done via the env variable in combination with the downward API (which should link to the official [docs](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)).

crates/stackable-operator/src/cli.rs
```diff
@@ -396,23 +434,42 @@ mod tests {
...
+                operator_environment: OperatorEnvironmentOpts {
+                    operator_namespace: "stackable-operators".to_string(),
+                    operator_service_name: "foo-operator".to_string(),
+                }
             }
         );
 
         // env with namespace
         unsafe { env::set_var(WATCH_NAMESPACE, "foo") };
+        unsafe { env::set_var(OPERATOR_SERVICE_NAME, "foo-operator") };
```
Techassi: note: I honestly don't know why we test this. This is again something what the `clap` project tests upstream. If this breaks, the whole Rust user-base would be in deep trouble, because basically all CLI tools use `clap`.

As such, I would just get rid of all those tests.

crates/stackable-webhook/src/lib.rs
```diff
@@ -41,40 +45,27 @@ pub mod tls;
...
-/// a response.
-///
-/// This trait is not intended to be  implemented by external crates and this
-/// library provides various ready-to-use implementations for it. One such an
-/// implementation is part of the [`ConversionWebhookServer`][1].
-///
-/// [1]: crate::servers::ConversionWebhookServer
-pub(crate) trait StatefulWebhookHandler<Req, Res, S> {
-    fn call(self, req: Req, state: S) -> Res;
-}
```
Techassi: question: Why was this removed?

crates/stackable-webhook/src/lib.rs
```diff
@@ -108,27 +98,60 @@ impl WebhookServer {
...
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, options);
+    /// let server = WebhookServer::new(router, options, sans);
     /// ```
-    pub fn new(router: Router, options: Options) -> Self {
+    pub async fn new(
+        router: Router,
+        options: Options,
+        subject_alterative_dns_names: Vec<String>,
+    ) -> Result<(Self, mpsc::Receiver<Certificate>)> {
```
Techassi: note: I would like to make the list of SANs part of the `Options` struct to keep the API interface of the `new` function clean and non-breaking.

crates/stackable-webhook/src/lib.rs
```diff
@@ -108,27 +98,60 @@ impl WebhookServer {
     /// ```
     /// use stackable_webhook::{WebhookServer, Options};
     /// use axum::Router;
+    /// use tokio::sync::mpsc;
```
Techassi: question: Why do we need to import `mpsc` in a doc test? It is not used anywhere.

crates/stackable-webhook/src/lib.rs
```diff
@@ -108,27 +98,60 @@ impl WebhookServer {
...
     ///
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, Options::default());
+    /// let server = WebhookServer::new(router, Options::default(), vec![]);
```
Techassi: note: I would like us to split the return value to indicate that it is tuple (pair) of values.

```suggestion
    /// let (server, cert_rx) = WebhookServer::new(router, Options::default(), vec![]);
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -108,27 +98,60 @@ impl WebhookServer {
...
+    /// use tokio::sync::mpsc;
     ///
     /// let options = Options::builder()
     ///     .bind_address([127, 0, 0, 1], 8080)
     ///     .build();
+    /// let sans = vec!["my-san-entry".to_string()];
     ///
     /// let router = Router::new();
-    /// let server = WebhookServer::new(router, options);
+    /// let server = WebhookServer::new(router, options, sans);
```
Techassi: suggestion: Same as above.

```suggestion
    /// let (server, cert_rx) = WebhookServer::new(router, options, sans);
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -41,40 +45,27 @@ pub mod tls;
...
-/// [1]: crate::servers::ConversionWebhookServer
-pub(crate) trait StatefulWebhookHandler<Req, Res, S> {
-    fn call(self, req: Req, state: S) -> Res;
-}
+/// A result type alias with the [`WebhookError`] type as the default error type.
+pub type Result<T, E = WebhookError> = std::result::Result<T, E>;
 
 #[derive(Debug, Snafu)]
-pub enum Error {
+pub enum WebhookError {
```
Techassi, line 66: question: Any particular reason why this was renamed? It is additionally a breaking change which we could avoid.

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -25,130 +22,89 @@ use tokio_rustls::{
...
+        socket_addr: SocketAddr,
+        router: Router,
+        subject_alterative_dns_names: Vec<String>,
+    ) -> Result<(Self, mpsc::Receiver<Certificate>)> {
+        let (cert_tx, cert_rx) = mpsc::channel(1);
+        let cert_resolver = Arc::new(
+            CertificateResolver::new(subject_alterative_dns_names, cert_tx)
+                .await
+                .context(CreateCertificateResolverSnafu)?,
+        );
```
Techassi: suggestion: Split this up and give it more room to breathe.

```suggestion
        let (cert_tx, cert_rx) = mpsc::channel(1);

        let cert_resolver = CertificateResolver::new(subject_alterative_dns_names, cert_tx)
                .await
                .context(CreateCertificateResolverSnafu)?
        let cert_resolver = Arc::new(cert_resolver);
```

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -25,130 +22,89 @@ use tokio_rustls::{
...
-
-            config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
-            let config = Arc::new(config);
-
-            Ok(Self {
-                socket_addr,
+    pub async fn new<'a>(
+        socket_addr: SocketAddr,
+        router: Router,
+        subject_alterative_dns_names: Vec<String>,
```
Techassi: suggestion: We might wanna bundle this up similar to the webhook server.

```suggestion
        router: Router,
        options: Options, // Contains socket_addr and SANs
```

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -25,130 +22,89 @@ use tokio_rustls::{
...
                 router,
-            })
-        })
-        .await
-        .context(TokioSpawnBlockingSnafu)??;
-
-        Ok(task)
+            },
+            cert_rx,
+        ))
```
Techassi: suggestion: Split this up.

```rust
let tls_server = Self { .. };
Ok((tls_server, cert_rx))
```

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -25,130 +22,89 @@ use tokio_rustls::{
...
-                .with_no_client_auth()
-                .with_single_cert(vec![certificate_der], private_key_der)
-                .context(InvalidTlsPrivateKeySnafu)?;
-
-            config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
-            let config = Arc::new(config);
-
-            Ok(Self {
-                socket_addr,
+    pub async fn new<'a>(
```
Techassi: suggestion: I know that there wasn't a doc comment before, but we now might want to add one. It should mention that we construct a `CertificateResolver` with the provided SANs and then continue to build a `ServerConfig` which uses above mentioned `CertificateResolver`.

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -25,130 +22,89 @@ use tokio_rustls::{
...
-        Ok(task)
+            },
+            cert_rx,
+        ))
     }
 
     /// Runs the TLS server by listening for incoming TCP connections on the
     /// bound socket address. It only accepts TLS connections. Internally each
     /// TLS stream get handled by a Hyper service, which in turn is an Axum
     /// router.
```
Techassi: note: This should mention that we also start the certificate rotation in a separate Tokio task.

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -171,12 +127,21 @@ impl TlsServer {
...
+            let tcp_stream = select! {
+                loop_result = &mut certificate_rotation_loop => {
+                    return loop_result.context(RunCertificateRotationLoopSnafu)?;
+                }
+                tcp_stream = tcp_listener.accept() => {
+                    tcp_stream
+                }
+            };
+
+            let (tcp_stream, remote_addr) = match tcp_stream {
```
Techassi: suggestion: I have an improved version of this locally. It removes the need for `pin_mut!`, removes the `RunCertificateRotationLoop` error variant, bubbles up the underlying rotation error and cleans up the TCP connection code.

Let me know if you want me to push it.

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -264,6 +229,22 @@ impl TlsServer {
...
+
+        loop {
+            interval.tick().await;
+
+            cert_resolver
+                .rotate_certificate()
+                .await
+                .context(RotateCertificateSnafu)?;
+        }
+    }
```
Techassi: note: With the above suggestion, this code could also be removed/simplified.

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -19,141 +66,246 @@ where
...
+    #[instrument(
+        name = "create_conversion_webhook_server",
+        skip(crds_and_handlers, client)
+    )]
+    pub async fn new<H>(
+        crds_and_handlers: impl IntoIterator<Item = (CustomResourceDefinition, H)>,
+        options: Options,
+        client: Client,
+        field_manager: impl Into<String> + Debug,
+        operator_environment: OperatorEnvironmentOpts,
```
Techassi: note: I like to avoid having a hard dependency on `OperatorEnvironmentOpts` from `stackable_operator`. Instead, both the namespace and the service name should be part of the `Options` struct (which can be specific for `ConversionWebhookServer`, `WebhookServer` and `TlsServer`).

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -19,141 +66,246 @@ where
...
+    /// # }
     /// ```
-    #[instrument(name = "create_conversion_webhhok_server", skip(handler))]
-    pub fn new<H>(handler: H, options: Options) -> Self
+    #[instrument(
+        name = "create_conversion_webhook_server",
+        skip(crds_and_handlers, client)
+    )]
+    pub async fn new<H>(
+        crds_and_handlers: impl IntoIterator<Item = (CustomResourceDefinition, H)>,
```
Techassi, line 165: note(non-blocking): In the future I want this to be `handlers: impl IntoIterator<Item = VersionedResource`, where `VersionedResource` enables access to the `try_convert` function as well as the CRD kind.

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -19,141 +66,246 @@ where
...
+            let crd_name = crd.name_any();
+            let handler_fn = |Json(review): Json<ConversionReview>| async {
+                let review = handler.call(review);
+                Json(review)
+            };
 
-        let router = Router::new().route("/convert", post(handler_fn));
-        Self { router, options }
-    }
+            router = router.route(&format!("/convert/{crd_name}"), post(handler_fn));
```
Techassi: suggestion: Split this.

```suggestion
            let route = format!("/convert/{crd_name}, crd_name = crd.name_any());
            router = router.route(&route), post(handler_fn));
```

crates/stackable-webhook/src/servers/conversion.rs
Techassi: note: A good place for this might be the `Options` for `ConversionWebhookServer`.

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -19,141 +66,246 @@ where
...
-    /// );
-    ///
-    /// // Define the handler function
-    /// fn handler(req: ConversionReview, state: Arc<State>) -> ConversionReview {
-    ///    // In here we can do the CRD conversion
-    ///    req
-    /// }
-    /// ```
-    #[instrument(name = "create_conversion_webhook_server_with_state", skip(handler))]
-    pub fn new_with_state<H, S>(handler: H, state: S, options: Options) -> Self
```
Techassi, line 121: question: Why is this removed?

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -19,141 +66,246 @@ where
...
+        tracing::info!(
+            crds = ?crds.iter().map(CustomResourceDefinition::name_any).collect::<Vec<_>>(),
+            "Reconciling CRDs"
+        );
+        let ca_bundle = current_cert
+            .to_pem(LineEnding::LF)
+            .context(ConvertCaToPemSnafu)?;
+
+        let crd_api: Api<CustomResourceDefinition> = Api::all(client.clone());
+        for mut crd in crds.iter().cloned() {
```
Techassi, line 307: note: I feel like we shouldn't clone here, but instead mutate the CRDs (so that we also have an up2date version of the CRD in memory).

I also feel like we should be a little more clever on when to run all of the following code. We can skip running all of the below code for like 99% percent of the time, as the certificate is still valid and nothing needs to be adjusted in the `conversion` section of the CRD.

crates/stackable-webhook/src/lib.rs
NickLarsenNZ: I prefer this too (well, rather I prefer to avoid having `.0` which doesn't communicate anything important)

NickLarsenNZ: Note, they are unused. They might need to be prefixed with `_`. Unsure what happens for docs.

crates/stackable-operator/src/cli.rs
sbernauer: 2ec9b73

sbernauer: I guess you know that I straight up copied from `KubernetesClusterInfoOpts`. changed that as well

crates/stackable-operator/src/cli.rs
sbernauer: 2ec9b73

Regarding dropping Kubernetes prefix one could argue yes. But just yesterday someone talked about cluster and nodes and I was confused. He meant NifiCluster and nodes....

crates/stackable-operator/src/cli.rs
sbernauer: removed it in a8a2381

crates/stackable-webhook/src/lib.rs
sbernauer: It imposed maintenance effort to me while I figured out how the code should look like.
I'm also pretty confident we are never going to need it for a conversion web-hook. Therefore I removed it to have less maintenance effort going forward.
Also, stateful conversion webhooks sounds like a potential nightmare once we want to have HA and multiple instances.
We can obviously always re-add it in the case we really need it in the future.

crates/stackable-webhook/src/lib.rs
sbernauer: 9089385

crates/stackable-webhook/src/lib.rs
sbernauer: Good catch, that's a leftover a3b3cc1

crates/stackable-webhook/src/lib.rs
sbernauer: Good catch, actually server here now is a future :D
Fixed in f8683e4

crates/stackable-webhook/src/lib.rs
sbernauer: It was to be consistent with `TlsServerError` and `CertificateResolverError`.
Previously I think (would need to check) stackable-webhook was all one single error enum, we now have per-module errors.

As AFAIK we currently have 0 users, now would be a good time to do the rename

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: Sounds like a duplicate of https://github.com/stackabletech/operator-rs/pull/1066#discussion_r2182141365 to me

crates/stackable-webhook/src/tls/mod.rs
sbernauer: dcb22bd

crates/stackable-webhook/src/tls/mod.rs
sbernauer: I'm interested, can you please paste it here as diff file? This way I can have a look and apply it if I'm happy.
And also take care of conflicts

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: Hope f1fee5d works for you

crates/stackable-webhook/src/tls/mod.rs
sbernauer: caddffb

crates/stackable-webhook/src/tls/mod.rs
Techassi: Sure, I will add it asap.

crates/stackable-webhook/src/tls/mod.rs
sbernauer: 7450d85

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: That doesn't really work out in terms of borrow checker (see below). Also I prefer the signature this way, as the *only* thing that we should change about the CRD is the `conversion` (which changes every `reconcile_crds` call).
Having multiple functions take owned or &mut CRDs makes this a bit less clear and some function might (accidentally) change things.
This clone is called every 7 days or so (cert lifetime), I'd say the performance is negligible here.

> we should be a little more clever

Currently `reconcile_crds` is called for every new cert, so every 7 days or so and every time the cert changes.
Even if there would be no change, k8s would not do anything, as the object is unchanged.

```diff
diff --git a/crates/stackable-webhook/src/servers/conversion.rs b/crates/stackable-webhook/src/servers/conversion.rs
index 03fc3b2..0d265b1 100644
--- a/crates/stackable-webhook/src/servers/conversion.rs
+++ b/crates/stackable-webhook/src/servers/conversion.rs
@@ -189,7 +189,7 @@ impl ConversionWebhookServer {
         Self::reconcile_crds(
             &client,
             &field_manager,
-            &crds,
+            crds.clone(),
             &operator_environment,
             &current_cert,
         )
@@ -240,14 +240,14 @@ impl ConversionWebhookServer {
         mut cert_rx: mpsc::Receiver<Certificate>,
         client: &Client,
         field_manager: &str,
-        crds: &[CustomResourceDefinition],
+        mut crds: Vec<CustomResourceDefinition>,
         operator_environment: &OperatorEnvironmentOptions,
     ) -> Result<(), ConversionWebhookError> {
         while let Some(current_cert) = cert_rx.recv().await {
             Self::reconcile_crds(
                 client,
                 field_manager,
-                crds,
+                &mut crds,
                 operator_environment,
                 &current_cert,
             )
@@ -261,7 +261,7 @@ impl ConversionWebhookServer {
     async fn reconcile_crds(
         client: &Client,
         field_manager: &str,
-        crds: &[CustomResourceDefinition],
+        crds: &mut [CustomResourceDefinition],
         operator_environment: &OperatorEnvironmentOptions,
         current_cert: &Certificate,
     ) -> Result<(), ConversionWebhookError> {
```

crates/stackable-operator/src/cli.rs
sbernauer: ac29c0e

crates/stackable-operator/src/cli.rs
Techassi: I do, thank you for updating both!

crates/stackable-webhook/src/tls/mod.rs
Techassi: Here you go:

```patch
diff --git a/crates/stackable-webhook/src/tls/mod.rs b/crates/stackable-webhook/src/tls/mod.rs
index 560331b..9bfcd65 100644
--- a/crates/stackable-webhook/src/tls/mod.rs
+++ b/crates/stackable-webhook/src/tls/mod.rs
@@ -2,15 +2,21 @@
 //! server, which can be used in combination with an Axum [`Router`].
 use std::{net::SocketAddr, sync::Arc};
 
-use axum::{Router, extract::Request};
+use axum::{
+    Router,
+    extract::{ConnectInfo, Request},
+    middleware::AddExtension,
+};
 use cert_resolver::{CertificateResolver, CertificateResolverError};
-use futures_util::pin_mut;
 use hyper::{body::Incoming, service::service_fn};
 use hyper_util::rt::{TokioExecutor, TokioIo};
 use opentelemetry::trace::{FutureExt, SpanKind};
 use snafu::{ResultExt, Snafu};
 use stackable_operator::time::Duration;
-use tokio::{net::TcpListener, select, sync::mpsc, time::interval};
+use tokio::{
+    net::{TcpListener, TcpStream},
+    sync::mpsc,
+};
 use tokio_rustls::{
     TlsAcceptor,
     rustls::{
@@ -48,9 +54,6 @@ pub enum TlsServerError {
 
     #[snafu(display("failed to set safe TLS protocol versions"))]
     SetSafeTlsProtocolVersions { source: tokio_rustls::rustls::Error },
-
-    #[snafu(display("failed to run certificate rotation loop"))]
-    RunCertificateRotationLoop { source: tokio::task::JoinError },
 }
 
 /// A server which terminates TLS connections and allows clients to communicate
@@ -109,8 +112,8 @@ impl TlsServer {
     ///
     /// It also starts a background task to rotate the certificate as needed.
     pub async fn run(self) -> Result<()> {
-        let certificate_rotation_loop =
-            tokio::spawn(async { Self::run_certificate_rotation_loop(self.cert_resolver).await });
+        let start = tokio::time::Instant::now() + *WEBHOOK_CERTIFICATE_ROTATION_INTERVAL;
+        let mut interval = tokio::time::interval_at(start, *WEBHOOK_CERTIFICATE_ROTATION_INTERVAL);
 
         let tls_acceptor = TlsAcceptor::from(Arc::new(self.config));
         let tcp_listener =
@@ -135,123 +138,123 @@ impl TlsServer {
             .router
             .into_make_service_with_connect_info::<SocketAddr>();
 
-        pin_mut!(certificate_rotation_loop);
         loop {
             let tls_acceptor = tls_acceptor.clone();
 
-            // Wait for either a new TCP connection or the certificate rotation loop to exit
-            let tcp_stream = select! {
-                loop_result = &mut certificate_rotation_loop => {
-                    return loop_result.context(RunCertificateRotationLoopSnafu)?;
-                }
-                tcp_stream = tcp_listener.accept() => {
-                    tcp_stream
-                }
-            };
-
-            let (tcp_stream, remote_addr) = match tcp_stream {
-                Ok((stream, addr)) => (stream, addr),
-                Err(err) => {
-                    tracing::trace!(%err, "failed to accept incoming TCP connection");
-                    continue;
+            // Wait for either a new TCP connection or the certificate rotation interval tick
+            tokio::select! {
+                // We opt for a biased execution of arms to make sure we always check if the
+                // certificate needs rotation based on the interval. This ensures, we always use
+                // a valid certificate for the TLS connection.
+                biased;
+
+                // This is cancellation-safe. If this branch is cancelled, the tick is NOT consumed.
+                // As such, we will not miss rotating the certificate.
+                _ = interval.tick() => {
+                    self.cert_resolver
+                        .rotate_certificate()
+                        .await
+                        .context(RotateCertificateSnafu)?
                 }
-            };
 
-            // Here, the connect info is extracted by calling Tower's Service
-            // trait function on `IntoMakeServiceWithConnectInfo`
-            let tower_service = router.call(remote_addr).await.unwrap();
-
-            let span = tracing::debug_span!("accept tcp connection");
-            tokio::spawn(
-                async move {
-                    let span = tracing::trace_span!(
-                        "accept tls connection",
-                        "otel.kind" = ?SpanKind::Server,
-                        "otel.status_code" = Empty,
-                        "otel.status_message" = Empty,
-                        "client.address" = remote_addr.ip().to_string(),
-                        "client.port" = remote_addr.port() as i64,
-                        "server.address" = Empty,
-                        "server.port" = Empty,
-                        "network.peer.address" = remote_addr.ip().to_string(),
-                        "network.peer.port" = remote_addr.port() as i64,
-                        "network.local.address" = Empty,
-                        "network.local.port" = Empty,
-                        "network.transport" = "tcp",
-                        "network.type" = self.socket_addr.semantic_convention_network_type(),
-                    );
-
-                    if let Ok(local_addr) = tcp_stream.local_addr() {
-                        let addr = &local_addr.ip().to_string();
-                        let port = local_addr.port();
-                        span.record("server.address", addr)
-                            .record("server.port", port as i64)
-                            .record("network.local.address", addr)
-                            .record("network.local.port", port as i64);
-                    }
-
-                    // Wait for tls handshake to happen
-                    let tls_stream = match tls_acceptor
-                        .accept(tcp_stream)
-                        .instrument(span.clone())
-                        .await
-                    {
-                        Ok(tls_stream) => tls_stream,
+                // This is cancellation-safe. If cancelled, no new connections are accepted.
+                tcp_connection = tcp_listener.accept() => {
+                    let (tcp_stream, remote_addr) = match tcp_connection {
+                        Ok((stream, addr)) => (stream, addr),
                         Err(err) => {
-                            span.record("otel.status_code", "Error")
-                                .record("otel.status_message", err.to_string());
-                            tracing::trace!(%remote_addr, "error during tls handshake connection");
-                            return;
+                            tracing::trace!(%err, "failed to accept incoming TCP connection");
+                            continue;
                         }
                     };
 
-                    // Hyper has its own `AsyncRead` and `AsyncWrite` traits and doesn't use tokio.
-                    // `TokioIo` converts between them.
-                    let tls_stream = TokioIo::new(tls_stream);
-
-                    // Hyper also has its own `Service` trait and doesn't use tower. We can use
-                    // `hyper::service::service_fn` to create a hyper `Service` that calls our app through
-                    // `tower::Service::call`.
-                    let hyper_service = service_fn(move |request: Request<Incoming>| {
-                        // This carries the current context with the trace id so that the TraceLayer can use that as a parent
-                        let otel_context = Span::current().context();
-                        // We need to clone here, because oneshot consumes self
-                        tower_service
-                            .clone()
-                            .oneshot(request)
-                            .with_context(otel_context)
-                    });
-
-                    let span = tracing::debug_span!("serve connection");
-                    hyper_util::server::conn::auto::Builder::new(TokioExecutor::new())
-                        .serve_connection_with_upgrades(tls_stream, hyper_service)
-                        .instrument(span.clone())
-                        .await
-                        .unwrap_or_else(|err| {
-                            span.record("otel.status_code", "Error")
-                                .record("otel.status_message", err.to_string());
-                            tracing::warn!(%err, %remote_addr, "failed to serve connection");
-                        })
+                    // Here, the connect info is extracted by calling Tower's Service
+                    // trait function on `IntoMakeServiceWithConnectInfo`
+                    let tower_service = router.call(remote_addr).await.unwrap();
+
+                    let span = tracing::debug_span!("accept tcp connection");
+                    tokio::spawn(async move {
+                        Self::handle_request(tcp_stream, remote_addr, tls_acceptor, tower_service, self.socket_addr)
+                    }.instrument(span));
                 }
-                .instrument(span),
-            );
+            };
         }
     }
 
-    async fn run_certificate_rotation_loop(cert_resolver: Arc<CertificateResolver>) -> Result<()> {
-        let mut interval = interval(*WEBHOOK_CERTIFICATE_ROTATION_INTERVAL);
-        // Let the interval tick once, so that the first loop iteration does not start immediately,
-        // thus generating a new cert.
-        interval.tick().await;
+    async fn handle_request(
+        tcp_stream: TcpStream,
+        remote_addr: SocketAddr,
+        tls_acceptor: TlsAcceptor,
+        tower_service: AddExtension<Router, ConnectInfo<SocketAddr>>,
+        socket_addr: SocketAddr,
+    ) {
+        let span = tracing::trace_span!(
+            "accept tls connection",
+            "otel.kind" = ?SpanKind::Server,
+            "otel.status_code" = Empty,
+            "otel.status_message" = Empty,
+            "client.address" = remote_addr.ip().to_string(),
+            "client.port" = remote_addr.port() as i64,
+            "server.address" = Empty,
+            "server.port" = Empty,
+            "network.peer.address" = remote_addr.ip().to_string(),
+            "network.peer.port" = remote_addr.port() as i64,
+            "network.local.address" = Empty,
+            "network.local.port" = Empty,
+            "network.transport" = "tcp",
+            "network.type" = socket_addr.semantic_convention_network_type(),
+        );
+
+        if let Ok(local_addr) = tcp_stream.local_addr() {
+            let addr = &local_addr.ip().to_string();
+            let port = local_addr.port();
+            span.record("server.address", addr)
+                .record("server.port", port as i64)
+                .record("network.local.address", addr)
+                .record("network.local.port", port as i64);
+        }
 
-        loop {
-            interval.tick().await;
+        // Wait for tls handshake to happen
+        let tls_stream = match tls_acceptor
+            .accept(tcp_stream)
+            .instrument(span.clone())
+            .await
+        {
+            Ok(tls_stream) => tls_stream,
+            Err(err) => {
+                span.record("otel.status_code", "Error")
+                    .record("otel.status_message", err.to_string());
+                tracing::trace!(%remote_addr, "error during tls handshake connection");
+                return;
+            }
+        };
 
-            cert_resolver
-                .rotate_certificate()
-                .await
-                .context(RotateCertificateSnafu)?;
-        }
+        // Hyper has its own `AsyncRead` and `AsyncWrite` traits and doesn't use tokio.
+        // `TokioIo` converts between them.
+        let tls_stream = TokioIo::new(tls_stream);
+
+        // Hyper also has its own `Service` trait and doesn't use tower. We can use
+        // `hyper::service::service_fn` to create a hyper `Service` that calls our app through
+        // `tower::Service::call`.
+        let hyper_service = service_fn(move |request: Request<Incoming>| {
+            // This carries the current context with the trace id so that the TraceLayer can use that as a parent
+            let otel_context = Span::current().context();
+            // We need to clone here, because oneshot consumes self
+            tower_service
+                .clone()
+                .oneshot(request)
+                .with_context(otel_context)
+        });
+
+        let span = tracing::debug_span!("serve connection");
+        hyper_util::server::conn::auto::Builder::new(TokioExecutor::new())
+            .serve_connection_with_upgrades(tls_stream, hyper_service)
+            .instrument(span.clone())
+            .await
+            .unwrap_or_else(|err| {
+                span.record("otel.status_code", "Error")
+                    .record("otel.status_message", err.to_string());
+                tracing::warn!(%err, %remote_addr, "failed to serve connection");
+            })
     }
 }
 

```

sbernauer: Picked it in 51321f0

sbernauer: Also added 082a59d

crates/stackable-operator/src/cli.rs
Techassi: Yeah, true. Let's keep the Kubernetes prefix for now. The rest of the renames look good.

crates/stackable-operator/src/cli.rs
Techassi: Can we use "project" instead of "mount" in the doc comment? The same term is used in the official docs [here](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables).

crates/stackable-webhook/src/lib.rs
Techassi: Alright, let's leave it as is for now. We can always re-add if needed.

This then needs a changelog entry.

crates/stackable-webhook/src/lib.rs
Techassi: Alright, let's keep it as is.

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+    cert_resolver: Arc<CertificateResolver>,
+
+    socket_addr: SocketAddr,
+    router: Router,
+}
+
+impl TlsServer {
+    /// Create a new [`TlsServer`].
+    ///
+    /// This create a [`CertificateResolver`] with the provided `subject_alterative_dns_names`,
```
Techassi: suggestion: Fix small typo.


```suggestion
    /// This creates a [`CertificateResolver`] with the provided `subject_alterative_dns_names`,
```

crates/stackable-webhook/src/tls/mod.rs
Techassi: Huh, didn't know this is _actually_ infallible. Nice catch!

crates/stackable-webhook/src/tls/mod.rs
Techassi: Resolved by 51321f0.

crates/stackable-operator/src/cli.rs
sbernauer: Sure thing, 9da5588

crates/stackable-webhook/src/lib.rs
sbernauer: Mentioned in 0903686

crates/stackable-operator/src/cli.rs
```diff
@@ -381,16 +418,17 @@ mod tests {
...
         let opts = ProductOperatorRun::parse_from([
             "run",
             "--product-config",
             "bar",
             "--watch-namespace",
             "foo",
+            "--operator-namespace",
+            "stackable-operators",
+            "--operator-service-name",
+            "foo-operator",
```
Techassi: note: I still feel like all these CLI unit tests are pretty much useless and should be removed to speed up the testing runs.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -34,7 +34,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Update `kube` to `1.1.0` ([#1049]).
-- BREAKING: Return type for `ListenerOperatorVolumeSourceBuilder::new()` is no onger a `Result` ([#1058]).
+- BREAKING: Return type for `ListenerOperatorVolumeSourceBuilder::new()` is no longer a `Result` ([#1058]).
```
Techassi: note: This is already released and as such needs to be removed here.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -34,7 +34,12 @@ All notable changes to this project will be documented in this file.
...
+- BREAKING: Require two new CLI arguments: `--operator-namespace` and `-operator-service-name`.
+  These are required, so that the operator knows what Service it needs to enter as CRD conversion webhook ([#1066]).
```
Techassi: note: This is an addition and as such needs to be moved to the "Added" section. Also, slight reword:


```suggestion
- BREAKING: Add two new required CLI arguments: `--operator-namespace` and `-operator-service-name`.
  These two values are used to construct the service name in the CRD conversion webhook ([#1066]).
```

crates/stackable-operator/CHANGELOG.md
```diff
@@ -34,7 +34,12 @@ All notable changes to this project will be documented in this file.
...
+- BREAKING: The `ProductOperatorRun` used for CLI arguments has some field renamed for consistency ([#1066]):
+  - `telemetry_arguments` -> `telemetry`
+  - `cluster_info_opts` -> `cluster_info`
```
Techassi: suggestion: Slight reword.


```suggestion
- BREAKING: Rename two fields of the `ProductOperatorRun` struct for consistency and clarity ([#1066]):
  - `telemetry_arguments` -> `telemetry`
  - `cluster_info_opts` -> `cluster_info`
```

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -1,14 +1,61 @@
...
+    api::{Patch, PatchParams},
+};
+use snafu::{OptionExt, ResultExt, Snafu};
+use stackable_operator::cli::OperatorEnvironmentOptions;
+use tokio::{sync::mpsc, try_join};
 use tracing::instrument;
+use x509_cert::{
+    Certificate,
+    der::{EncodePem, pem::LineEnding},
+};
```
Techassi: note: Please combine these imports with the others above.

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -1,14 +1,61 @@
...
+    ConvertCaToPem { source: x509_cert::der::Error },
+
+    #[snafu(display("failed to reconcile CRDs"))]
+    ReconcileCRDs {
+        #[snafu(source(from(ConversionWebhookError, Box::new)))]
+        source: Box<ConversionWebhookError>,
+    },
+
+    #[snafu(display("failed to update CRD {crd_name:?}"))]
+    UpdateCRD {
```
Techassi: note: I think we should stick with PascalCase here.


```suggestion
    ReconcileCrd {
        #[snafu(source(from(ConversionWebhookError, Box::new)))]
        source: Box<ConversionWebhookError>,
    },

    #[snafu(display("failed to update CRD {crd_name:?}"))]
    UpdateCrd {
```

Techassi (changes requested):
Left a bunch of comments.

I encountered the ConversionWebhookServer which is missing the changes we talked about a few weeks back. As such, it doesn't make a whole lot of sense to continue the review before these changes are implemented.

crates/stackable-operator/CHANGELOG.md
sbernauer: fixed in merge conflict resolution

crates/stackable-operator/CHANGELOG.md
sbernauer: fixed in merge conflict resolution

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: hmm, what imports exactly can be combined? rustfmt groups them like this

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: fe1749d

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- BREAKING: Add two new required CLI arguments: `--operator-namespace` and `-operator-service-name`.
```
Techassi:
```suggestion
- BREAKING: Add two new required CLI arguments: `--operator-namespace` and `--operator-service-name`.
```

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -1,14 +1,62 @@
...
+#[derive(Debug, Snafu)]
+pub enum ConversionWebhookError {
+    #[snafu(display("failed to create webhook server"))]
+    CreateWebhookServer { source: WebhookError },
+
+    #[snafu(display("failed to run webhook server"))]
+    RunWebhookServer { source: WebhookError },
+
+    #[snafu(display("failed to receive certificate from channel"))]
+    ReceiverCertificateFromChannel,
```
Techassi:
```suggestion
    ReceiveCertificateFromChannel,
```

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -19,141 +67,266 @@ where
...
-{
-    fn call(self, req: ConversionReview, state: S) -> ConversionReview {
-        self(req, state)
-    }
+// TODO: Add a builder, maybe with `bon`.
+#[derive(Debug)]
+pub struct ConversionWebhookOptions {
+    /// The environment the operator is running in, notably the namespace and service name it is
+    /// reachable at.
+    pub operator_environment: OperatorEnvironmentOptions,
```
Techassi: note: Let's avoid pulling in a type from `stackable_operator` here. Instead, split it into two separate fields.


```suggestion
    pub namespace: String,
    pub service_name: String,
```

crates/stackable-webhook/src/servers/mod.rs
```diff
@@ -2,4 +2,5 @@
 //! purposes.
 mod conversion;
 
-pub use conversion::*;
+pub use conversion::{ConversionWebhookError, ConversionWebhookOptions, ConversionWebhookServer};
+pub use kube::core::conversion::ConversionReview;
```
Techassi: note: This is already re-exported via `stackable_webhook::servers::conversion::ConversionReview`. As such, please remove it from here again.

crates/stackable-webhook/src/tls/cert_resolver.rs
```diff
@@ -0,0 +1,156 @@
...
+                .certificate_der()
+                .context(EncodeCertificateDerSnafu)?;
+            let private_key_der = certificate
+                .private_key_der()
+                .context(EncodePrivateKeyDerSnafu)?;
+            let certificate_key =
+                CertifiedKey::from_der(vec![certificate_der], private_key_der, &tls_provider)
+                    .context(DecodeCertifiedKeyFromDerSnafu)?;
+
+            Ok((certificate.certificate().clone(), Arc::new(certificate_key)))
```
Techassi: suggestion: Rename this to `certificate_pair` to better indicate what this is.


```suggestion
            let certificate_pair = ca
                .generate_ecdsa_leaf_certificate(
                    "Leaf",
                    "webhook",
                    subject_alterative_dns_names.iter().map(|san| san.as_str()),
                    WEBHOOK_CERTIFICATE_LIFETIME,
                )
                .context(GenerateLeafCertificateSnafu)?;

            let certificate_der = certificate_pair
                .certificate_der()
                .context(EncodeCertificateDerSnafu)?;
            let private_key_der = certificate_pair
                .private_key_der()
                .context(EncodePrivateKeyDerSnafu)?;
            let certificate_key =
                CertifiedKey::from_der(vec![certificate_der], private_key_der, &tls_provider)
                    .context(DecodeCertifiedKeyFromDerSnafu)?;

            Ok((certificate_pair.certificate().clone(), Arc::new(certificate_key)))
```

crates/stackable-webhook/src/tls/cert_resolver.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// FIXME: This should *not* construct a CA cert and cert, but only a cert!
+    /// This needs some changes in stackable-certs though.
+    /// See [the relevant decision](https://github.com/stackabletech/decisions/issues/56)
+    async fn generate_new_cert(
+        subject_alterative_dns_names: Arc<Vec<String>>,
+    ) -> Result<(Certificate, Arc<CertifiedKey>)> {
+        // The certificate generations can take a while, so we use `spawn_blocking`
+        tokio::task::spawn_blocking(move || {
+            let tls_provider = default_provider();
```
Techassi: note: We shouldn't rely on `ring` being the default crypto provider. Instead we want to use `CryptoProvider::get_default()` instead.

crates/stackable-webhook/src/tls/cert_resolver.rs
```diff
@@ -0,0 +1,156 @@
...
+    EncodeCertificateDer {
+        source: CertificatePairError<ecdsa::Error>,
+    },
+
+    #[snafu(display("failed to encode private key as DER"))]
+    EncodePrivateKeyDer {
+        source: CertificatePairError<ecdsa::Error>,
+    },
+
+    #[snafu(display("failed to decode CertifiedKey from DER"))]
```
Techassi: suggestion: We are _creating_ a new `CertifiedKey` instead of decoding it. Also, `CertifiedKey` seems more like an internal detail. A better fit might be "packaged certificate chain".


```suggestion
    #[snafu(display("failed to create packaged certificate chain from DER"))]
```

crates/stackable-webhook/src/tls/cert_resolver.rs
```diff
@@ -0,0 +1,156 @@
...
+/// This struct serves as [`ResolvesServerCert`] to always hand out the current certificate for TLS
+/// client connections.
+///
+/// It offers the [`Self::rotate_certificate`] function to create a fresh certificate and basically
+/// hot-reload the certificate in the running webhook.
+#[derive(Debug)]
+pub struct CertificateResolver {
+    /// Using a [`ArcSwap`] (over e.g. [`tokio::sync::RwLock`]), so that we can easily
+    /// (and performant) bridge between async write and sync write.
+    current_certified_key: ArcSwap<CertifiedKey>,
```
Techassi: question: Isn't it possible to lock the `Arc` right before we swap in the newly generated certificate and immediately dropping the lock when done. I think there shouldn't be a huge overhead and writing happens way less then reading anyway.

question: Further, where do we need to "bridge between async write and sync write"? I suspect this should be "async read and sync write".

crates/stackable-webhook/src/tls/cert_resolver.rs
```diff
@@ -0,0 +1,156 @@
...
+    ) -> Result<Self> {
+        let subject_alterative_dns_names = Arc::new(subject_alterative_dns_names);
+        let (cert, certified_key) = Self::generate_new_cert(subject_alterative_dns_names.clone())
+            .await
+            .context(GenerateNewCertificateSnafu)?;
+
+        cert_tx
+            .send(cert)
+            .await
+            .map_err(|_err| CertificateResolverError::SendCertificateToChannel)?;
```
Techassi: note: I feel like this can be moved into its own function, because we repeat the exact same code below.

crates/stackable-webhook/src/options.rs
```diff
@@ -10,62 +10,66 @@ use crate::constants::DEFAULT_SOCKET_ADDRESS;
...
 
 /// The [`OptionsBuilder`] which allows to selectively customize the webhook
-/// server [`Options`].
+/// server [`WebhookOptions`].
 ///
 /// Usually, this struct is not constructed manually, but instead by calling
-/// [`Options::builder()`] or [`OptionsBuilder::default()`].
+/// [`WebhookOptions::builder()`] or [`OptionsBuilder::default()`].
 #[derive(Debug, Default)]
 pub struct OptionsBuilder {
```
Techassi: note: We renamed `Options` to `WebhookOptions`. As such, I think we should also rename the builder accordingly.


```suggestion
pub struct WebhookOptionsBuilder {
```

crates/stackable-webhook/CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- BREAKING: Re-write the `ConversionWebhookServer`.
+  It can now do CRD conversions, handle multiple CRDs and takes care of reconciling the CRDs ([#1066]).
```
Techassi, line 10: note: I still feel like the _webhook_ should not take care of reconciling the CRDs, but the operator should do it instead. Let's tackle this in a follow-up PR and leave it as is to get something out of the door.

crates/stackable-webhook/CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
```
Techassi: note: I feel like this should go under the "Changed" section instead. We additionally need to mention the `Options` renames.

crates/stackable-webhook/CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
...
+
+- BREAKING: Re-write the `ConversionWebhookServer`.
+  It can now do CRD conversions, handle multiple CRDs and takes care of reconciling the CRDs ([#1066]).
+- BREAKING: The `TlsServer` can now handle certificate rotation.
+  To achieve this, a new `CertificateResolver` was added.
+  Also, `TlsServer::new` now returns an additional `mpsc::Receiver<Certificate>`, so that the caller
+  can get notified about certificate rotations happening ([#1066]).
```
Techassi, line 14: note: (Addition to the comment above) This would be the perfect trigger for the operator (the caller of this function) to reconcile the CRDs.

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Yeah this is sub-optimal formatting sadly. You need to manually move them to the top imports above line 11.

sbernauer: I'm too stupid, can you please provide a patch that "survives" `cargo fmt`? :man_facepalming:

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: 0e1a160

crates/stackable-webhook/src/servers/mod.rs
sbernauer: Correct, removed in 8dc1235

crates/stackable-webhook/src/tls/cert_resolver.rs
sbernauer: 0c490c2

crates/stackable-webhook/CHANGELOG.md
sbernauer: 0dd99ed69c1fbc7ef4d2a975b966a59a29006900

crates/stackable-webhook/CHANGELOG.md
sbernauer: I personally find it great that we can take care of all this from operator-rs and operators get it "for free".
We might want to move it to somewhere else in op-rs (e.g. stackable-operator), but I don't think every operator should implement it on it's own.

crates/stackable-operator/src/cli.rs
sbernauer: I don't think they have a noticeable effect on the test runtime. Compare to the really slow doc tests, those take 1000x longer (not joking ^^)
I think the maintenance effort is a valid point, removed the CLI parsing tests in 5e41bf1

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: Hmm, I intentionally sticked with the struct that describes the k8s env. I understand your concerns, but I also really don't like too many function arguments when there is a perfectly fine struct for what we want as "information about the environment of the operator".
Maybe tomorrow we need the k8s version or if we are on OpenShift or whatnot

crates/stackable-webhook/src/tls/cert_resolver.rs
sbernauer: 37818fd

crates/stackable-webhook/src/options.rs
sbernauer: c925bd0

crates/stackable-webhook/src/tls/cert_resolver.rs
sbernauer: Would need to think about it again, too long ago
typo fixed in 8c13fd4, thx

crates/stackable-webhook/src/tls/cert_resolver.rs
Techassi: Any reason why you didn't include the certificate generation in that function as well? That's basically also the same code across two functions.

crates/stackable-webhook/CHANGELOG.md
Techassi: > I personally find it great that we can take care of all this from operator-rs and operators get it "for free".

Sure, I also like that. And that's definitely something we should aim for. But the webhook code should not be concerned with reconciling the CRD(s). This is something the operator (or a specific piece of code in `stackable_operator`) should do.

But as I said, let's get something out of the door first.

crates/stackable-webhook/src/tls/cert_resolver.rs
Techassi: The sentence is still wrong, right? We are writing a new certificate in a sync way (it blocks, as such, we use `spawn_blocking`). The reading is done in an async way. Or am I looking at this completely wrong?

I still don't quite understand why we "optimize" for something, which an `Arc` or a `RwLock` can achieve. For example, a `RwLock` is especially designed for many readers and a single writer / many read operations and few write operations.

crates/stackable-webhook/src/servers/conversion.rs
Techassi: The comment of the field perfectly sums this up:

> The environment the operator is running in, notably the namespace and service name it is reachable at.

Currently, we only need the namespace and the service name. That's it. Why are we optimizing for something which _might_ or _might not_ come in the future? Regarding how the crates are currently structured, I would like to avoid more entanglement between those.

As discussed multiple times already, this needs better structure in the long run, which I can hopefully pick up soon.

> I also really don't like too many function arguments

There is only a single parameter: `ConversionWebhookOptions`. Everything is contained in there. And having two fields instead of one, doesn't really justify mixing even more types between crates (which are not designed for that currently).

crates/stackable-webhook/src/tls/cert_resolver.rs
sbernauer: Pls have a look at 39fe001

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: Switched to separate namespace and service_name in 568cda7

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: Same as https://github.com/stackabletech/operator-rs/pull/1066#discussion_r2273350790

crates/stackable-webhook/src/tls/cert_resolver.rs
sbernauer: I had a look how the axum-server crates solved this, as it does basically what we want to do here:
https://github.com/programatik29/axum-server/blob/f60b6150bba0aecf4910f877fd0bc12ac24d030b/src/tls_rustls/mod.rs#L159
I didn't know ArcSwap before, but it basically should offer a better performance than RwLock, although that's propably not super important here.
I spiked a switch to RwLock, nothing fancy. *But* I had to use `blocking_read`, which
> This function panics if called within an asynchronous execution context.

I'm not sure if that will happen, and it's currently hard to get a test setup
```diff
diff --git a/crates/stackable-webhook/src/tls/cert_resolver.rs b/crates/stackable-webhook/src/tls/cert_resolver.rs
index 775ab95..228ed0c 100644
--- a/crates/stackable-webhook/src/tls/cert_resolver.rs
+++ b/crates/stackable-webhook/src/tls/cert_resolver.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use arc_swap::ArcSwap;
 use snafu::{OptionExt, ResultExt, Snafu};
 use stackable_certs::{CertificatePairError, ca::CertificateAuthority, keys::ecdsa};
-use tokio::sync::mpsc;
+use tokio::sync::{RwLock, mpsc};
 use tokio_rustls::rustls::{
     crypto::CryptoProvider, server::ResolvesServerCert, sign::CertifiedKey,
 };
@@ -56,7 +56,7 @@ pub enum CertificateResolverError {
 pub struct CertificateResolver {
     /// Using a [`ArcSwap`] (over e.g. [`tokio::sync::RwLock`]), so that we can easily
     /// (and performant) bridge between async write and sync read.
-    current_certified_key: ArcSwap<CertifiedKey>,
+    current_certified_key: RwLock<Arc<CertifiedKey>>,
     subject_alterative_dns_names: Arc<Vec<String>>,
 
     cert_tx: mpsc::Sender<Certificate>,
@@ -73,7 +73,7 @@ impl CertificateResolver {
 
         Ok(Self {
             subject_alterative_dns_names,
-            current_certified_key: ArcSwap::new(certified_key),
+            current_certified_key: RwLock::new(certified_key),
             cert_tx,
         })
     }
@@ -84,7 +84,7 @@ impl CertificateResolver {
                 .await?;
 
         // TODO: Sign the new cert somehow with the old cert. See https://github.com/stackabletech/decisions/issues/56
-        self.current_certified_key.store(certified_key);
+        *self.current_certified_key.write().await = certified_key;
 
         Ok(())
     }
@@ -151,6 +151,6 @@ impl ResolvesServerCert for CertificateResolver {
         &self,
         _client_hello: tokio_rustls::rustls::server::ClientHello<'_>,
     ) -> Option<Arc<tokio_rustls::rustls::sign::CertifiedKey>> {
-        Some(self.current_certified_key.load().clone())
+        Some(Arc::clone(&self.current_certified_key.blocking_read()))
     }
 }
```

sbernauer: Well I guess I was correct...
```
thread 'tokio-runtime-worker' panicked at src/tls/cert_resolver.rs:154:53:
Cannot block the current thread from within a runtime. This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```
So I'd say my comment stands correct, we need to bridge between async and somewhat efficiently.
Using a proven to work solution similar to what axum-server is doing sounds like a good solution to me

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Sure, will post one asap.

Techassi: Okay, well. Apparently you cannot do what I had in mind anymore. Then let's consider this as resolved.

crates/stackable-webhook/src/tls/cert_resolver.rs
Techassi: Ohhhhh, the `CertificateResolver` trait is sync. Well that's interesting. Then let's keep it as is.

crates/stackable-webhook/src/tls/cert_resolver.rs
Techassi: What do you think about this small improvement?

```patch
diff --git a/crates/stackable-webhook/src/tls/cert_resolver.rs b/crates/stackable-webhook/src/tls/cert_resolver.rs
index 775ab951..6320c0ce 100644
--- a/crates/stackable-webhook/src/tls/cert_resolver.rs
+++ b/crates/stackable-webhook/src/tls/cert_resolver.rs
@@ -69,7 +69,8 @@ impl CertificateResolver {
     ) -> Result<Self> {
         let subject_alterative_dns_names = Arc::new(subject_alterative_dns_names);
         let certified_key =
-            Self::generate_new_cert(&cert_tx, subject_alterative_dns_names.clone()).await?;
+            Self::generate_new_certificate_inner(subject_alterative_dns_names.clone(), &cert_tx)
+                .await?;
 
         Ok(Self {
             subject_alterative_dns_names,
@@ -79,9 +80,7 @@ impl CertificateResolver {
     }
 
     pub async fn rotate_certificate(&self) -> Result<()> {
-        let certified_key =
-            Self::generate_new_cert(&self.cert_tx, self.subject_alterative_dns_names.clone())
-                .await?;
+        let certified_key = self.generate_new_certificate().await?;
 
         // TODO: Sign the new cert somehow with the old cert. See https://github.com/stackabletech/decisions/issues/56
         self.current_certified_key.store(certified_key);
@@ -89,6 +88,11 @@ impl CertificateResolver {
         Ok(())
     }
 
+    async fn generate_new_certificate(&self) -> Result<Arc<CertifiedKey>> {
+        let subject_alterative_dns_names = self.subject_alterative_dns_names.clone();
+        Self::generate_new_certificate_inner(subject_alterative_dns_names, &self.cert_tx).await
+    }
+
     /// Creates a new certificate and returns the certified key.
     ///
     /// The certificate is send to the passed `cert_tx`.
@@ -96,9 +100,9 @@ impl CertificateResolver {
     /// FIXME: This should *not* construct a CA cert and cert, but only a cert!
     /// This needs some changes in stackable-certs though.
     /// See [the relevant decision](https://github.com/stackabletech/decisions/issues/56)
-    async fn generate_new_cert(
-        cert_tx: &mpsc::Sender<Certificate>,
+    async fn generate_new_certificate_inner(
         subject_alterative_dns_names: Arc<Vec<String>>,
+        cert_tx: &mpsc::Sender<Certificate>,
     ) -> Result<Arc<CertifiedKey>> {
         // The certificate generations can take a while, so we use `spawn_blocking`
         let (cert, certified_key) = tokio::task::spawn_blocking(move || {

```

sbernauer: works for me, 172ab6d

crates/stackable-webhook/src/tls/mod.rs
sbernauer: done in e3e1119

Techassi (approved):

PR #1072, opened 2025-07-09
chore: Bump Rust version to 1.87.0 and workflow actions

sbernauer:
# Description

Part of https://github.com/stackabletech/operator-rs/issues/1018

No changelog entry, as this should only be repo-internal changes.
Lot's of new clippy lints to fix.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/product_config_utils.rs
+++ b/crates/stackable-operator/src/product_config_utils.rs
@@ -160,7 +160,6 @@ pub fn config_for_role_and_group<'a>(
     Ok(result)
 }
 
-#[allow(clippy::doc_overindented_list_items)]
 /// Given the configuration parameters of all `roles` partition them by `PropertyNameKind` and
 /// merge them with the role groups configuration parameters.
 ///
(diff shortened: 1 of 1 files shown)


crates/stackable-operator/src/product_config_utils.rs
```diff
@@ -160,6 +160,7 @@ pub fn config_for_role_and_group<'a>(
     Ok(result)
 }
 
+#[allow(clippy::doc_overindented_list_items)]
```
Techassi: note: Don't allow that. We instead should use `:` instead of `-` for the list of arguments.

```rust
/// Given the configuration parameters of all `roles` partition them by `PropertyNameKind` and
/// merge them with the role groups configuration parameters.
///
/// The output is a map keyed by the role names. The value is also a map keyed by role group names
/// and the values are the merged configuration properties "bucketed" by `PropertyNameKind`.
///
/// # Arguments
/// - `resource`: Not used directly. It's passed on to the `Configuration::compute_*` calls.
/// - `roles`: A map keyed by role names. The value is a tuple of a vector of `PropertyNameKind`
///   like (Cli, Env or Files) and [`crate::role_utils::Role`] with a boxed [`Configuration`].
```

Techassi (changes requested):
I'm not a huge fan of all these Boxes, but oh well, what can you do. This can only be fixed upstream (if possible at all).

I left one comment requesting changes.

Everything else looks good.

crates/stackable-operator/src/product_config_utils.rs
sbernauer: a68427f

sbernauer:
> I would also like us to bump to Rust 1.88.0 instead.

I would prefer that as well, but nixpkgs unstable is currently on 1.87, which is used in `make run-dev`.
I personally think it's not worth the extra effort to fiddle around with the niv/nixpkgs/crate2nix/whatnot setup.
We can always bump to 1.88 as needed later, unrelated to the SDP release cycle.

We already have a discussion about this in https://stackable-workspace.slack.com/archives/C02FZ581UCD/p1752043094194079?thread_ts=1752042485.505039&cid=C02FZ581UCD

Techassi:
@NickLarsenNZ and I decided to move forward with Fenix after the 25.7.0 release. Once that transition is done, we can always bump to the latest available stable Rust version without the need to wait for upstream Nixpkgs.

For now, we can keep it at 1.87.0.

Techassi (approved):
Approved as is, but we should invest some time on reducing the error sizes down the line. Introducing these Boxes all over the place makes the code harder to read and maintain.

PR #1074, opened 2025-07-10
feat!: Re-export stackable-certs and stackable-webhook crates

sbernauer:
# Description

As we want to start using stackable-webhook with https://github.com/stackabletech/operator-rs/pull/1066, we want to re-export it the same way as the other crates we have.
For that to work we need to resolve cyclic decencies `operator -> certs -> operator` and `operator -> webhook -> operator`.
I achieved this by moving common time and SecretRef code into stackable_shared.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance


crates/stackable-shared/src/commons/mod.rs
Techassi, line 1: I think we should get rid of the `commons` module. The imports would look like this then: `stackable_shared::secret::SecretRef`.

crates/stackable-shared/src/time/mod.rs
```diff
@@ -1,7 +1,6 @@
 mod duration;
 mod serde_impl;
 
-#[cfg(feature = "time")]
```
Techassi: note: I think this should still be an opt-in feature.

crates/stackable-shared/src/commons/mod.rs
sbernauer: 36c63b3

crates/stackable-shared/src/time/mod.rs
sbernauer: 36c63b3

Techassi: The feature gate is in the wrong place. The `time` module is always there, only the `time_impl` module (which contains impls for the `time` crate) should be gated by this.

This feature also needs to be forwarded in `stackable_operator` as it re-exports the shared crate.

sbernauer: Ah I see. Should be fixed in f8105ef

Techassi (approved):

PR #1078, opened 2025-08-13
feat: Add ProbeBuilder

sbernauer:
# Description

Adds a builder for a `Probe` similar to the `ResourceRequirementsBuilder`
Needed for https://github.com/stackabletech/superset-operator/pull/654

### Author

- [x] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/builder/pod/probe.rs
+++ b/crates/stackable-operator/src/builder/pod/probe.rs
@@ -1,3 +1,34 @@
+//! Kubernetes [`Probe`] builder.
+//!
+//! The upstream [`Probe`] struct does not prevent invalid probe configurations
+//! which leads to surprises at runtime which can be deeply hidden.
+//! You need to specify at least an action and interval (in this order).
+//!
+//! ### Usage example
+//!
+//! ```
+//! use stackable_operator::{
+//!     builder::pod::probe::ProbeBuilder,
+//!     time::Duration,
+//! };
+//! # use k8s_openapi::api::core::v1::HTTPGetAction;
+//! # use k8s_openapi::apimachinery::pkg::util::intstr::IntOrString;
+//!
+//! let probe = ProbeBuilder::http_get_port_scheme_path(8080, None, None)
+//!     .with_period(Duration::from_secs(10))
+//!     .build()
+//!     .expect("failed to build probe");
+//!
+//! assert_eq!(
+//!     probe.http_get,
+//!     Some(HTTPGetAction {
+//!         port: IntOrString::Int(8080),
+//!         ..Default::default()
+//!     })
+//! );
+//! assert_eq!(probe.period_seconds, Some(10));
+//! ```
+
 use std::num::TryFromIntError;
 
 use k8s_openapi::{
@@ -23,36 +54,6 @@ pub enum Error {
     PeriodIsZero {},
 }
 
-/// Kubernetes [`Probe`] builder.
-///
-/// The upstream [`Probe`] struct does not prevent invalid probe configurations
-/// which leads to surprises at runtime which can be deeply hidden.
-/// You need to specify at least an action and interval (in this order).
-///
-/// ### Usage example
-///
-/// ```
-/// use stackable_operator::{
-///     builder::pod::probe::ProbeBuilder,
-///     time::Duration,
-/// };
-/// # use k8s_openapi::api::core::v1::HTTPGetAction;
-/// # use k8s_openapi::apimachinery::pkg::util::intstr::IntOrString;
-///
-/// let probe = ProbeBuilder::http_get_port_scheme_path(8080, None, None)
-///     .with_period(Duration::from_secs(10))
-///     .build()
-///     .expect("failed to build probe");
-///
-/// assert_eq!(
-///     probe.http_get,
-///     Some(HTTPGetAction {
-///         port: IntOrString::Int(8080),
-///         ..Default::default()
-///     })
-/// );
-/// assert_eq!(probe.period_seconds, Some(10));
-/// ```
 #[derive(Debug)]
 pub struct ProbeBuilder<Action, Period> {
     // Mandatory field


crates/stackable-operator/src/builder/pod/probe.rs
```diff
@@ -0,0 +1,330 @@
...
+        field: String,
+        duration: Duration,
+    },
+
+    #[snafu(display("The probe period is zero, but it needs to be a positive duration"))]
+    PeriodIsZero {},
+}
+
+#[derive(Debug)]
+pub struct ProbeBuilder<Action, Period> {
```
NickLarsenNZ, line 58: I do think we should have docs on pub items.

It doesn't have to be super comprehensive, and examples can be added to the module level docs.

```suggestion
/// Valid Kubernetes probe builder
///
/// The upstream struct [NameOfStruct] does not prevent invalid probe configurations
/// which leads to surprises at runtime which can be deeply hidden.
#[derive(Debug)]
pub struct ProbeBuilder<Action, Period> {
```

crates/stackable-operator/src/builder/pod/probe.rs
```diff
@@ -0,0 +1,330 @@
...
+    apimachinery::pkg::util::intstr::IntOrString,
+};
+use snafu::{ResultExt, Snafu, ensure};
+
+use crate::time::Duration;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display(
+        "The probe's {field:?} duration of {duration} is too long, as it's seconds doesn't fit into an i32"
```
NickLarsenNZ, line 45: How does that appear? Does it look like human time?
Eg:
- `duration of 4,734e+8 is too long`
- `duration of 15 is too long`, or
- `duration of 15 years is too long`

crates/stackable-operator/src/builder/pod/probe.rs
```diff
@@ -0,0 +1,330 @@
...
+            success_threshold: 1,
+            failure_threshold: 1,
+            timeout: Duration::from_secs(1),
+            initial_delay: Duration::from_secs(0),
+            termination_grace_period: Duration::from_secs(0),
+        }
+    }
+}
+
+pub enum ProbeAction {
```
NickLarsenNZ, line 77: To save documenting each variant, we can refer to upstream docs.

```suggestion
/// Available probes
///
/// Only one probe can be configured at a time. For more details about each 
/// type, see [container-probes] documentation.
///
/// [container-probes]: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes
pub enum ProbeAction {
```

crates/stackable-operator/src/builder/pod/probe.rs
```diff
@@ -0,0 +1,330 @@
...
+        self,
+        command: impl IntoIterator<Item = impl Into<String>>,
+    ) -> ProbeBuilder<ProbeAction, ()> {
+        self.with_exec_action(ExecAction {
+            command: Some(command.into_iter().map(Into::into).collect()),
+        })
+    }
+
+    /// There is a convenience helper: [`Self::with_exec_action_helper`].
+    pub fn with_exec_action(self, exec_action: ExecAction) -> ProbeBuilder<ProbeAction, ()> {
```
NickLarsenNZ:
```suggestion
    /// Set an exec probe with the given configuration.
    ///
    /// For simple cases, there is a a convenience helper: [`Self::with_exec_action_helper`].
    pub fn with_exec_action(self, exec_action: ExecAction) -> ProbeBuilder<ProbeAction, ()> {
```

NickLarsenNZ: I'll stop adding documenting comments, but this should apply to other pub methods too.

crates/stackable-operator/src/builder/pod/probe.rs
```diff
@@ -0,0 +1,330 @@
...
+            })
+        );
+        assert_eq!(probe.period_seconds, Some(10));
+    }
+
+    #[test]
+    fn test_probe_builder_complex() {
+        let probe = ProbeBuilder::default()
+            .with_exec_action_helper(["sleep", "1"])
+            .with_period(Duration::from_secs(5))
```
NickLarsenNZ: Part of me feels like the word "helper" should be replaced with what it does.

But actually, I have a different recommendation overall...

Since there are a small finite set of probe types that can be created, we should drop the `default` impl... and make the developer choose one...

```rust
        let probe = ProbeBuilder::exec_with_command(["sleep", "1"])
            .with_period(Duration::from_secs(5))
```

and for the non "helper" case, it could just be:

```rust
        let probe = ProbeBuilder::exec(the_exec_stuff)
            .with_period(Duration::from_secs(5))
```

NickLarsenNZ: To expand on things... I don't believe there is a good "default" probe, and therefore there shouldn't be a default method.

NickLarsenNZ (commented):
Mostly reviewed, but I'd like to discuss the change I commented on.

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: ecc2373

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: https://github.com/stackabletech/operator-rs/blob/34588a2bed6a85701db6a25fab65648c8c65878c/crates/stackable-operator/src/time/duration.rs#L178-200

I guess something like `The probe's "terminationGracePeriod" duration of 42d is too long` or `15d16h3s`

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: The default impl is only for the *Builder*, you are correct that there is no such thing as a default probe or default period.
clippy suggests adding default() over new():
```bash
warning: you should consider adding a `Default` implementation for `ProbeBuilder<(), ()>`
  --> crates/stackable-operator/src/builder/pod/probe.rs:72:5
   |
72 | /     pub fn new() -> Self {
73 | |         Self {
74 | |             action: (),
75 | |             period: (),
...  |
83 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default
   = note: `#[warn(clippy::new_without_default)]` on by default
help: try adding this
   |
71 + impl Default for ProbeBuilder<(), ()> {
72 +     fn default() -> Self {
73 +         Self::new()
74 +     }
75 + }
```
But yeah, I will think about starting with the action functions directly

sbernauer: 7779e12

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: 7779e12

crates/stackable-operator/src/builder/pod/probe.rs
NickLarsenNZ: I would suggest making it module level (`//!`), but I actuall like that I can hover over the struct and see the usage straight away.

_You can resolve this comment after reading_

crates/stackable-operator/src/builder/pod/probe.rs
NickLarsenNZ: This was marked as resolved, but no link to the change :shrug: 
I'll reopen it in case it was accidental.

crates/stackable-operator/src/builder/pod/probe.rs
NickLarsenNZ: One other thing I neglected to say is that the common pattern for builders is where the end developer doesn't import/use the *Builder struct, but instead starts with the actual thing they are building.

To illustrate:

## Wrong (or against common pattern):

```rust
let probe = ProbeBuilder::exec_command(['rm', '-fr', '/']).build();
```

## Correct

```rust
let probe = Probe::exec_command_builder(['rm', '-fr', '/']).build();
```

---

That would require a trait (eg: `ProbeExt` and import as `_` to get the methods on `Probe`) and impl onto Probe, so I understand if you don't want to change things this late in the PR. Worth considering for future builders though.

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: That's normally when you click `Commit suggestion` on GitHub ;)
It picked it in 55f20c698e95670837727de0e820f9e9560f566f

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: No opinion, moved to module docs in fd5a099

crates/stackable-operator/src/builder/pod/probe.rs
sbernauer: I was thinking of the same problem as I started. I also had the same problem with `bon`.
I'm not sure if an extension trait hurts the usability, as users need to be aware that there is such a thing. Same as with the current builder...
I agree with you, but also happy to ship like this for now

NickLarsenNZ: I agree, let's ship it.

If it is important enough, I can refactor later :+1:

PR #1086, opened 2025-09-02
fix(stackable-versioned): Correctly emit enum fields in From impls

Techassi:
This PR correctly emits enum variant fields in `From` impl blocks. Encountered in https://github.com/stackabletech/secret-operator/pull/634.


Techassi:
> I'm not sure I can follow the code 100%, but we have a test 😅

Is there anything I can help with? I can add a few comments to the code to explain what exactly it generates.

sbernauer:
Looking at `crates/stackable-versioned-macros/tests/snapshots/stackable_versioned_macros__snapshots__pass@enum_fields.rs.snap` it all makes sense, happy to merge as-is

Techassi:
Alright, cool.

PR #1088, opened 2025-09-04
refactor(stackable-versioned): Introduce idents for fields/variants

Techassi:
Part of https://github.com/stackabletech/secret-operator/issues/627 and https://github.com/stackabletech/secret-operator/pull/634.

This PR finally moves all variant and field idents into a single place instead of formatting the same idents at multiple different places. This is in preparation to **support nested tracking in enums**, which is needed for the SecretClass.

### DIFF
--- a/crates/stackable-versioned-macros/src/utils/mod.rs
+++ b/crates/stackable-versioned-macros/src/utils/mod.rs
@@ -41,7 +41,10 @@ impl ContainerIdentExt for Ident {
     }
 
     fn as_parameter_ident(&self) -> IdentString {
-        let ident = format_ident!("__sv_{}", self.to_string().to_lowercase());
+        let ident = format_ident!(
+            "__sv_{lowercase_ident}",
+            lowercase_ident = self.to_string().to_lowercase()
+        );
         IdentString::new(ident)
     }
 }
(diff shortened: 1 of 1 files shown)


crates/stackable-versioned-macros/src/utils/mod.rs
```diff
@@ -58,82 +56,30 @@ impl ContainerIdentExt for IdentString {
...
+pub trait ItemIdentExt {
+    fn json_path_ident(&self) -> IdentString;
 }
 
-impl ToTokens for VariantIdent {
-    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
-        self.0.to_tokens(tokens);
+impl ItemIdentExt for IdentString {
+    fn json_path_ident(&self) -> IdentString {
+        format_ident!("__sv_{}_path", self.as_str().to_lowercase()).into()
```
NickLarsenNZ: I think it's nicer to give the interpolation a name to lower cognitive load

```suggestion
        format_ident!("__sv_{ident_name}_path", ident_name = self.as_str().to_lowercase()).into()
```

or similar (`cargo fmt` would probably change that in some way).

NickLarsenNZ (dismissed):
I skim-read it, looks fine to me.

crates/stackable-versioned-macros/src/utils/mod.rs
Techassi: Oh yeah, that makes sense. I will update it.

Techassi: Done in b50a0644.

NickLarsenNZ (approved):
LGTM

PR #1089, opened 2025-09-05
feat(stackable-versioned): Add argument to hint for wrapped types

Techassi:
Part of https://github.com/stackabletech/issues/issues/642 and https://github.com/stackabletech/secret-operator/pull/634.

This PR adds a new argument `#[versioned(hint)]` to provide hints for two supported wrapped types:

- `#[versioned(hint(option))]`: This hints that the field is wrapped in `Option<_>`. In the conversion `From` impl blocks this will result in a `.map(Into::into)` call.
- `#[versioned(hint(vec))]`: This hints that the field is wrapped in `Vec<_>`. In the conversion `From` impl blocks this will result in a `.into_iter().map(Into::into).collect()` call.

### DIFF
--- a/crates/stackable-versioned/CHANGELOG.md
+++ b/crates/stackable-versioned/CHANGELOG.md
@@ -4,11 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Add new `#[versioned(hint)]` argument to provide type hints for struct fields ([#1089]).
+  - `#[versioned(hint(option))]` for fields wrapped in an `Option`. Generates `.map(Into::into)`.
+  - `#[versioned(hint(vec))]` for fields wrapped in an `Vec`. Generates `.into_iter().map(Into::into).collect()`.
+
 ### Fixed
 
 - Correctly emit enum variant fields in `From` impl blocks ([#1086]).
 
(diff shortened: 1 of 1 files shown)


crates/stackable-versioned-macros/src/codegen/item/field.rs
```diff
@@ -60,19 +62,31 @@ impl VersionedField {
...
+    ///
+    /// - Only emit the field if it exists for the currently generated version.
+    /// - Emit field with new name and type if there was a name and/or type change.
+    /// - Handle deprecated fields accordingly.
+    ///
+    /// ### Example
+    ///
+    /// ```ignore
+    /// struct Foo {
+    ///     bar: usize, // < This functions generates one or more of these definitions
```
NickLarsenNZ:
```suggestion
    ///     bar: usize, // This function generates one or more of these definitions
```

NickLarsenNZ (approved):
I skim-read this one. Looks fine to me.

One suggestion (will approve anyway since it is a non-functional thing).

crates/stackable-versioned-macros/src/codegen/item/field.rs
Techassi: Done in d4e9478b.

NickLarsenNZ (approved):
LGTM

PR #1090, opened 2025-09-08
fix: Invalid termination_grace_period default in ProbeBuilder

sbernauer:
# Description

Noticed in https://testing.stackable.tech/job/superset-operator-it-weekly/69

`create Pod superset-node-default-0 in StatefulSet superset-node-default failed error: Pod "superset-node-default-0" is invalid: [spec.containers[0].livenessProbe.terminationGracePeriodSeconds: Invalid value: 0: must be greater than 0, spec.containers[0].readinessProbe.terminationGracePeriodSeconds: Invalid value: 0: must be greater than 0, spec.containers[0].readinessProbe.terminationGracePeriodSeconds: Invalid value: 0: must not be set for readinessProbes, spec.containers[0].startupProbe.terminationGracePeriodSeconds: Invalid value: 0: must be greater than 0]`

Also copy some docs from k8s docs

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

### Author

- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Integration tests passed (for non trivial changes)
- [ ] Changes need to be "offline" compatible

### Reviewer

- [ ] Code contains useful comments
- [ ] Code contains useful logging statements
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)

### Acceptance

- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added

PR #1085, opened 2025-08-27
feat!: Add a new CLI flag/env to disabling CRD maintenance

sbernauer:
# Description

We got feedback that some customers value breaking their cluster over giving operators RBAC permissions for CRDs :D
So WCGW?

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

### Author

- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Integration tests passed (for non trivial changes)
- [ ] Changes need to be "offline" compatible

### Reviewer

- [ ] Code contains useful comments
- [ ] Code contains useful logging statements
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)

### Acceptance

- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added

PR #1092, opened 2025-09-09
chore: Release workspace members (stackable-operator 0.97.0)

sbernauer:
### Added

- BREAKING: Add a new CLI flag/env to disabling CRD maintenance: `--disable-crd-maintenance` ([#1085]).

### Removed

- BREAKING: Remove the Merge implementation for PodTemplateSpec ([#1087]).
  It was broken because the instance was overridden by the given defaults.
  This function is not used by the Stackable operators.

### Fixed

- Don't default the `termination_grace_period` of the `ProbeBuilder` to 0, as this is an invalid value ([#1090]).

[#1085]: https://github.com/stackabletech/operator-rs/pull/1085
[#1087]: https://github.com/stackabletech/operator-rs/pull/1087
[#1090]: https://github.com/stackabletech/operator-rs/pull/1090

### Added

- Add new `#[versioned(hint)]` argument to provide type hints for struct fields ([#1089]).
  - `#[versioned(hint(option))]` for fields wrapped in an `Option`. Generates `.map(Into::into)`.
  - `#[versioned(hint(vec))]` for fields wrapped in an `Vec`. Generates `.into_iter().map(Into::into).collect()`.

### Fixed

- Correctly emit enum variant fields in `From` impl blocks ([#1086]).

[#1086]: https://github.com/stackabletech/operator-rs/pull/1086
[#1089]: https://github.com/stackabletech/operator-rs/pull/1089

### Added

- BREAKING: Support disabling CRD maintenance using a new boolean flag in `ConversionWebhookOptions` ([#1085]).

[#1085]: https://github.com/stackabletech/operator-rs/pull/1085

### DIFF
--- a/crates/stackable-versioned/CHANGELOG.md
+++ b/crates/stackable-versioned/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
-## [0.9.0] - 2025-09-09
+## [0.8.2] - 2025-09-09
 
 ### Added
 
--- a/crates/stackable-versioned/CHANGELOG.md
+++ b/crates/stackable-versioned/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [0.9.0] - 2025-09-09
+
 ### Added
 
 - Add new `#[versioned(hint)]` argument to provide type hints for struct fields ([#1089]).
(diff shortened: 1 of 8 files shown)


crates/stackable-versioned/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [0.9.0] - 2025-09-09
```
Techassi: I feel like this should be `0.8.2`.

sbernauer: 07e941d

PR #1099, opened 2025-09-23
feat!: Add CRD maintainer

Techassi:
Part of https://github.com/stackabletech/secret-operator/pull/634 and https://github.com/stackabletech/decisions/issues/62.

This PR adds a new CRD maintainer which reconciles the CRDs when the conversion webhook server certificate is rotated. It additionally provides a oneshot channel to indicate that the initial CRD rollout is done and default custom resources can be applied.

This is a crude first implementation of the CRD and default CR maintenance mechanism. This PR will be used as a test-balloon in https://github.com/stackabletech/secret-operator/pull/634.

In the future, this mechanism should be implemented in a more generic way, similar to `kube::Controller`.

### DIFF
--- a/crates/stackable-webhook/src/maintainer.rs
+++ b/crates/stackable-webhook/src/maintainer.rs
@@ -94,7 +94,7 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
     /// # async fn main() {
     /// # let (certificate_tx, certificate_rx) = channel(1);
     /// let options = CustomResourceDefinitionMaintainerOptions {
-    ///     operator_name: "my-service-name",
+    ///     operator_service_name: "my-service-name",
     ///     operator_namespace: "my-namespace",
     ///     field_manager: "my-field-manager",
     ///     webhook_https_port: 8443,
@@ -142,9 +142,9 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
     /// [`std::task::Poll::Ready`] and thus doesn't consume any resources.
     pub async fn run(mut self) -> Result<(), Error> {
         let CustomResourceDefinitionMaintainerOptions {
+            operator_service_name,
             operator_namespace,
             webhook_https_port,
-            operator_name,
             field_manager,
             disabled,
         } = self.options;
@@ -198,7 +198,7 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
                         conversion_review_versions: vec!["v1".to_owned()],
                         client_config: Some(WebhookClientConfig {
                             service: Some(ServiceReference {
-                                name: operator_name.to_owned(),
+                                name: operator_service_name.to_owned(),
                                 namespace: operator_namespace.to_owned(),
                                 path: Some(format!("/convert/{crd_name}")),
                                 port: Some(webhook_https_port.into()),
@@ -237,8 +237,8 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
 // TODO (@Techassi): Make this a builder instead
 /// This contains required options to customize a [`CustomResourceDefinitionMaintainer`].
 pub struct CustomResourceDefinitionMaintainerOptions<'a> {
-    /// The service name used by the operator/conversion webhook and as a field manager.
-    pub operator_name: &'a str,
+    /// The service name used by the operator/conversion webhook.
+    pub operator_service_name: &'a str,
 
     /// The namespace the operator/conversion webhook runs in.
     pub operator_namespace: &'a str,
--- a/crates/stackable-webhook/src/servers/conversion.rs
+++ b/crates/stackable-webhook/src/servers/conversion.rs
@@ -195,10 +195,11 @@ impl ConversionWebhookServer {
     ///   to a handler function. In most cases, the generated `CustomResource::try_merge` function
     ///   should be used. It provides the expected `fn(ConversionReview) -> ConversionReview`
     ///   signature.
-    /// - `operator_name`: The name of the operator. This is used to construct the webhook service
-    ///   name.
+    /// - `operator_service_name`: The name of the Kubernetes service name which points to the
+    ///   operator/conversion webhook. This is used to construct the service reference in the CRD
+    ///   `spec.conversion` field.
     /// - `operator_namespace`: The namespace the operator runs in. This is used to construct the
-    ///   webhook service name.
+    ///   service reference in the CRD `spec.conversion` field.
     /// - `disable_maintainer`: A boolean value to indicate if the [`CustomResourceDefinitionMaintainer`]
     ///   should be disabled.
     /// - `client`: A [`kube::Client`] used to maintain the custom resource definitions.
@@ -264,7 +265,7 @@ impl ConversionWebhookServer {
         // TODO (@Techassi): Use a trait type here which can be used to build all part of the
         // conversion webhook server and a CRD maintainer.
         crds_and_handlers: impl IntoIterator<Item = (CustomResourceDefinition, H)> + Clone,
-        operator_name: &'a str,
+        operator_service_name: &'a str,
         operator_namespace: &'a str,
         field_manager: &'a str,
         disable_maintainer: bool,
@@ -284,8 +285,8 @@ impl ConversionWebhookServer {
 
         // TODO (@Techassi): These should be moved into a builder
         let webhook_options = ConversionWebhookOptions {
+            service_name: operator_service_name,
             namespace: operator_namespace,
-            service_name: operator_name,
             socket_addr,
         };
 
@@ -298,8 +299,8 @@ impl ConversionWebhookServer {
         let maintainer_options = CustomResourceDefinitionMaintainerOptions {
             webhook_https_port: socket_addr.port(),
             disabled: disable_maintainer,
+            operator_service_name,
             operator_namespace,
-            operator_name,
             field_manager,
         };
 


crates/stackable-operator/src/client.rs
```diff
@@ -253,6 +253,25 @@ impl Client {
             })
     }
 
+    /// Optionally creates a resource if it does not exist yet.
+    ///
+    /// The name used for lookup is extracted from the resource via [`ResourceExt::name_any()`].
+    /// This function either returns the existing resource or the newly created one.
+    pub async fn create_opt<T>(&self, resource: &T) -> Result<T>
```
sbernauer: I think I would prefer a more speaking name such as `create_if_missing`.
`create_opt` could also mean "i create (or not) based on the phase of the moon" or on some other factor

Techassi: Done in 2fc5c28d.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Add `CustomResourceDefinitionMaintainer` which applies and patches CRDs triggered by TLS
+  certificate rotations of the `ConversionWebhookServer`. It additionally provides a `oneshot`
+  channel which can be used to trigger creation/patching of any default custom resources deployed by
```
sbernauer: I would prefer to not "hardcode" default here, it can also be used for other stuff
```suggestion
  channel which can for example be used to trigger creation/patching of any custom resources deployed by
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -86,6 +87,15 @@ pub struct WebhookServer {
 }
 
 impl WebhookServer {
+    /// The default HTTPS port `8443`
+    pub const DEFAULT_HTTPS_PORT: u16 = 8443;
+    /// The default IP address [`Ipv4Addr::UNSPECIFIED`] (`0.0.0.0`) the webhook server binds to,
+    /// which represents binding on all network addresses.
+    pub const DEFAULT_LISTEN_ADDRESS: IpAddr = IpAddr::V4(Ipv4Addr::UNSPECIFIED);
```
sbernauer, line 97: I found this comment helpful and would prefer to keep it (it e.g. actually hit me in trino-lb)
```suggestion
    //
    // TODO: We might want to switch to `Ipv6Addr::UNSPECIFIED)` here, as this *normally* binds to IPv4
    // and IPv6. However, it's complicated and depends on the underlying system...
    // If we do so, we should set `set_only_v6(false)` on the socket to not rely on system defaults.
    pub const DEFAULT_LISTEN_ADDRESS: IpAddr = IpAddr::V4(Ipv4Addr::UNSPECIFIED);
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -35,12 +37,11 @@ use tokio::{
     sync::mpsc,
 };
 use tower::ServiceBuilder;
-use x509_cert::Certificate;
+pub use x509_cert::Certificate;
```
sbernauer: Should we really be re-exporting this? I see a point for kube, k8s-openapi and such but this looks a bit excessive. But you might have total valid reasons ;)

crates/stackable-webhook/src/maintainer.rs
sbernauer, line 1: Hmm, I honestly would have expected this to be part of stackable-webhook instead of stackable-operator.
(At least currently) only the webhook cares about maintaining the CRD. And if people would like to use Rust CRD versioning the stackable-webhook crate would be self-contained.

crates/stackable-operator/src/crd/maintainer.rs
```diff
@@ -0,0 +1,213 @@
...
+/// - Apply the CRDs when starting up
+/// - Reconcile the CRDs when the conversion webhook certificate is rotated
+pub struct CustomResourceDefinitionMaintainer {
+    client: Client,
+    certificate_rx: mpsc::Receiver<Certificate>,
+
+    definitions: Vec<CustomResourceDefinition>,
+    options: CustomResourceDefinitionMaintainerOptions,
+
+    initial_reconcile_tx: Option<oneshot::Sender<()>>,
```
sbernauer: I was really confused about this being optional. Than I read the comment further down.
I would prefer to keep the logic that we only do it on the first loop iterator within the `run` function:
```diff

diff --git a/crates/stackable-operator/src/crd/maintainer.rs b/crates/stackable-operator/src/crd/maintainer.rs
index 6c9041c..ad5217f 100644
--- a/crates/stackable-operator/src/crd/maintainer.rs
+++ b/crates/stackable-operator/src/crd/maintainer.rs
@@ -44,7 +44,7 @@ pub struct CustomResourceDefinitionMaintainer {
     definitions: Vec<CustomResourceDefinition>,
     options: CustomResourceDefinitionMaintainerOptions,
 
-    initial_reconcile_tx: Option<oneshot::Sender<()>>,
+    initial_reconcile_tx: oneshot::Sender<()>,
 }
 
 impl CustomResourceDefinitionMaintainer {
@@ -84,7 +84,6 @@ impl CustomResourceDefinitionMaintainer {
         options: CustomResourceDefinitionMaintainerOptions,
     ) -> (Self, oneshot::Receiver<()>) {
         let (initial_reconcile_tx, initial_reconcile_rx) = oneshot::channel();
-        let initial_reconcile_tx = Some(initial_reconcile_tx);
 
         let maintainer = Self {
             definitions: definitions.into_iter().collect(),
@@ -110,6 +109,7 @@ impl CustomResourceDefinitionMaintainer {
             webhook_https_port: https_port,
             disabled,
         } = self.options;
+        let mut initial_reconcile_tx = Some(self.initial_reconcile_tx);
 
         // If the maintainer is disabled or there are no custom resource definitions, immediately
         // return without doing any work.
@@ -181,7 +181,7 @@ impl CustomResourceDefinitionMaintainer {
             // via the oneshot channel. This channel can only be used exactly once. The sender's
             // send method consumes self, and as such, the sender is wrapped in an Option to be
             // able to call take to consume the inner value.
-            if let Some(initial_reconcile_tx) = self.initial_reconcile_tx.take() {
+            if let Some(initial_reconcile_tx) = initial_reconcile_tx.take() {
                 match initial_reconcile_tx.send(()) {
                     Ok(_) => {}
                     Err(_) => return SendInitialReconcileHeartbeatSnafu.fail(),
```

crates/stackable-operator/Cargo.toml
```diff
@@ -12,11 +12,11 @@ full = ["certs", "telemetry", "versioned", "time", "webhook", "clap"]
 default = ["telemetry", "versioned", "clap"]
 
 clap = []
-certs = ["dep:stackable-certs"]
+certs = ["dep:stackable-certs", "dep:x509-cert"]
```
sbernauer: marker: If we move stuff into stackable-webhook this can likely be dropped

crates/stackable-operator/src/crd/maintainer.rs
```diff
@@ -0,0 +1,213 @@
...
+
+            // After the reconciliation of the CRDs, the initial reconcile heartbeat is sent out
+            // via the oneshot channel. This channel can only be used exactly once. The sender's
+            // send method consumes self, and as such, the sender is wrapped in an Option to be
+            // able to call take to consume the inner value.
+            if let Some(initial_reconcile_tx) = self.initial_reconcile_tx.take() {
+                match initial_reconcile_tx.send(()) {
+                    Ok(_) => {}
+                    Err(_) => return SendInitialReconcileHeartbeatSnafu.fail(),
+                }
```
sbernauer: optional: We could use the ensure! macro for this
```suggestion
                ensure!(
                    initial_reconcile_tx.send(()).is_ok(),
                    SendInitialReconcileHeartbeatSnafu
                );
```

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -200,34 +155,15 @@ impl ConversionWebhookServer {
...
             socket_addr,
             namespace: operator_namespace,
             service_name: operator_service_name,
-            maintain_crds,
-            field_manager,
         } = &options;
 
         // This is how Kubernetes calls us, so it decides about the naming.
         // AFAIK we can not influence this, so this is the only SAN entry needed.
+        // FIXME (@Techassi): The cluster domain should be included here to form FQDN of the service
```
sbernauer: I think I would not call it a fixme, but maybe something like
`TODO (@Techassi): The cluster domain should be included here, so that (non Kubernetes) HTTP clients can also use the FQDN of the service in case they e.g. want to test the endpoint`

crates/stackable-webhook/src/maintainer.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// ## Return Values
+    ///
+    /// This function returns a 2-tuple (pair) of values:
+    ///
+    /// - The [`CustomResourceDefinitionMaintainer`] itself. This is used to run the maintainer.
+    ///   See [`CustomResourceDefinitionMaintainer::run`] for more details.
+    /// - The [`oneshot::Receiver`] which will be used to send out a message once the initial
+    ///   CRD reconciliation ran. This signal can be used to trigger the deployment of custom
+    ///   resources defined by the maintained CRDs.
+    pub fn new(
```
sbernauer, line 119: Can you please add an example code snippet similar to `ConversionWebhookServer::new`?

sbernauer (changes requested):
Is there any example code in secret-op I can look at?
At a first glance it looks like https://github.com/stackabletech/secret-operator/pull/634 is not using it yet

crates/stackable-webhook/src/maintainer.rs
Techassi: In my books, CRD maintenance is not related to webhooks at all. There are other reasons to maintain CRDs other than rotating the caBundle for conversion webhooks. But I can see how this would simplify the API interface.

I will think about what to do.

crates/stackable-webhook/src/lib.rs
Techassi: This is a tricky thing. Ideally we want to hide the underlying type from `x509_cert`, because that leaks abstraction details, but this is currently not possible.

I will also think about this.

crates/stackable-operator/src/crd/maintainer.rs
Techassi: Sure, that should be possible.

crates/stackable-webhook/src/maintainer.rs
Techassi: Sure, I will see what I can do.

crates/stackable-operator/src/crd/maintainer.rs
Techassi: Done in ac5b9cbe.

crates/stackable-webhook/src/maintainer.rs
Techassi: Added in bbf49f5c.

crates/stackable-webhook/src/lib.rs
Techassi: Improved in 106dd162.

crates/stackable-operator/Cargo.toml
Techassi: Improved in 106dd162.

crates/stackable-operator/Cargo.toml
```diff
@@ -24,7 +24,7 @@ stackable-operator-derive = { path = "../stackable-operator-derive" }
 stackable-shared = { path = "../stackable-shared" }
 stackable-telemetry = { path = "../stackable-telemetry", optional = true, features = ["clap"] }
 stackable-versioned = { path = "../stackable-versioned", optional = true }
-stackable-webhook = { path = "../stackable-webhook", optional = true }
+stackable-webhook = { path = "../stackable-webhook", optional = true, features = ["maintainer"]}
```
sbernauer:
```suggestion
stackable-webhook = { path = "../stackable-webhook", optional = true, features = ["maintainer"] }
```

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Updated in d0c4a57c.

crates/stackable-webhook/src/maintainer.rs
Techassi: I thought about it, and I want to keep it located where it is now. These parts will definitely be moved around and improved down the line. Real-world usage will teach us these improvements.

crates/stackable-operator/Cargo.toml
Techassi: This got removed in 99eae00f.

crates/stackable-operator/Cargo.toml
```diff
@@ -24,7 +24,7 @@ stackable-operator-derive = { path = "../stackable-operator-derive" }
 stackable-shared = { path = "../stackable-shared", features = ["chrono", "time"] }
 stackable-telemetry = { path = "../stackable-telemetry", optional = true, features = ["clap"] }
 stackable-versioned = { path = "../stackable-versioned", optional = true }
-stackable-webhook = { path = "../stackable-webhook", optional = true }
+stackable-webhook = { path = "../stackable-webhook", optional = true}
```
sbernauer:
```suggestion
stackable-webhook = { path = "../stackable-webhook", optional = true }
```

crates/stackable-webhook/CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- BREAKING: The `ConversionWebhookServer` now returns a pair of values ([#1099]):
```
sbernauer:
```suggestion
- BREAKING: `ConversionWebhookServer::new` now returns a pair of values ([#1099]):
```

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -236,127 +179,71 @@ impl ConversionWebhookServer {
...
-                operator_namespace,
-                operator_service_name,
-                current_cert,
-            )
-            .await
-            .context(ReconcileCrdsSnafu)?;
+    /// Creates and returns a tuple consisting of a [`ConversionWebhookServer`], a [`CustomResourceDefinitionMaintainer`],
+    /// and a [`oneshot::Receiver`].
+    ///
+    /// See the referenced items for more details on usage.
```
sbernauer, line 207: I think this is actually the single most important code snippet, as this should be the thing all operators use.
It would be great if you could throw in the code from https://github.com/stackabletech/secret-operator/pull/634 in here

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -236,127 +179,71 @@ impl ConversionWebhookServer {
...
+    pub async fn with_maintainer<'a, H>(
+        // TODO (@Techassi): Use a trait type here which can be used to build all part of the
+        // conversion webhook server and a CRD maintainer.
+        crds_and_handlers: impl IntoIterator<Item = (CustomResourceDefinition, H)> + Clone,
+        operator_name: &'a str,
```
sbernauer: There is some inter-mangling between the service name and the field manager.
Separated them here:
```diff
diff --git a/crates/stackable-webhook/src/maintainer.rs b/crates/stackable-webhook/src/maintainer.rs
index 649ac7d..7234898 100644
--- a/crates/stackable-webhook/src/maintainer.rs
+++ b/crates/stackable-webhook/src/maintainer.rs
@@ -94,10 +94,11 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
     /// # async fn main() {
     /// # let (certificate_tx, certificate_rx) = channel(1);
     /// let options = CustomResourceDefinitionMaintainerOptions {
-    ///     operator_name: "my-service-name",
+    ///     operator_service_name: "my-service-name",
     ///     operator_namespace: "my-namespace",
     ///     webhook_https_port: 8443,
     ///     disabled: true,
+    ///     field_manager: "my-operator",
     /// };
     ///
     /// let client = Client::try_default().await.unwrap();
@@ -141,10 +142,11 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
     /// [`std::task::Poll::Ready`] and thus doesn't consume any resources.
     pub async fn run(mut self) -> Result<(), Error> {
         let CustomResourceDefinitionMaintainerOptions {
+            operator_service_name,
             operator_namespace,
             webhook_https_port,
-            operator_name,
             disabled,
+            field_manager,
         } = self.options;
 
         // If the maintainer is disabled or there are no custom resource definitions, immediately
@@ -196,7 +198,7 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
                         conversion_review_versions: vec!["v1".to_owned()],
                         client_config: Some(WebhookClientConfig {
                             service: Some(ServiceReference {
-                                name: operator_name.to_owned(),
+                                name: operator_service_name.to_owned(),
                                 namespace: operator_namespace.to_owned(),
                                 path: Some(format!("/convert/{crd_name}")),
                                 port: Some(webhook_https_port.into()),
@@ -211,7 +213,7 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
 
                 // Deploy the updated CRDs using a server-side apply.
                 let patch = Patch::Apply(&crd);
-                let patch_params = PatchParams::apply(operator_name);
+                let patch_params = PatchParams::apply(field_manager);
                 crd_api
                     .patch(&crd_name, &patch_params, &patch)
                     .await
@@ -235,8 +237,8 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
 // TODO (@Techassi): Make this a builder instead
 /// This contains required options to customize a [`CustomResourceDefinitionMaintainer`].
 pub struct CustomResourceDefinitionMaintainerOptions<'a> {
-    /// The service name used by the operator/conversion webhook and as a field manager.
-    pub operator_name: &'a str,
+    /// The name of the Kubernetes service which points to the operator/webhook.
+    pub operator_service_name: &'a str,
 
     /// The namespace the operator/conversion webhook runs in.
     pub operator_namespace: &'a str,
@@ -246,4 +248,7 @@ pub struct CustomResourceDefinitionMaintainerOptions<'a> {
 
     /// Indicates if the maintainer should be disabled.
     pub disabled: bool,
+
+    /// The fieldManager used when creating/patching CRDs.
+    pub field_manager: &'a str,
 }
diff --git a/crates/stackable-webhook/src/servers/conversion.rs b/crates/stackable-webhook/src/servers/conversion.rs
index 9b4beab..a71e2a7 100644
--- a/crates/stackable-webhook/src/servers/conversion.rs
+++ b/crates/stackable-webhook/src/servers/conversion.rs
@@ -194,9 +194,10 @@ impl ConversionWebhookServer {
         // TODO (@Techassi): Use a trait type here which can be used to build all part of the
         // conversion webhook server and a CRD maintainer.
         crds_and_handlers: impl IntoIterator<Item = (CustomResourceDefinition, H)> + Clone,
-        operator_name: &'a str,
+        operator_service_name: &'a str,
         operator_namespace: &'a str,
         disable_maintainer: bool,
+        field_manager: &'a str,
         client: Client,
     ) -> Result<
         (
@@ -213,9 +214,9 @@ impl ConversionWebhookServer {
 
         // TODO (@Techassi): These should be moved into a builder
         let webhook_options = ConversionWebhookOptions {
-            namespace: operator_namespace,
-            service_name: operator_name,
             socket_addr,
+            namespace: operator_namespace,
+            service_name: operator_service_name,
         };
 
         let (conversion_webhook_server, certificate_rx) =
@@ -225,10 +226,11 @@ impl ConversionWebhookServer {
 
         // TODO (@Techassi): These should be moved into a builder
         let maintainer_options = CustomResourceDefinitionMaintainerOptions {
+            operator_service_name,
+            operator_namespace,
             webhook_https_port: socket_addr.port(),
             disabled: disable_maintainer,
-            operator_namespace,
-            operator_name,
+            field_manager,
         };
 
         let (maintainer, initial_reconcile_rx) = CustomResourceDefinitionMaintainer::new(
```

Techassi: Do we actually need them to be separate? I purposely combined them to make the API easier to use. We currently use the same value for both these parameters anyway.

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Alright, I can add an example.

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: I think it's risky to enforce that the Kubernetes service name *needs* to be the same as the field manager.

Example: The k8s service name is passed via CLI arg, the field manager is hardcoded in Rust code.
If a user sets --k8s-service-name=foo, shit hits the fan as the field manager changes and CRD patches will fail because of conflicts between field managers.

sbernauer: (random relevant link for the record: https://github.com/stackabletech/operator-templating/blob/7ad8caf83241b221389c55747b624765a6786670/template/deploy/helm/%5B%5Boperator%5D%5D/templates/service.yaml#L9)

Techassi: > the field manager is hardcoded

I think that's something we should not do. Instead, we want to use the same name across all code (which comes from the CLI). But that's something for the future.

For now, I made the field manager a separate field again (see 16ed4109).

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Added in 0508e1b7.

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: Thanks for adding the field manager!
I think hard-coding the manager is fine for now, I don't know if other operators support configuring this.
But we should also change `operator_name` to `operator_service_name` as my patch did. Reason being is that the thing we add to the CRD is the name of the service pointing to the webhook (which is in our case the operator, but does not always needs to be).

crates/stackable-webhook/src/servers/conversion.rs
sbernauer: LGTM, thx!

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -236,127 +179,143 @@ impl ConversionWebhookServer {
...
+    /// Creates and returns a tuple consisting of a [`ConversionWebhookServer`], a [`CustomResourceDefinitionMaintainer`],
+    /// and a [`oneshot::Receiver`].
+    ///
+    /// ## Parameters
+    ///
+    /// - `crds_and_handlers`: An iterator over a 2-tuple (pair) mapping a [`CustomResourceDefinition`]
+    ///   to a handler function. In most cases, the generated `CustomResource::try_merge` function
+    ///   should be used. It provides the expected `fn(ConversionReview) -> ConversionReview`
+    ///   signature.
+    /// - `operator_name`: The name of the operator. This is used to construct the webhook service
```
sbernauer:
```suggestion
    /// - `operator_service_name`: The name of the Kubernetes service which points to the
    /// operator/webhook. This is put into the maintained CRD.
```

crates/stackable-webhook/src/servers/conversion.rs
```diff
@@ -236,127 +179,143 @@ impl ConversionWebhookServer {
...
+    ///   name.
+    /// - `operator_namespace`: The namespace the operator runs in. This is used to construct the
+    ///   webhook service name.
```
sbernauer:
```suggestion
    /// - `operator_namespace`: The namespace the operator runs in. This is put into the
    /// maintained CRD.
```

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Done in f9e16d9e.

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Done in f9e16d9e.

crates/stackable-webhook/src/servers/conversion.rs
Techassi: Done in f9e16d9e.

PR #1100, opened 2025-09-25
feat(operator/eos): Use 'now' as built time if debug build

Techassi:
Part of https://github.com/stackabletech/issues/issues/733.

During development, it makes sense to always use 'now' (determined at runtime instead of build time) as the built date. This also works nicely with the Tilt workflow (which uses Nix and defaults to `1980-01-01` as the built date).


maltesander (approved):
LGTM!

PR #1103, opened 2025-10-02
chore(operator/eos): Adjust the warning message

Techassi:
This PR adjusts the end-of-support warning message as discussed in today's meeting.

### DIFF
--- a/crates/stackable-operator/Cargo.toml
+++ b/crates/stackable-operator/Cargo.toml
@@ -21,7 +21,7 @@ webhook = ["dep:stackable-webhook"]
 [dependencies]
 stackable-certs = { path = "../stackable-certs", optional = true }
 stackable-operator-derive = { path = "../stackable-operator-derive" }
-stackable-shared = { path = "../stackable-shared", features = ["chrono", "time"]}
+stackable-shared = { path = "../stackable-shared", features = ["chrono", "time"] }
 stackable-telemetry = { path = "../stackable-telemetry", optional = true, features = ["clap"] }
 stackable-versioned = { path = "../stackable-versioned", optional = true }
 stackable-webhook = { path = "../stackable-webhook", optional = true }


crates/stackable-operator/src/eos/mod.rs
```diff
@@ -125,37 +127,37 @@ impl EndOfSupportChecker {
...
+        let build_age = Duration::try_from(now - self.built_datetime)
+            .expect("time delta of now and built datetime must not be less than 0")
+            .to_string();
+
         tracing::warn!(
-            eos.date = self.datetime.to_rfc3339(),
-            "the operator reached end-of-support"
+            eos.built.datetime = built_datetime,
+            eos.build.age = build_age,
+            "This operator version was built on {built_datetime} ({build_age} days ago) and may has reached end-of-support. \
```
lfrancke:
```suggestion
            "This operator version was built on {built_datetime} ({build_age} days ago) and may have reached end-of-support. \
```

lfrancke (changes requested):

crates/stackable-operator/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- Update the end-of-support warning message ([#1103])
```
sbernauer, line 19: I personally wouldn't add a new changelog entry, but just add this PR number to `Add end-of-support checker`.
EOS was never released, this change might confuse users, as they will not see the change.
Not blocking this PR though

crates/stackable-operator/src/eos/mod.rs
```diff
@@ -125,37 +127,37 @@ impl EndOfSupportChecker {
...
+        let built_datetime = self.built_datetime.to_rfc3339();
+        let build_age = Duration::try_from(now - self.built_datetime)
+            .expect("time delta of now and built datetime must not be less than 0")
+            .to_string();
+
         tracing::warn!(
-            eos.date = self.datetime.to_rfc3339(),
-            "the operator reached end-of-support"
+            eos.built.datetime = built_datetime,
+            eos.build.age = build_age,
```
sbernauer, line 157: Is built vs build intentional?

Techassi: It is!

- The `eos.built.datetime` refers to when this operator was _buitt_. It additionally mirrors the name of `built`'s `BUILT_TIME_UTC` constant which is the source of the value in most cases.
- The `eos.build.age` refers to the age of the build (this binary). It is the noun (contrary to above where the past-tense verb is used).

crates/stackable-operator/CHANGELOG.md
Techassi: I think it is fine as is.

The changelog should record _any_ change. And this change was done in a separate PR: two PRs - two entries. Combining multiple changes into an editorialized summary is something done as part of release notes, which is not something we do for any of these crates.

sbernauer: Well I guess we agree to disagree. I like that the changelog is not as "dumb" as simply running `git log`. For me these are the notes that document changes for library users :)
Bot blocking though

crates/stackable-operator/src/eos/mod.rs
```diff
@@ -125,37 +127,37 @@ impl EndOfSupportChecker {
...
+            "This operator version was built on {built_datetime} ({build_age} ago) and may has reached end-of-support. \
```
sbernauer: Did you accidentally revert https://github.com/stackabletech/operator-rs/pull/1103#discussion_r2398980404?
```suggestion
            "This operator version was built on {built_datetime} ({build_age} ago) and may have reached end-of-support. \
```

Techassi: Ah damn, seems like I did. My bad!

crates/stackable-operator/Cargo.toml
```diff
@@ -21,7 +21,7 @@ webhook = ["dep:stackable-webhook"]
 [dependencies]
 stackable-certs = { path = "../stackable-certs", optional = true }
 stackable-operator-derive = { path = "../stackable-operator-derive" }
-stackable-shared = { path = "../stackable-shared" }
+stackable-shared = { path = "../stackable-shared", features = ["chrono", "time"]}
```
sbernauer:
```suggestion
stackable-shared = { path = "../stackable-shared", features = ["chrono", "time"] }
```

sbernauer (dismissed):

lfrancke (approved):

sbernauer (approved):

PR #1105, opened 2025-10-07
feat!: Add new `ListenerClass.pinnedNodePorts` field

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/770
Implementation is in https://github.com/stackabletech/listener-operator/pull/340

Extend ListenerClass CRD with an extra field that allows configuring whether pinning should or should not be used for NodePorts, as this causes problems on (cloud) environments with rotating nodes.

### Author

- [x] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/crds/ListenerClass.yaml
+++ b/crates/stackable-operator/crds/ListenerClass.yaml
@@ -44,7 +44,7 @@ spec:
               pinnedNodePorts:
                 default: false
                 description: |-
-                  Wether or not a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
+                  Whether or not a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
 
                   By pinning the Pod to a specific (stable) Kubernetes node, stable addresses can be
                   provided using NodePorts. The pinning is achieved by listener-operator setting the


NickLarsenNZ:
What does it result in? Some changes to generated Services or just the generated Listener?

sbernauer:
> What (k8s resources changes) does it result in?

Actually the `csi::v1::Volume`, which in turn influences the created PVC (I guess). Feel free to have a look at `rust/operator-binary/src/csi_server/controller.rs` in https://github.com/stackabletech/listener-operator/pull/340

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -66,5 +66,19 @@ pub mod versioned {
         /// Defaults to `HostnameConservative`.
         #[serde(default = "ListenerClassSpec::default_preferred_address_type")]
         pub preferred_address_type: core_v1alpha1::PreferredAddressType,
+
+        /// Wether a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
+        ///
+        /// By pinning the Pod to a specific (stable) Kubernetes node, stable addresses can be
+        /// provided using NodePorts. The stickiness is achieved by listener-operator by setting the
```
Techassi:
```suggestion
        /// provided using NodePorts. The stickiness is achieved by listener-operator setting the
```

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -66,5 +66,19 @@ pub mod versioned {
...
+        /// `volume.kubernetes.io/selected-node` annotation on the Listener PVC.
+        ///
+        /// However, this only works on setups with long-living nodes. If your nodes are rotated on
+        /// a regular basis, the Pods previously running on a removed node will be stuck in Pending
+        /// until you delete the PVC with the stickiness.
```
Techassi: question: Could we somehow detect this and automatically remove the stickiness? Or is that something we don't want?

sbernauer: Ideally we have auto-detection in listener-op, which detects what ListenerClasses it should create, based on detecting GKE/AKS/IONOS etc. You might recall this from the CRD maintainer discussions.
But that's a follow-up

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -66,5 +66,19 @@ pub mod versioned {
         /// Defaults to `HostnameConservative`.
         #[serde(default = "ListenerClassSpec::default_preferred_address_type")]
         pub preferred_address_type: core_v1alpha1::PreferredAddressType,
+
+        /// Wether a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
```
sbernauer: From https://github.com/stackabletech/listener-operator/pull/340#discussion_r2425464803
```suggestion
        /// Whether or not a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
```

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -66,5 +66,19 @@ pub mod versioned {
         /// Defaults to `HostnameConservative`.
         #[serde(default = "ListenerClassSpec::default_preferred_address_type")]
         pub preferred_address_type: core_v1alpha1::PreferredAddressType,
+
+        /// Wether or not a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
```
NickLarsenNZ:
```suggestion
        /// Whether or not a Pod exposed using a NodePort should be pinned to a specific Kubernetes node.
```

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -66,5 +66,19 @@ pub mod versioned {
...
+        ///
+        /// By pinning the Pod to a specific (stable) Kubernetes node, stable addresses can be
+        /// provided using NodePorts. The pinning is achieved by listener-operator setting the
+        /// `volume.kubernetes.io/selected-node` annotation on the Listener PVC.
+        ///
+        /// However, this only works on setups with long-living nodes. If your nodes are rotated on
+        /// a regular basis, the Pods previously running on a removed node will be stuck in Pending
+        /// until you delete the PVC with the pinning.
+        ///
+        /// Because of this we don't enable pinning by default to support all environments.
```
NickLarsenNZ, line 80: I commented elsewhere, but i think this bit should be clarified a bit.

It's isn't about how long the nodes live, but it's about the uniqueness of nodes. I can't recall the qualifying factor (eg: is IP enough, or is it node hostname, or labels?)

sbernauer: I don't now to be honest. I'm pretty sure it's *not* IP, but likely just the node name.
Details are in https://github.com/stackabletech/listener-operator/pull/340

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -66,5 +66,19 @@ pub mod versioned {
...
+        #[serde(default)]
```
nightkr, line 81: This is a dangerous default. Having it be configurable is one thing (I'd probably still be against it, but my vague opinion doesn't really count for much anymore :p), but unpinned + NodePort should be a big warning sign that you're doing something Weird and are _probably_ using the wrong tool (and about to walk into a rake).

lfrancke: I'm the person who looked at customer issues we had and said "let's change it". To me it sounds totally sane - otherwise I probably wouldn't have  asked for this. Can you elaborate what I'm missing?

We've outlined the reasons a bit here: https://github.com/stackabletech/issues/issues/770

And let me close by saying that your opinion _does_ and _will_ count going forward. Thanks for still looking into this. We do have jobs available if you'd like to be paid to work on this ;-)

PR #1107, opened 2025-10-09
fix!: Default ListenerClass externalTrafficPolicy to null

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/770
Used by https://github.com/stackabletech/listener-operator/pull/347

As we start actually using LoadBalancers, we need to make sure they work everywhere.
@Maleware actually run into this in https://github.com/stackabletech/hbase-operator/issues/698

As we are running into the kube schema bug with this, we switch to our patched version.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/crd/listener/class/mod.rs
+++ b/crates/stackable-operator/src/crd/listener/class/mod.rs
@@ -64,9 +64,11 @@ pub mod versioned {
         /// to these nodes accordingly. Some cloud providers or bare metal installations do not implement
         /// some of the required features.
         //
-        // Pls note that we shouldn't mandate the default, but just let Kubernetes choose what to do
-        // (currently this means defaulting to Cluster), as this sound the most future-proof to me.
-        // Maybe in the future k8s defaults to Local if the LoadBalancer supports it
+        // Please note that Option is used here instead of a different default traffic policy. This will be
+        // deserialized as `None` and will thus forward the selection of the traffic policy to Kubernetes
+        // (which currently defaults to `Cluster`). This should be the most sensible option in most cases.
+        // There is the possibility Kubernetes will automatically choose `Local` if support for it on the
+        // LoadBalancer has been detected.
         pub service_external_traffic_policy: Option<core_v1alpha1::KubernetesTrafficPolicy>,
 
         /// Whether addresses should prefer using the IP address (`IP`) or the hostname (`Hostname`).


crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -50,13 +50,21 @@ pub mod versioned {
...
+        ///
+        /// * `Cluster`: Kubernetes defaults to `Cluster`, which means that traffic is routed to any
+        ///    node in the Kubernetes cluster that has a pod running the service.
+        /// * `Local`: Means that traffic is only routed to pods running on the same node as the
+        ///    Service.
+        ///
+        /// `Local` has a better performance as it avoids a network hop, but requires a "clever"
+        /// LoadBalancer, that respects what Pods run on which nodes and routes traffic only to that
+        /// nodes accordingly. Some cloud providers (such as IONOS) or bare metal installations
+        /// don't have such features, so the default is `Cluster` to work everywhere.
```
Techassi: suggestion: Small reword.

```suggestion
        ///
        /// * `Cluster`: Kubernetes default. Traffic is routed to any node in the Kubernetes cluster
        ///   that has a pod running the service.
        /// * `Local`: Traffic is only routed to pods running on the same node as the Service.
        ///
        /// The `Local` mode has better performance as it avoids a network hop, but requires a more
        /// sophisticated LoadBalancer, that respects what Pods run on which nodes and routes traffic
        /// only to these nodes accordingly. Some cloud providers or bare metal installations do not
        /// implement some of the required features.
```

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -50,13 +50,21 @@ pub mod versioned {
...
+        pub service_external_traffic_policy: Option<core_v1alpha1::KubernetesTrafficPolicy>,
```
Techassi, line 72: question: Why do we make this optional instead of changing the default?

crates/stackable-operator/src/crd/listener/core/mod.rs
```diff
@@ -45,6 +45,8 @@ pub mod versioned {
 
         /// Preserves the client source IP and avoid a second hop for LoadBalancer and NodePort type
         /// Services, but makes clients responsible for spreading the load.
+        ///
+        /// Does not work everywhere, e.g. not on IONOS!
```
Techassi: suggestion: Don't point fingers.

```suggestion
        ///
        /// Does not work everywhere out-of-the-box. Some cloud vendors do not implement the
        /// required features and bare metal installations need proper configuration.
```

crates/stackable-operator/src/crd/listener/class/mod.rs
sbernauer: Previously we choose a default explicitly (Local).
I looked at a few helm-charts and examples and most people don't set this field by default.
I also think we shouldn't mandate the default, but just let Kubernetes choose what to do (currently this means defaulting to Cluster), as this sound the most future-proof to me. Maybe in the future k8s defaults to Local if the LoadBalancer supports it :shrug:

crates/stackable-operator/src/crd/listener/core/mod.rs
sbernauer: There might be bare metal installations that can do this ;) And what would kind be? neither cloud nor bare-metal.

What about this?
```suggestion
        ///
        /// Does not work on all Kubernetes installations.
```

Techassi: I'm happy with your suggestion.

crates/stackable-operator/src/crd/listener/class/mod.rs
Techassi: Alright, thanks for the explanation. Do we want to record some of that decision making as a comment (dev or doc)?

sbernauer: Added a dev comment in fbad861, hope that works

crates/stackable-operator/crds/ListenerClass.yaml
```diff
@@ -77,16 +77,17 @@ spec:
...
-                  `externalTrafficPolicy` that should be set on the created [`Service`] objects.
+                  Service Internal Traffic Policy enables internal traffic restrictions to only route internal
+                  traffic to endpoints within the node the traffic originated from. The "internal" traffic
+                  here refers to traffic originated from Pods in the current cluster. This can help to reduce
+                  costs and improve performance. See [Kubernetes docs][k8s-docs].
 
-                  The default is `Local` (in contrast to `Cluster`), as we aim to direct traffic to a node running the workload
-                  and we should keep testing that as the primary configuration. Cluster is a fallback option for providers that
-                  break Local mode (IONOS so far).
+                  [k8s-docs]: https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/
```
Techassi, line 86: note: This description seems to be completely out-of-date when compared to the doc comment below.

crates/stackable-operator/src/crd/listener/class/mod.rs
```diff
@@ -50,13 +50,24 @@ pub mod versioned {
...
+        /// * `Local`: Traffic is only routed to pods running on the same node as the Service.
+        ///
+        /// The `Local` mode has better performance as it avoids a network hop, but requires a more
+        /// sophisticated LoadBalancer, that respects what Pods run on which nodes and routes traffic only
+        /// to these nodes accordingly. Some cloud providers or bare metal installations do not implement
+        /// some of the required features.
+        //
+        // Pls note that we shouldn't mandate the default, but just let Kubernetes choose what to do
+        // (currently this means defaulting to Cluster), as this sound the most future-proof to me.
+        // Maybe in the future k8s defaults to Local if the LoadBalancer supports it
```
Techassi: suggestion: Slight reword.

```suggestion
        //
        // Please note that Option is used here instead of a different default traffic policy. This will be
        // deserialized as `None` and will thus forward the selection of the traffic policy to Kubernetes
        // (which currently defaults to `Cluster`). This should be the most sensible option in most cases.
        // There is the possibility Kubernetes will automatically choose `Local` if support for it on the
        // LoadBalancer has been detected.
```

crates/stackable-operator/src/crd/listener/class/mod.rs
Techassi: Follow-up [comment](https://github.com/stackabletech/operator-rs/pull/1107#discussion_r2472209509).

crates/stackable-operator/crds/ListenerClass.yaml
sbernauer: Good catch!
Well, that's sadly another kube schema bug we didn't fix yet. It picks the description from
`pub enum KubernetesTrafficPolicy`
instead of 
`pub service_external_traffic_policy: Option<core_v1alpha1::KubernetesTrafficPolicy>`
(which should be picked as it is more concrete)
CC @NickLarsenNZ

As soon as this bug is fixed the description on the field (and not the enum) should show up

Techassi: I guess ideally we might want to try to combine them. But that is more than a "fix".

sbernauer: That's also an interesting idea :)

sbernauer: But I personally think we shouldn't block this PR on that, especially as we want it in 25.11

Techassi: Yeah I'm not blocking this PR if we feel like this fix should be included in 25.11.0. I'll let you decide that.

sbernauer: Yes, it definitely should be, thanks!

Techassi (approved):

lfrancke:
Is this something we should mention separately in the release notes?

PR #1116, opened 2025-10-31
fix(webhook): Force apply the CRD in the CRD maintainer

Techassi:
This ensures the operator will be the sole manager of the CRD (and all its fields) going forward. This avoids running into conflicts when CRDs were previously deployed by helm or stackablectl.

### DIFF
--- a/crates/stackable-webhook/src/maintainer.rs
+++ b/crates/stackable-webhook/src/maintainer.rs
@@ -234,7 +234,6 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
                 // See https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts
                 let patch_params = PatchParams::apply(field_manager).force();
 
-                // Finally apply the patch
                 crd_api
                     .patch(&crd_name, &patch_params, &patch)
                     .await


crates/stackable-webhook/src/maintainer.rs
```diff
@@ -213,7 +213,28 @@ impl<'a> CustomResourceDefinitionMaintainer<'a> {
...
+                // Overwrite value, become sole manager: If overwriting the value was intentional
+                // (or if the applier is an automated process like a controller) the applier should
+                // set the force query parameter to true [...], and make the request again. This
+                // forces the operation to succeed, changes the value of the field, and removes the
+                // field from all other managers' entries in managedFields.
+                //
+                // See https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts
+                let patch_params = PatchParams::apply(field_manager).force();
+
+                // Finally apply the patch
```
sbernauer: Not sure how I feel about this comment :see_no_evil: 
```suggestion
```

sbernauer (dismissed):
Tested, works :partying_face:

crates/stackable-webhook/src/maintainer.rs
Techassi: Sure, I can remove it.

PR #1118, opened 2025-11-11
feat: Support `objectOverrides`

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/712
Decision: https://github.com/stackabletech/decisions/issues/64
Needs https://github.com/stackabletech/documentation/pull/807 for the documentation part

The actual code change is oversee-able, most of the added lines are tests.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/crds/DummyCluster.yaml
+++ b/crates/stackable-operator/crds/DummyCluster.yaml
@@ -637,7 +637,7 @@ spec:
               objectOverrides:
                 default: []
                 description: |-
-                  A list of generic Kubernetes objects, which are merged on the objects that the operator
+                  A list of generic Kubernetes objects, which are merged into the objects that the operator
                   creates.
 
                   List entries are arbitrary YAML objects, which need to be valid Kubernetes objects.
--- a/crates/stackable-operator/src/deep_merger/crd.rs
+++ b/crates/stackable-operator/src/deep_merger/crd.rs
@@ -8,7 +8,7 @@ use crate::utils::crds::raw_object_list_schema;
 
 #[derive(Clone, Debug, Deserialize, Default, JsonSchema, Serialize, PartialEq)]
 pub struct ObjectOverrides(
-    /// A list of generic Kubernetes objects, which are merged on the objects that the operator
+    /// A list of generic Kubernetes objects, which are merged into the objects that the operator
     /// creates.
     ///
     /// List entries are arbitrary YAML objects, which need to be valid Kubernetes objects.


crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Support `objectOverrides` ([#1118]).
```
Techassi: note: I think we should add a small explanation how this works/what this does.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed
+
+- BREAKING: `ClusterResources` now requires the objects added to implement `DeepMerge`.
+  This is very likely a stackable-operator internal change, but technically breaking ([#1118]).
+
+### Removed
+
+- BREAKING: `ClusterResources` no longer derives `Eq` and `PartialEq` ([#1118]).
```
Techassi: question: Whats the reason those two trait implementations are removed?

crates/stackable-operator/src/cluster_resources.rs
```diff
@@ -563,7 +578,10 @@ impl ClusterResources {
                 .unwrap_or_else(|err| warn!("{}", err));
         }
 
-        let mutated = resource.maybe_mutate(&self.apply_strategy);
+        let mut mutated = resource.maybe_mutate(&self.apply_strategy);
+
+        // We apply the object overrides of the user at the very last to offer maximum flexibility.
```
Techassi:
```suggestion
        // We apply the object overrides of the user at the very end to offer maximum flexibility.
```

crates/stackable-operator/src/crd/listener/listeners/v1alpha1_impl.rs
```diff
@@ -1,7 +1,148 @@
...
+            },
+        );
+        merge_strategies::list::map(
+            &mut self.ports,
+            other.ports,
+            &[|lhs, rhs| lhs.name == rhs.name],
+            |current_item, other_item| {
+                DeepMerge::merge_from(current_item, other_item);
+            },
+        );
```
Techassi, line 39: note: Please add two dev comments to quickly explain what this code does and why we need this specialized merge.

crates/stackable-operator/src/crd/listener/listeners/v1alpha1_impl.rs
```diff
@@ -1,7 +1,148 @@
...
+        merge_strategies::map::granular(
+            &mut self.node_ports,
+            other.node_ports,
+            |current_item, other_item| {
+                DeepMerge::merge_from(current_item, other_item);
+            },
+        );
```
Techassi, line 65: note: Same as above.

crates/stackable-operator/src/crd/listener/listeners/v1alpha1_impl.rs
```diff
@@ -1,7 +1,148 @@
...
+    fn merge_from(&mut self, other: Self) {
+        DeepMerge::merge_from(&mut self.address, other.address);
+        self.address_type = other.address_type;
+        merge_strategies::map::granular(
+            &mut self.ports,
+            other.ports,
+            |current_item, other_item| {
+                DeepMerge::merge_from(current_item, other_item);
+            },
+        );
```
Techassi, line 79: note: Again, same as above.

crates/stackable-operator/src/crd/listener/listeners/v1alpha1_impl.rs
```diff
@@ -1,7 +1,148 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deep_merge_listener() {
+        let mut base: ListenerSpec = serde_yaml::from_str(
```
Techassi: note: Can we please move those YAML contents into actual files (for example located at `stackable-operator/fixtures/object-overrides` or similar) instead of inline strings.

crates/stackable-operator/src/lib.rs
```diff
@@ -22,6 +22,7 @@ pub mod kvp;
 pub mod logging;
 pub mod memory;
 pub mod namespace;
+pub mod patchinator;
```
Techassi: question: Maybe `object_overrides` would be a better name?

crates/stackable-operator/src/deep_merger/crd.rs
```diff
@@ -0,0 +1,20 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::utils::crds::raw_object_list_schema;
+
+#[derive(Clone, Debug, Deserialize, JsonSchema, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ObjectOverrides {
+    /// A list of generic Kubernetes objects, which are merged onto the objects that the operator
+    /// creates.
```
Techassi:
```suggestion
    /// A list of generic Kubernetes objects, which are merged with objects created by the operator.
```

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+        "failed to parse dynamic object as apiVersion {target_api_version:?} and kind {target_kind:?}"
+    ))]
+    ParseDynamicObject {
+        source: kube::core::dynamic::ParseDynamicObjectError,
+        target_api_version: String,
+        target_kind: String,
+    },
+}
+
+pub fn apply_patches<R>(base: &mut R, patches: &ObjectOverrides) -> Result<(), Error>
```
Techassi: note: Please add a few doc comments explaining what this function does.

suggestion: Rename this function to `apply_overrides` and change the parameter name to `overrides`.

```suggestion
pub fn apply_overrides<R>(base: &mut R, overrides: &ObjectOverrides) -> Result<(), Error>
```

crates/stackable-operator/src/deep_merger/crd.rs
```diff
@@ -0,0 +1,20 @@
...
+    ///
+    /// List entries are arbitrary YAML objects, which need to be valid Kubernetes objects.
+    ///
+    /// Read the [Object overrides documentation](DOCS_BASE_URL_PLACEHOLDER/concepts/overrides#object-overrides)
+    /// for more information.
+    #[serde(default)]
+    #[schemars(schema_with = "raw_object_list_schema")]
+    pub object_overrides: Vec<DynamicObject>,
```
Techassi: question: Seeing that this struct only contains a single field, this is make sense to turn this into a tuple struct instead? This could also enable us to implement `Deref(Mut)` on it for ergonomics.

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+where
+    R: kube::Resource<DynamicType = ()> + DeepMerge + DeserializeOwned,
+{
+    for patch in &patches.object_overrides {
+        apply_patch(base, patch)?;
+    }
+    Ok(())
+}
+
+pub fn apply_patch<R>(base: &mut R, patch: &DynamicObject) -> Result<(), Error>
```
Techassi: question: Does this need to be pub? I don't think so. As such, we could make this private instead.

suggestion: Rename in a similar fashion like above.

```suggestion
pub fn apply_override<R>(base: &mut R, override: &DynamicObject) -> Result<(), Error>
```

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+where
+    R: kube::Resource<DynamicType = ()> + DeepMerge + DeserializeOwned,
+{
+    use kube::ResourceExt;
```
Techassi: note: Do we use this style of trait imports anywhere else in this crate/repo (other than generated code)? I don't think so. As such we should keep it consistent and move it to the top of this file.

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    let Some(patch_type) = &patch.types else {
+        return Ok(());
+    };
+    if patch_type.api_version != R::api_version(&()) || patch_type.kind != R::kind(&()) {
```
Techassi: note: These `&()` are honestly kinda weird, but I guess there is nothing we can do to get rid of them.

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    Ok(())
+}
+
+pub fn apply_patch<R>(base: &mut R, patch: &DynamicObject) -> Result<(), Error>
+where
+    R: kube::Resource<DynamicType = ()> + DeepMerge + DeserializeOwned,
+{
+    use kube::ResourceExt;
+
+    let Some(patch_type) = &patch.types else {
```
Techassi:
```suggestion
    let Some(object_type) = &override.types else {
```

crates/stackable-operator/src/patchinator/mod.rs
Techassi:
```suggestion
    if object_type.api_version != R::api_version(&()) || object_type.kind != R::kind(&()) {
```

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+        return Ok(());
+    };
+    if patch_type.api_version != R::api_version(&()) || patch_type.kind != R::kind(&()) {
+        return Ok(());
+    }
+    let Some(patch_name) = &patch.metadata.name else {
```
Techassi:
```suggestion
    let Some(object_name) = &override.metadata.name else {
```

crates/stackable-operator/src/patchinator/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    // If there is a namespace on the base object, it needs to match as well
+    // Note that it is not set for cluster-scoped objects.
+    if base.namespace() != patch.metadata.namespace {
+        return Ok(());
+    }
+
+    let deserialized_patch =
+        patch
+            .clone()
```
Techassi: note: If we need to clone every single override anyway, we should instead take an owned value (in both functions). Taking a reference is signaling that no owned data is needed, which is not true. We should make this clear when these functions are called.

crates/stackable-operator/src/deep_merger/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+        },
+        apimachinery::pkg::util::intstr::IntOrString,
+    };
+    use kube::api::ObjectMeta;
+
+    use super::*;
+
+    /// Using [`serde_yaml`] to generate the test data
+    fn generate_service_account() -> ServiceAccount {
+        serde_yaml::from_str(
```
Techassi: note: Again, I would like to see this (and all other inline YAML strings) getting moved into actual YAML files.

crates/stackable-operator/src/deep_merger/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    kind: TrinoCluster
+    name: trino
+    uid: c85bfb53-a28e-4782-baaf-3c218a25f192
+",
+        )
+        .unwrap()
+    }
+
+    /// Generate the test data programmatically (as operators would normally do)
+    fn generate_stateful_set() -> StatefulSet {
```
Techassi, line 114: question: But, do we need to? I feel like this can also be a static YAML file (to actually only test the merging/overrides).

Techassi (changes requested):
Mostly lgtm, a few various question, comments and suggestions.

crates/stackable-operator/CHANGELOG.md
sbernauer: It's because Vec<DynamicObject> is now stored in the ClusterResources, and DynamicObject does not implement Eq. I was at least able to derive PartialEq again: bf1d753

crates/stackable-operator/src/patchinator/mod.rs
sbernauer: Moved in cb0de44

crates/stackable-operator/src/patchinator/mod.rs
sbernauer: Yeah, that's just how kube works

crates/stackable-operator/src/patchinator/mod.rs
sbernauer: Changed to owned in 4873ac2

crates/stackable-operator/src/deep_merger/crd.rs
sbernauer: Yeah, we could have
```rust
#[derive(Clone, Debug, Deserialize, Default, JsonSchema, Serialize, PartialEq)]
pub struct ObjectOverrides2(
    /// A list of generic Kubernetes objects, which are merged onto the objects that the operator
    /// creates.
    ///
    /// List entries are arbitrary YAML objects, which need to be valid Kubernetes objects.
    ///
    /// Read the [Object overrides documentation](DOCS_BASE_URL_PLACEHOLDER/concepts/overrides#object-overrides)
    /// for more information.
    #[schemars(schema_with = "raw_object_list_schema")]
    Vec<DynamicObject>,
);
```

and than call them as
```rust
        #[serde(flatten)]
        object_overrides: ObjectOverrides,
        #[serde(default)]
        object_overrides_2: ObjectOverrides2,
```
I think we should optimize for ease of use, not sure which one that is.
The ObjectOverrides2 have the downside that users need to remember to #[serde(default)] it and that we can't add new fields. But it has the upside that one is not forced to name the field configIverrides.
Any opinions? Can we think of any field additions that might come?

crates/stackable-operator/CHANGELOG.md
sbernauer: Added a bit of content in 9f5019b

crates/stackable-operator/src/deep_merger/crd.rs
sbernauer: What about this?
```suggestion
    /// A list of generic Kubernetes objects, which are merged onto the objects created by the operator.
```
It makes it clear what is merged onto what

crates/stackable-operator/src/deep_merger/mod.rs
sbernauer: I personally prefer having them in-line as this way it's IMHO much easier to read and maintain compared to cluttered over 20 files.
I was thinking of using https://docs.rs/indoc/2.0.7/indoc/, but figured it's not worth the effort, but happy to do so

crates/stackable-operator/src/deep_merger/mod.rs
sbernauer: Nope, we don't. I only did this to increase the test diversity :sweat_smile:
My though process was maybe somethings with defaults with and without serde might be different and stuff.
It's not super important, but I also don't think this will hurt us in any way, so I would say better safe than sorry.

crates/stackable-operator/src/lib.rs
sbernauer: I really like the "cleverness" of the code to apply arbitrary patches to any k8s resource.
To me personally these are Kubernetes resources (but it's also fine to call them objects) and a list of patches.
I see that we need to call it objectOverrides in the context of SDP because of consistency, but would prefer to keep the code generic, so that we (or technically someone else) might be able to use it for other stuff in the future.

And yeah, I also like the name ^^

crates/stackable-operator/src/crd/listener/listeners/v1alpha1_impl.rs
sbernauer: I struggle to find anything that is not already on the trait docs in https://docs.rs/k8s-openapi/latest/k8s_openapi/trait.DeepMerge.html
What are you thinking of?

sbernauer: Was able to add two comments in e28d64e

crates/stackable-operator/src/patchinator/mod.rs
sbernauer: Added some docs in 0d6c134

crates/stackable-operator/src/patchinator/mod.rs
sbernauer: No, our current code does not require this function to be pub.
On the other hand, one (maybe someone else?) might want to do some resource patching without the need to construct an ObjectOverrides struct just for this purpose.

crates/stackable-operator/src/cluster_resources.rs
```diff
@@ -348,7 +354,10 @@ impl ClusterResource for Deployment {
 ///     plural = "AppClusters",
 ///     namespaced,
 /// )]
-/// struct AppClusterSpec {}
+/// struct AppClusterSpec {
+///     #[serde(flatten)]
```
sbernauer: need to be updated to the decision outcome of https://github.com/stackabletech/operator-rs/pull/1118#discussion_r2553980792

crates/stackable-operator/src/deep_merger/mod.rs
sbernauer: as discussed during on-site I switched to indoc

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Support `objectOverrides`, which are a list of generic Kubernetes objects, which are merged onto the objects that the operator creates.
+  Alongside, a `deep_merger` module was added, which takes a Kubernetes object and a list of depp merged and applies them to the object ([#1118]).
```
Techassi:
```suggestion
- Support `objectOverrides`, a list of generic Kubernetes objects, which are merged with objects created by the operator.
  Alongside, a `deep_merger` module was added, which takes a Kubernetes object and a list of overrides and merges them with the provided object ([#1118]).
```

crates/stackable-operator/src/deep_merger/crd.rs
Techassi: > The ObjectOverrides2 have the downside that users need to remember to `#[serde(default)]` it

The current solution needs `#[serde(flatten)]`. So users already need to remember to use a specific attribute. The doc comments on `ObjectOverrides2` can mention that `#[serde(default)]` has to be used.

> Can we think of any field additions that might come?

Honestly, no. This struct literally has a single use-case.

So all in all, changing this is the right move I think.

crates/stackable-operator/CHANGELOG.md
sbernauer: I think the merge order matters, WDYT of that?
```suggestion
- Support `objectOverrides`, a list of generic Kubernetes objects, which are merged into the objects created by the operator.
  Alongside, a `deep_merger` module was added, which takes a Kubernetes object and a list of overrides and merges them into the provided object ([#1118]).
```

crates/stackable-operator/src/patchinator/mod.rs
Techassi: Alright, let's keep it as is then.

crates/stackable-operator/src/deep_merger/mod.rs
```diff
@@ -0,0 +1,485 @@
...
+    object_overrides: ObjectOverrides,
+) -> Result<(), Error>
+where
+    R: kube::Resource<DynamicType = ()> + DeepMerge + DeserializeOwned,
+{
+    for object_override in object_overrides.object_overrides {
+        apply_deep_merge(base, object_override)?;
+    }
+    Ok(())
+}
```
Techassi: suggestion: We could make this an associated function on `ObjectOverrides`.

sbernauer: Done in fc38a7c

crates/stackable-operator/src/deep_merger/crd.rs
sbernauer: Done in fe29dcb

crates/stackable-operator/src/cluster_resources.rs
sbernauer: done

crates/stackable-operator/CHANGELOG.md
Techassi: The verb "merge" doesn't combine with "to". We should instead use "into" or "with".

crates/stackable-operator/crds/DummyCluster.yaml
```diff
@@ -634,6 +634,20 @@ spec:
...
+              objectOverrides:
+                default: []
+                description: |-
+                  A list of generic Kubernetes objects, which are merged on the objects that the operator
+                  creates.
+
+                  List entries are arbitrary YAML objects, which need to be valid Kubernetes objects.
+
+                  Read the [Object overrides documentation](https://docs.stackable.tech/home/nightly/concepts/overrides#object-overrides)
+                  for more information.
```
Techassi: note: Use "merged into" or "merged with".

crates/stackable-operator/src/deep_merger/crd.rs
```diff
@@ -0,0 +1,39 @@
...
+pub struct ObjectOverrides(
+    /// A list of generic Kubernetes objects, which are merged on the objects that the operator
+    /// creates.
+    ///
+    /// List entries are arbitrary YAML objects, which need to be valid Kubernetes objects.
+    ///
+    /// Read the [Object overrides documentation](DOCS_BASE_URL_PLACEHOLDER/concepts/overrides#object-overrides)
+    /// for more information.
+    //
+    // Remember to use `#[serde(default)]` when including this into a CRD!
```
Techassi, line 19: note: I would turn this into a doc comments. Otherwise, they are not visible at all on the caller side which will definitely lead to users forgetting to use it.

sbernauer: If I do that it will leak into our CRD...
Or I need to get hacky with #[docs("bluh")] and duplicate the docs...

crates/stackable-operator/crds/DummyCluster.yaml
sbernauer: 53d9af5

crates/stackable-operator/src/deep_merger/crd.rs
Techassi: Oh yeah, true. Welp, let's leave it as is.

PR #1119, opened 2025-11-18
feat(stackable-webhook)!: Add support for mutating webhooks

sbernauer:
# Description

Part of https://github.com/stackabletech/commons-operator/issues/111
Used by https://github.com/stackabletech/commons-operator/pull/387

BREAKING: Refactor the entire `WebhookServer` mechanism, so multiple webhooks can run in parallel.
  Put individual webhooks (currently `ConversionWebhook` and `MutatingWebhook`) behind the `Webhook` trait.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ k8s-openapi = { version = "0.26.0", default-features = false, features = ["schem
 # We use rustls instead of openssl for easier portability, e.g. so that we can build stackablectl without the need to vendor (build from source) openssl
 # We use ring instead of aws-lc-rs, as this currently fails to build in "make run-dev"
 # We pin the kube version, as we use a patch for 2.0.1
-kube = { git = "https://github.com/stackabletech/kube-rs", branch = "2.0.1-fix-schema-hoisting", version = "=2.0.1", default-features = false, features = ["client", "jsonpatch", "runtime", "derive","admission", "rustls-tls", "ring"] }
+kube = { git = "https://github.com/stackabletech/kube-rs", branch = "2.0.1-fix-schema-hoisting", version = "=2.0.1", default-features = false, features = ["client", "jsonpatch", "runtime", "derive", "admission", "rustls-tls", "ring"] }
 opentelemetry = "0.31.0"
 opentelemetry_sdk = { version = "0.31.0", features = ["rt-tokio"] }
 opentelemetry-appender-tracing = "0.31.0"


crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,87 @@
-//! Utility types and functions to easily create ready-to-use webhook servers
```
NickLarsenNZ, line 1: I think we can still have _some_ docs here. At least a title describing what the crate does (unless this doesn't render how I think in docs?)

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,87 @@
...
+
+    #[snafu(display("failed to update certificate"))]
+    UpdateCertificate { source: WebhookError },
+
+    #[snafu(display("failed to encode CA certificate as PEM format"))]
+    EncodeCertificateAuthorityAsPem { source: x509_cert::der::Error },
 }
 
-/// A ready-to-use webhook server.
+/// An HTTPS server that serves a bunch of webhooks.
```
NickLarsenNZ:
```suggestion
/// An HTTPS server that serves one or more webhooks.
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,87 @@
...
+///
+/// It handles TLS certificate rotation.
```
NickLarsenNZ:
```suggestion
/// It also handles TLS certificate rotation.
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,87 @@
...
+///     socket_addr: WebhookServer::DEFAULT_SOCKET_ADDRESS,
+///     operator_namespace: "my-namespace".to_owned(),
+///     operator_service_name: "my-operator".to_owned(),
+/// };
+/// let webhook_server = WebhookServer::new(webhook_options, webhooks).await.unwrap();
+/// # }
+/// ```
 pub struct WebhookServer {
+    options: WebhookServerOptions,
+    webhooks: Vec<Box<dyn Webhook>>,
```
NickLarsenNZ, line 80: I'm sure the use of `Box<dyn X>>` will come up at some point. As discussed, I'm happy with it since you have shown it to not be used in any hot paths and dyn gives some flexibility here.

I do think with a bunch of effort, we could use static dispatch, but this works and we need it now.

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,87 @@
...
     tls_server: TlsServer,
+    cert_rx: mpsc::Receiver<Certificate>,
+}
+
+#[derive(Clone, Debug)]
+pub struct WebhookServerOptions {
+    /// The default HTTPS socket address the [`TcpListener`][tokio::net::TcpListener]
+    /// binds to.
+    pub socket_addr: SocketAddr,
```
NickLarsenNZ: I believe this isn't the _default_, but rather the _chosen_?

```suggestion
    /// The HTTPS socket address the [`TcpListener`][tokio::net::TcpListener]
    /// binds to.
    pub socket_addr: SocketAddr,
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,87 @@
...
+
+    /// The namespace the operator/webhook is running in.
+    pub operator_namespace: String,
```
NickLarsenNZ: IMO, we should abstract away from _operator_ here (yeah, I know, operator-rs).

We don't need to do it now though, so you can read this and ignore it. Just sharing thoughts.

```suggestion
    /// The namespace the webhook is running in.
    pub webhook_namespace: String,
```

NickLarsenNZ: Same for the next field.

crates/stackable-webhook/src/lib.rs
```diff
@@ -161,17 +120,26 @@ impl WebhookServer {
...
-        let router = router
+        let mut router = Router::new()
             .layer(service_builder)
             // The health route is below the AxumTraceLayer so as not to be instrumented
             .route("/health", get(|| async { "ok" }));
 
+        for webhook in webhooks.iter() {
+            router = webhook.register_routes(router);
+        }
+
```
NickLarsenNZ: These will not be instrumented. They will need to move before the `layer(service_builder)`.

At the same time we should consider renaming `service_builder`.

Something like this (maybe I need to do `let router = router.blah`):

```rust
        let mut router = Router::new();

        for webhook in webhooks.iter() {
            router = webhook.register_routes(router);
        }

        // Enrich spans for routes above.
        // Routes defined below it will not be instrumented to reduce noise.
        router.layer(tracing_service_builder);
        router.route("/health", get(|| async { "ok" }));
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -161,17 +120,26 @@ impl WebhookServer {
...
         tracing::debug!("create TLS server");
-        let (tls_server, cert_rx) = TlsServer::new(router, options)
+        let (tls_server, cert_rx) = TlsServer::new(router, options.clone())
```
NickLarsenNZ: thought: We could probably make `TlsServer::new()` borrow the options, but I do wonder why we also return the options.

crates/stackable-webhook/src/lib.rs
```diff
@@ -200,19 +168,59 @@ impl WebhookServer {
...
+        tokio::pin!(future_server);
+        tokio::pin!(future_signal);
+
+        tokio::select! {
+            res = &mut future_server => {
+                // If the server future errors, propagate the error
+                res?;
+            }
+            _ = &mut future_signal => {
+                tracing::info!("shutdown signal received, stopping server");
```
NickLarsenNZ: I thought the future signal already emitted this. I recall seeing two messages somewhere, but maybe that was the kube crate.

crates/stackable-webhook/src/lib.rs
```diff
@@ -200,19 +168,59 @@ impl WebhookServer {
...
     async fn run_server(self) -> Result<()> {
         tracing::debug!("run webhook server");
 
-        self.tls_server.run().await.context(RunTlsServerSnafu)
+        let Self {
+            options,
+            mut webhooks,
+            tls_server,
+            mut cert_rx,
+            // initial_reconcile_tx,
```
NickLarsenNZ: I guess this can go?

crates/stackable-webhook/src/lib.rs
```diff
@@ -200,19 +168,59 @@ impl WebhookServer {
...
+                            .handle_certificate_rotation(&cert, &ca_bundle, &options)
+                            .await
+                            .context(UpdateCertificateSnafu)?;
+                    }
+                }
+            }
+
+            // We need to hint the return type to the compiler
+            #[allow(unreachable_code)]
+            Ok(())
```
NickLarsenNZ: We might be able to avoid this through the `!` (Never) type. But I guess that won't work with returning errors via `?`.

I am curious though.

NickLarsenNZ: I tried, and no. It isn't trivial.

crates/stackable-webhook/src/servers/conversion_webhook.rs
```diff
@@ -0,0 +1,185 @@
...
+        options: &WebhookServerOptions,
+    ) -> Result<(), WebhookError> {
+        if self.disable_crd_maintenance {
+            return Ok(());
+        }
+
+        tracing::info!(
+            k8s.crd.names = ?self.crds_and_handlers.iter().map(|(crd, _)| crd.name_any()).collect::<Vec<_>>(),
+            "reconciling custom resource definitions"
+        );
```
NickLarsenNZ: The same attribute can be sent with multiple values (though maybe that depends on the subscriber?).

So this could be inverted to iterating with forEach, and emitting the event.

Something like this:

```rust
self.crds_and_handlers.forEach(|(crd, _)| {
    tracing::info!(
        k8s.crd.name = ?crd.name_any(),
        "reconciling custom resource definition"
    );
});
```

But further below, we iterate, and can just emit the even there at the time that it happens?

NickLarsenNZ: > But further below, we iterate, and can just emit the even there at the time that it happens?

We already do, so this event can probably go.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,185 @@
...
+            };
+
+            let route = format!("/convert/{crd_name}");
+            router = router.route(&route, post(handler_fn));
+        }
+
+        router
+    }
+
+    async fn handle_certificate_rotation(
```
NickLarsenNZ, line 250: We should probably instrument this function (so that it has it's own defined span).

It might also be worth creating inner spans for each iteration of the CRDs.

crates/stackable-webhook/src/servers/conversion_webhook.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        let crd_api: Api<CustomResourceDefinition> = Api::all(self.client.clone());
+
+        for mut crd in self.crds_and_handlers.iter().map(|(crd, _)| crd).cloned() {
```
NickLarsenNZ: I _think_ we can move this block into the `map` body. Not sure yet.

If we can, then it makes creating a span and attaching it to each iteration much easier through the `instrument` method.

I can take a look into that (if I become the blocker, just resolve the conversation and I can look at it later).

crates/stackable-webhook/src/webhooks/mod.rs
NickLarsenNZ, line 1: We can probably rename this module to webhooks.

_Unless mod.rs has the server, in which case this can become `server` and the webhooks can move into `webhooks`_.

crates/stackable-webhook/src/servers/mutating_webhook.rs
```diff
@@ -0,0 +1,224 @@
...
+            handler_state,
+            _resource: PhantomData,
+            disable_mutating_webhook_configuration_maintenance,
+            client,
+            field_manager,
+        }
+    }
+
+    fn http_path(&self) -> String {
+        format!("/mutate/{}", self.mutating_webhook_configuration.name_any())
```
NickLarsenNZ:
```suggestion
        format!("/mutate/{mutating_webhook_configuration_name}", mutating_webhook_configuration_name = self.mutating_webhook_configuration.name_any())
```

or

```suggestion
        let mutating_webhook_configuration_name = self.mutating_webhook_configuration.name_any();
        format!("/mutate/{mutating_webhook_configuration_name}")
```

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,224 @@
...
+                        AdmissionResponse::invalid(format!("failed to convert to request: {err}"))
+                            .into_review(),
+                    );
+                }
+            };
+
+            let response = handler(handler_state, request).await;
+            let review = response.into_review();
+            Json(review)
+        };
```
NickLarsenNZ, line 190: This _could_ probably move into a private method (so it can still access self.handler).
But I'm ok leaving it here too.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        let mut mutating_webhook_configuration = self.mutating_webhook_configuration.clone();
+        let mwc_name = mutating_webhook_configuration.name_any();
+        tracing::info!(
+            k8s.mutatingwebhookconfiguration.name = mwc_name,
+            "reconciling mutating webhook configurations"
+        );
+
+        for webhook in mutating_webhook_configuration.webhooks.iter_mut().flatten() {
```
NickLarsenNZ, line 214: Can there be more than one _webhook_? Or are we iterating over none or one?

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -86,15 +86,23 @@ impl TlsServer {
...
+        options: WebhookServerOptions,
     ) -> Result<(Self, mpsc::Receiver<Certificate>)> {
         let (certificate_tx, certificate_rx) = mpsc::channel(1);
 
-        let WebhookOptions {
+        let WebhookServerOptions {
             socket_addr,
-            subject_alterative_dns_names,
+            operator_namespace,
+            operator_service_name,
```
NickLarsenNZ: I do think we should remove the `operator_` prefix. The operator will be choosing to launch it in its namespace and with its service, but this code should be regardless of whether an operator is in use.

crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -120,6 +128,10 @@ impl TlsServer {
         Ok((tls_server, certificate_rx))
     }
 
+    pub fn socket_addr(&self) -> &SocketAddr {
+        &self.socket_addr
+    }
```
NickLarsenNZ, line 133: We could probably make the field `pub` to remove the need for the method.

crates/stackable-webhook/src/lib.rs
sbernauer: Technically it can also be zero webhooks :P
But I'll pick it anyway ;)

crates/stackable-webhook/src/lib.rs
sbernauer: aggreed

crates/stackable-webhook/src/lib.rs
sbernauer: I also failed

crates/stackable-webhook/src/servers/conversion_webhook.rs
sbernauer: This was existing code copied across, we both agreed to leave it as is for now

crates/stackable-webhook/src/lib.rs
sbernauer: Renamed in ade4a1b

crates/stackable-webhook/src/lib.rs
sbernauer: Removed the clone in 7120310

crates/stackable-webhook/src/lib.rs
sbernauer: Sure thing, fixed in a070857

crates/stackable-webhook/src/tls/mod.rs
sbernauer: Since ade4a1b0e0761e6c69aa750687bf46ada27ddc6e they are now called
```rust
        let WebhookServerOptions {
            socket_addr,
            webhook_namespace,
            webhook_service_name,
        } = options;
```

crates/stackable-webhook/src/tls/mod.rs
sbernauer: I *think* the getter is best practice, as we can can change internals (e.g. struct names) without breaking consumers.
But no strong opinion.

crates/stackable-webhook/src/lib.rs
sbernauer: Moved in 6fc8a979629f18de6531cefb37a1b1b2823523a4

crates/stackable-webhook/src/servers/conversion_webhook.rs
sbernauer: See 2721d72

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
sbernauer: See 2721d72

crates/stackable-webhook/src/servers/conversion_webhook.rs
sbernauer: Related to 2721d72

crates/stackable-webhook/src/webhooks/mod.rs
sbernauer: Renamed in 3ff8b2a

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: There can be 0..n. But having no webhook doesn't really make sense.
There can be multiple webhooks in here, but they will all share the same path.
Added some docs in 92f14b3

crates/stackable-webhook/src/lib.rs
sbernauer: Added some docs in 6ca52f2

crates/stackable-webhook/src/servers/conversion_webhook.rs
NickLarsenNZ: Nice!

crates/stackable-webhook/src/tls/mod.rs
NickLarsenNZ: It's fine as is.

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,99 @@
...
-//! Most users will only use the top-level exported generic [`WebhookServer`]
-//! which enables complete control over the [Router] which handles registering
-//! routes and their handler functions.
+//! Currently the following webhooks are supported:
 //!
-//! ```
-//! use stackable_webhook::{WebhookServer, WebhookOptions};
-//! use axum::Router;
+//! * CRD conversion webhooks: [`ConversionWebhook`](`webhooks::ConversionWebhook`)
+//! * Mutating webhooks: [`MutatingWebhook`](`webhooks::MutatingWebhook`)
```
NickLarsenNZ: Do the links with with backticks (`) in the parentheses?

It can be simplified like:

```suggestion
//! * CRD conversion webhooks: [`webhooks::ConversionWebhook`]
//! * Mutating webhooks: [`webhooks::MutatingWebhook`]
```

or even

```suggestion
//! * [`webhooks::ConversionWebhook`]
//! * [`webhooks::MutatingWebhook`]
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -156,22 +124,34 @@ impl WebhookServer {
         // by the Axum project.
         //
         // See https://docs.rs/axum/latest/axum/middleware/index.html#applying-multiple-middleware
-        // TODO (@NickLarsenNZ): rename this server_builder and keep it specific to tracing, since it's placement in the chain is important
-        let service_builder = ServiceBuilder::new().layer(trace_layer);
+        // TODO (@NickLarsenNZ): keep this server_builder specific to tracing, since it's placement in the chain is important
```
NickLarsenNZ: I believe this is now done, so the comment can be dropped

```suggestion
```

crates/stackable-webhook/src/lib.rs
sbernauer:
```
[`ConversionWebhook`](`webhooks::ConversionWebhook`)
```
and
```
[`webhooks::ConversionWebhook`]
```
actually render differently

sbernauer: Should we have `ConversionWebhook` or `webhooks::ConversionWebhook` in the docs?
I think I personally prefer `ConversionWebhook`

crates/stackable-webhook/src/lib.rs
sbernauer: perfect

crates/stackable-webhook/src/lib.rs
NickLarsenNZ: I prefer the longer (to save having to add an extra reference) and treat it as part of its name. The backticks might be messing with it. I usually don't use backticks since it is emphasised with a link style.

NickLarsenNZ: ie:

```suggestion
//! * [webhooks::ConversionWebhook]
//! * [webhooks::MutatingWebhook]
```

sbernauer: I don't care to be honest, picking your suggestion

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: If you are ok with it I think I would leave it here :see_no_evil:

crates/stackable-webhook/src/lib.rs
sbernauer: Looking at https://docs.rs/tokio/latest/src/tokio/signal/mod.rs.html I can not find any logging statements.
Maybe you meant https://github.com/kube-rs/kube/blob/d4bb23ef4c441708f0f7655ea804bcd23bff6f71/kube-runtime/src/controller/mod.rs#L1630?

NickLarsenNZ: Yeah I think it could have been that.

NickLarsenNZ (dismissed):
LGTM

Techassi (changes requested):
Blocking for now, because there are a whole bunch of notes I have on this. These are best discussed verbally instead of writing back and forth.

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,99 @@
...
+///     socket_addr: WebhookServer::DEFAULT_SOCKET_ADDRESS,
+///     webhook_namespace: "my-namespace".to_owned(),
+///     webhook_service_name: "my-operator".to_owned(),
+/// };
+/// let webhook_server = WebhookServer::new(webhook_options, webhooks).await.unwrap();
+/// # }
+/// ```
 pub struct WebhookServer {
+    options: WebhookServerOptions,
+    webhooks: Vec<Box<dyn Webhook>>,
```
Techassi, line 80: note: In my opinion, this crate is only aimed at implementing Kubernetes webhooks, not _any_ generic webhook. Kubernetes has a well-known set of supported webhooks: the conversion and admission webhooks. As such, we should use an enum in combination with `enum_dispatch`.

crates/stackable-webhook/src/lib.rs
```diff
@@ -1,92 +1,99 @@
...
     tls_server: TlsServer,
+    cert_rx: mpsc::Receiver<Certificate>,
+}
+
+#[derive(Clone, Debug)]
+pub struct WebhookServerOptions {
```
Techassi, line 87: note: This had a whole bunch of documentation. Why is this removed? I spent a significant amount of time writing up proper documentation for all the public items in this crate.

note: This had a builder. Why did it get removed?

crates/stackable-webhook/src/lib.rs
```diff
@@ -99,52 +106,13 @@ impl WebhookServer {
...
-    ///     .add_subject_alterative_dns_name("my-san-entry")
-    ///     .build();
-    ///
-    /// let router = Router::new();
-    /// let (server, cert_rx) = WebhookServer::new(router, options)
-    ///     .await
-    ///     .expect("failed to create WebhookServer");
-    /// # }
-    /// ```
+    /// Please read their documentation for details.
```
Techassi, line 114: note: Again, this had so much more documentation before, which is still valid after the change. Also, it should still provide a (basic) example on how to use it, and not just refer to the downstream impl.

crates/stackable-webhook/src/lib.rs
```diff
@@ -156,22 +124,33 @@ impl WebhookServer {
...
-        // TODO (@NickLarsenNZ): rename this server_builder and keep it specific to tracing, since it's placement in the chain is important
-        let service_builder = ServiceBuilder::new().layer(trace_layer);
+        let trace_service_builder = ServiceBuilder::new().layer(trace_layer);
 
         // Create the root router and merge the provided router into it.
         tracing::debug!("create core router and merge provided router");
+        let mut router = Router::new();
+        for webhook in &webhooks {
+            router = webhook.register_routes(router);
+        }
```
Techassi, line 137: suggestion: We can add a debug statement here to log the registration of routes we add the the core webhook server.

crates/stackable-webhook/src/lib.rs
```diff
@@ -200,19 +179,58 @@ impl WebhookServer {
...
+            tls_server,
+            mut cert_rx,
+        } = self;
+        let tls_server = tls_server
+            .run()
+            .map_err(|err| WebhookServerError::RunTlsServer { source: err });
+
+        let cert_update_loop = async {
+            loop {
+                while let Some(cert) = cert_rx.recv().await {
```
Techassi: question: Why do we have two nested loops here?

crates/stackable-webhook/src/servers/conversion.rs
Techassi, line 1: note: I'm basically missing 90% of the documentation from here in the new impl.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,201 @@
...
+    crds_and_handlers: Vec<(CustomResourceDefinition, H)>,
+    disable_crd_maintenance: bool,
+    client: Client,
+    /// The field manager used when maintaining the CRDs.
+    field_manager: String,
+    // This channel can only be used exactly once. The sender's send method consumes self, and
+    // as such, the sender is wrapped in an Option to be able to call take to consume the inner
+    // value.
+    initial_reconcile_tx: Option<oneshot::Sender<()>>,
+}
```
Techassi, line 99: note: This struct and it's fields need documentation.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl<H> ConversionWebhook<H> {
+    pub fn new(
+        crds_and_handlers: impl IntoIterator<Item = (CustomResourceDefinition, H)>,
+        disable_crd_maintenance: bool,
+        client: Client,
+        field_manager: String,
+    ) -> (Self, oneshot::Receiver<()>) {
```
Techassi, line 126: note: Please use an options struct for appropriate parameters like we did before.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,201 @@
...
+        (
+            Self {
+                crds_and_handlers: crds_and_handlers.into_iter().collect(),
+                disable_crd_maintenance,
+                client,
+                field_manager,
+                initial_reconcile_tx: Some(initial_reconcile_tx),
+            },
+            initial_reconcile_rx,
+        )
```
Techassi: note: This looks fairly cluttered, can we make this nicer to look at.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+#[async_trait]
+impl<H> Webhook for ConversionWebhook<H>
+where
+    H: FnOnce(ConversionReview) -> ConversionReview + Clone + Send + Sync + 'static,
+{
+    fn register_routes(&self, mut router: Router) -> Router {
+        for (crd, handler) in self.crds_and_handlers.clone() {
```
Techassi: question: Can we only clone the handler here as this seems to be the only thing we need to won at this point?

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[instrument(skip(self))]
+    async fn handle_certificate_rotation(
+        &mut self,
+        _new_certificate: &Certificate,
+        new_ca_bundle: &ByteString,
+        options: &WebhookServerOptions,
+    ) -> Result<(), WebhookError> {
+        if self.disable_crd_maintenance {
+            return Ok(());
+        }
```
Techassi: suggestion: We should maybe add a `is_enabled` function to the `Webhook` trait to basically skip adding this particular webhook when handling the rotation at the top-level `Webhookserver`.

crates/stackable-webhook/src/webhooks/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// The HTTPS server periodically rotates it's certificate.
+    ///
+    /// Typically, some caller of the webhook (e.g. Kubernetes) needs to know the certificate to be
+    /// able to establish the TLS connection.
+    /// Webhooks are informed about new certificates by this function and can react accordingly.
+    async fn handle_certificate_rotation(
+        &mut self,
+        new_certificate: &Certificate,
+        new_ca_bundle: &ByteString,
```
Techassi: note: We should only provide one of those two parameters, and I would go with the more generic one (the `Certificate`). All implementers of this trait only use one of them, which indicates that having both is unnecessary.

We can however discuss which one of the two parameters we pass, because if we are sure the base64-encoded PEM string is enough, we can get rid of the additional work done on every cert rotation.

crates/stackable-webhook/src/webhooks/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        options: &WebhookServerOptions,
+    ) -> Result<(), WebhookError>;
+}
+
+/// Returns the client config that can be used in admission webhooks.
+///
+/// It is used to concat the correct HTTP endpoint, which is calculated from the given parameters.
+/// (CRD conversions require a similar, but different, client config).
```
Techassi: note: The HTTP endpoint is passed straight into the client config, nothing is "calculated" here.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+    handler: H,
+    handler_state: Arc<S>,
+    _resource: PhantomData<R>,
+
+    disable_mutating_webhook_configuration_maintenance: bool,
+    client: Client,
+
+    /// The field manager used when maintaining the MutatingWebhookConfigurations.
+    field_manager: String,
+}
```
Techassi, line 131: note: These fields should be documented.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Your [`MutatingWebhookConfiguration`] can contain 0..n webhooks, but it is recommended to
+    /// only have a single entry in there, as the clientConfig of all entries will be set to the
+    /// same service, port and HTTP path.
+    pub fn new(
+        mutating_webhook_configuration: MutatingWebhookConfiguration,
+        handler: H,
+        handler_state: Arc<S>,
+        disable_mutating_webhook_configuration_maintenance: bool,
+        client: Client,
+        field_manager: String,
```
Techassi: note: we should bundle up a whole bunch of these parameters to avoid these needlessly large function signatures.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+
+impl<H, S, R> MutatingWebhook<H, S, R> {
+    /// All webhooks need to set the admissionReviewVersions to `["v1"]`, as this mutating webhook
+    /// only supports that version! A failure to do so will result in a panic.
+    ///
+    /// Your [`MutatingWebhookConfiguration`] can contain 0..n webhooks, but it is recommended to
+    /// only have a single entry in there, as the clientConfig of all entries will be set to the
+    /// same service, port and HTTP path.
+    pub fn new(
+        mutating_webhook_configuration: MutatingWebhookConfiguration,
```
Techassi, line 135: question: Should we maybe construct this on our own (opinionated) way? From the looks of it, it seems fairly painful to use...

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+        client: Client,
+        field_manager: String,
+    ) -> Self {
+        for webhook in mutating_webhook_configuration.webhooks.iter().flatten() {
+            assert_eq!(
+                webhook.admission_review_versions,
+                vec!["v1"],
+                "We decide how we de-serialize the JSON and with that what AdmissionReview version we support (currently only v1)"
+            );
+        }
```
Techassi, line 147: note: If we would construct those configs on our own, we could get rid of this.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+#[async_trait]
+impl<H, S, R, Fut> Webhook for MutatingWebhook<H, S, R>
+where
+    H: Fn(Arc<S>, AdmissionRequest<R>) -> Fut + Clone + Send + Sync + 'static,
+    Fut: Future<Output = AdmissionResponse> + Send + 'static,
+    R: Resource + Send + Sync + DeserializeOwned + Serialize + 'static,
+    S: Send + Sync + 'static,
+{
+    fn register_routes(&self, router: Router) -> Router {
+        let handler_state = self.handler_state.clone();
```
Techassi, line 174: note: This state can potentially mutate the Kubernetes state and as such would have side effects. In the future, we need to design this API in a way which properly needs to handle side effects and dry runs.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+        );
+
+        for webhook in mutating_webhook_configuration.webhooks.iter_mut().flatten() {
+            // We know how we can be called (and with what certificate), so we can always set that
+            webhook.client_config =
+                get_webhook_client_config(options, new_ca_bundle.to_owned(), self.http_path());
+        }
+
+        let mwc_api: Api<MutatingWebhookConfiguration> = Api::all(self.client.clone());
+        // Other than with the CRDs we don't need to force-apply the MutatingWebhookConfiguration
```
Techassi, line 221: note: I assume this is the case because this was (and probably never will be) maintained by Helm or any other deployment tool. If so, we should add this here as a comment.

crates/stackable-webhook/src/webhooks/validating_webhook.rs
```diff
@@ -0,0 +1,162 @@
...
+
+/// Validating webhook, which let's you intercept object creations/modification and allow or deny
+/// the object.
+///
+/// As the webhook is typed with the Resource type `R`, it can only handle a single resource
+/// validation. Use multiple [`ValidatingWebhook`] if you need to validate multiple resource kinds.
+///
+/// ### Example usage
+///
+/// TODO
```
Techassi: note: Well... We should add that...

crates/stackable-webhook/src/webhooks/validating_webhook.rs
```diff
@@ -0,0 +1,162 @@
...
+pub struct ValidatingWebhook<H, S, R> {
+    validating_webhook_configuration: ValidatingWebhookConfiguration,
+    handler: H,
+    handler_state: Arc<S>,
+    _resource: PhantomData<R>,
+
+    disable_validating_webhook_configuration_maintenance: bool,
```
Techassi: suggestion: This is a mouthful with a bunch of superfluous information: The fact that we are talking about validation webhooks is already indicated by the struct name itself and the maintenance _is_ dealing with the validating webhook config.

```suggestion
    disable_maintenance: bool,
```

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,217 @@
...
+///     AdmissionResponse::from(&request)
+/// }
+/// ```
+pub struct MutatingWebhook<H, S, R> {
+    mutating_webhook_configuration: MutatingWebhookConfiguration,
+    handler: H,
+    handler_state: Arc<S>,
+    _resource: PhantomData<R>,
+
+    disable_mutating_webhook_configuration_maintenance: bool,
```
Techassi: note: Same as in the validating webhook.

crates/stackable-webhook/src/webhooks/validating_webhook.rs
```diff
@@ -0,0 +1,162 @@
...
+    client: Client,
```
Techassi: note: Add documentation to the fields.

crates/stackable-webhook/src/webhooks/validating_webhook.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Your [`ValidatingWebhookConfiguration`] can contain 0..n webhooks, but it is recommended to
+    /// only have a single entry in there, as the clientConfig of all entries will be set to the
+    /// same service, port and HTTP path.
+    pub fn new(
+        validating_webhook_configuration: ValidatingWebhookConfiguration,
+        handler: H,
+        handler_state: Arc<S>,
+        disable_validating_webhook_configuration_maintenance: bool,
+        client: Client,
+        field_manager: String,
```
Techassi: note: We should reduce the number of parameters here.

Techassi (changes requested):

crates/stackable-webhook/src/lib.rs
sbernauer: Are you thinking of
```rust
tracing::debug!("Registering webhook routes");
```
or something more advanced? I can't really think of what else to log here :see_no_evil:

crates/stackable-webhook/src/lib.rs
sbernauer: I don't know :sweat_smile: Removed the outer loop in aedb264

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
sbernauer: I moved the Self{} into it's own variable in 9494797, hope that works

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
sbernauer: Yes, I can offer
```diff
diff --git a/crates/stackable-webhook/src/webhooks/conversion_webhook.rs b/crates/stackable-webhook/src/webhooks/conversion_webhook.rs
index e01b571..b51b195 100644
--- a/crates/stackable-webhook/src/webhooks/conversion_webhook.rs
+++ b/crates/stackable-webhook/src/webhooks/conversion_webhook.rs
@@ -155,7 +155,8 @@ where
     H: FnOnce(ConversionReview) -> ConversionReview + Clone + Send + Sync + 'static,
 {
     fn register_routes(&self, mut router: Router) -> Router {
-        for (crd, handler) in self.crds_and_handlers.clone() {
+        for (crd, handler) in &self.crds_and_handlers {
+            let handler = handler.clone();
             let crd_name = crd.name_any();
             let handler_fn = |Json(review): Json<ConversionReview>| async {
                 let review = handler(review);
```
However, this way we have n memory allocations instead of a single (bigger one), not sure what is more efficient.
In any case this is called one per startup, so absolutely doesn't matter at all.
Picked your suggestion in 3a70a42

crates/stackable-webhook/src/webhooks/mod.rs
sbernauer: I'd say the calculation happens in Kubernetes, where it takes a bunch of fields and constructs a URI out of it.
But nevertheless changed it to `determined`in 67864fa, how that works for you

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: Added a little bit of clarification in 552e86c

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
sbernauer: We should not call it `is_eanbled`, as the webhook itself is always enabled. The question is if it does something on cert rotation.
Added in 85481e8, although not really sure if this improve readability

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
sbernauer: Added some field level docs in 1c4e2ad, struct is follow-up

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: Added some field level docs in 1c4e2ad

crates/stackable-webhook/src/webhooks/validating_webhook.rs
sbernauer: Added some field level docs in 1c4e2ad

crates/stackable-webhook/src/lib.rs
Techassi: Ah, we don't have access to _which_ route we are adding at this point, but we can maybe add such a log statement where we do know the routes.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
Techassi: Yeah, this doesn't improve a whole lot. I'll let you decide which approach you like better.

crates/stackable-webhook/src/lib.rs
sbernauer: Added two debug statements in fe28485

crates/stackable-webhook/src/webhooks/validating_webhook.rs
sbernauer: I removed the validating webhook from this PR, as a correct implementation shares code with the mutating webhook -> follow up

crates/stackable-webhook/src/webhooks/validating_webhook.rs
sbernauer: I removed the validating webhook from this PR, as a correct implementation shares code with the mutating webhook -> follow up

crates/stackable-webhook/src/webhooks/validating_webhook.rs
sbernauer: I removed the validating webhook from this PR, as a correct implementation shares code with the mutating webhook -> follow up

crates/stackable-webhook/src/lib.rs
sbernauer: I re-added updated docs in 0a2014c

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: I would be happy to do it in the same way as the conversion webhook does it (`disable_crd_maintenance`) and call it `disable_mwc_maintenance`. Opinions?
I don't like `disable_maintenance` too much, as it hides what is being maintained (we just happen to know that it's a MWC)

crates/stackable-webhook/src/lib.rs
```diff
@@ -200,19 +179,60 @@ impl WebhookServer {
...
+
+                    webhook
+                        .handle_certificate_rotation(&cert, &ca_bundle, &options)
+                        .await
+                        .context(UpdateCertificateSnafu)?;
+                }
+            }
+
+            // We need to hint the return type to the compiler
+            #[allow(unreachable_code)]
```
Techassi: note: This should get removed now as the superfluous `loop` got removed.

```suggestion
```

sbernauer: Ahhh now I also makes sense why it said it's unreachable...
790bd0a74ee00ace441a33de09546ae2b9cdf1bf

crates/stackable-webhook/src/webhooks/validating_webhook.rs
sbernauer: I removed the validating webhook from this PR, as a correct implementation shares code with the mutating webhook -> follow up

crates/stackable-webhook/src/lib.rs
Techassi: We will leave it as is for now and I will explore this in the next couple of days.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
```diff
@@ -0,0 +1,270 @@
...
+            let handler = handler.clone();
+            let crd_name = crd.name_any();
+            let handler_fn = |Json(review): Json<ConversionReview>| async {
+                let review = handler(review);
+                Json(review)
+            };
+
+            let route = format!("/convert/{crd_name}");
+            tracing::debug!(
+                crd.name = crd_name,
```
Techassi: suggestion: Use the semantic conventions we defined.

```suggestion
                k8s.crd.name = crd_name,
```

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
Techassi: The design of this API will most likely change once we support validating webhooks. We want to continue exploring this in the future.

crates/stackable-webhook/src/webhooks/mod.rs
Techassi: Resolved in 054fe29.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
Techassi: Resolved in 054fe29.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
Techassi: Resolved in 054fe29.

crates/stackable-webhook/src/lib.rs
Techassi: Can we please re-add a few lines of documentation to the struct itself?

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
Techassi: It might also make sense to add the CRD kind here - similar to what we do in the `reconcile_crd` function.

crates/stackable-webhook/src/webhooks/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+        new_ca_bundle: &ByteString,
+        options: &WebhookServerOptions,
+    ) -> Result<(), WebhookError>;
+}
+
+/// Returns the client config that can be used in admission webhooks.
+///
+/// It is used to contact the correct HTTP endpoint, which is determined from the given parameters.
+/// (CRD conversions require a similar, but different, client config).
+fn get_webhook_client_config(
```
Techassi: note: The verb "get" implies we get this config from _somewhere_ but we are actually constructing it in-place. As such I would recommend using create, build or construct instead.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,233 @@
...
+    mutating_webhook_configuration: MutatingWebhookConfiguration,
+
+    /// The async handler that get's a [`AdmissionRequest`] and returns an [`AdmissionResponse`]
+    handler: H,
+
+    /// The internal state of the webhook. You can define yourself what exactly this state is.
+    handler_state: Arc<S>,
+
+    /// This field is not needed, it only tracks the type of the Kubernetes resource we are mutating
+    _resource: PhantomData<R>,
```
Techassi: note: I recommend moving this all the way to the end, as this only holds generic type information and no actual data.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
```diff
@@ -0,0 +1,233 @@
...
+    client: Client,
+}
+
+pub struct MutatingWebhookOptions {
+    /// Whether MutatingWebhookConfigurations should be maintained
+    pub disable_mwc_maintenance: bool,
+
+    /// The field manager used when maintaining the MutatingWebhookConfigurations
+    pub field_manager: String,
+}
```
Techassi, line 131: note: Please add a few lines of documentation to the struct itself.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
Techassi: We will leave this as is for now - to be explored in the future.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
Techassi: That's sadly we can most likely not design in any better way, leaving as is for now.

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
Techassi: Resolved in 054fe29.

crates/stackable-webhook/src/webhooks/conversion_webhook.rs
sbernauer: Added in ec5a33a

crates/stackable-webhook/src/webhooks/mod.rs
sbernauer: Renamed to create_webhook_client_config in 69603ca

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: 1b257d3

crates/stackable-webhook/src/lib.rs
sbernauer: 1b257d3

crates/stackable-webhook/src/webhooks/mutating_webhook.rs
sbernauer: Moved in cacab52

crates/stackable-webhook/src/lib.rs
```diff
@@ -200,19 +182,58 @@ impl WebhookServer {
...
-        self.tls_server.run().await.context(RunTlsServerSnafu)
+        let Self {
+            options,
+            mut webhooks,
+            tls_server,
+            mut cert_rx,
+        } = self;
+        let tls_server = tls_server
+            .run()
+            .map_err(|err| WebhookServerError::RunTlsServer { source: err });
```
Techassi, line 223: note: I think we should just return here, if the Vec of webhooks is empty. There is no need to run the TLS server and the certificate renewal.

sbernauer: Hmm, sounds like a very minor optimization to me to be honest. I think we should not return, but block forever, as otherwise WebhookServer::run will return early as well and some downstream select! might stop the operator or other stuff.
WDYT?

Techassi: Downstream code should use the correct joining mechanism (in most cases keep running until everything returns `OK(...)` or one returns `Err(...)`). Unnecessarily blocking here would be the wrong thing to do.

To make sure users choose the correct mechanism, we could add one or two sentences to the doc comments for the `run` function.

Techassi: Added in aa4f9ec3.

sbernauer: LGTM, thanks!

PR #1122, opened 2025-12-04
chore: Add unwrap and panic lints

Techassi:
This PR adds various lints around `unwrap`s and `panic`s.

Both `unwrap`s and `panic`s are allowed in tests to ease the transition. In the future, tests should be restructured to get rid of these currently allowed pieces of code.


maltesander (approved):
LGTM!

PR #1124, opened 2025-12-08
Depend on the patched version of kube-rs

siegfriedweber:
# Description

Depend on the patched version of kube-rs available at https://github.com/stackabletech/kube-rs, ensuring the operators automatically benefit from the fixes.

The patched version of kube-rs avoids that `null` entries are not added to enums, see stackabletech/kube-rs#1. These `null` entries currently prevent the CRD documentations from being rendered.

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -13,7 +13,7 @@ All notable changes to this project will be documented in this file.
 
 - BREAKING: `ClusterResources` now requires the objects added to implement `DeepMerge`.
   This is very likely a stackable-operator internal change, but technically breaking ([#1118]).
-- Depend on the patched version of kube-rs available at https://github.com/stackabletech/kube-rs,
+- Depend on the patched version of kube-rs available at <https://github.com/stackabletech/kube-rs>,
   ensuring the operators automatically benefit from the fixes ([#1124]).
 
 ### Removed


Cargo.toml
```diff
@@ -37,8 +37,8 @@ json-patch = "4.0.0"
 k8s-openapi = { version = "0.26.0", default-features = false, features = ["schemars", "v1_34"] }
 # We use rustls instead of openssl for easier portability, e.g. so that we can build stackablectl without the need to vendor (build from source) openssl
 # We use ring instead of aws-lc-rs, as this currently fails to build in "make run-dev"
-# We pin the kube version, as we use a patch for 2.0.1 below
-kube = { version = "=2.0.1", default-features = false, features = ["client", "jsonpatch", "runtime", "derive", "rustls-tls", "ring"] }
+# We pin the kube version, as we use a patch for 2.0.1
+kube = { git = "https://github.com/stackabletech/kube-rs", branch = "2.0.1-fix-schema-hoisting", version = "=2.0.1", default-features = false, features = ["client", "jsonpatch", "runtime", "derive", "rustls-tls", "ring"] }
```
razvan, line 41: Ignoring the problem that triggered this PR, what is the benefit of making the patched version explicit here?
I thought this is exactly the same as having a `[patch]` table.

siegfriedweber: `[patch]` is only applied, if you compile this project. If you add stackable-operator as dependency to an operator, then `[patch]` is ignored and the upstream kube-rs version is used.

It would be possible to add a `[patch]` section to each operator, like it was done in https://github.com/stackabletech/listener-operator/blob/25.11.0/Cargo.toml#L42-L43. But I think, it is better if stackable-operator just defines the correct (patched) version.

razvan: Hm, I guess [this paragraph](https://doc.rust-lang.org/1.80.0/cargo/reference/overriding-dependencies.html#working-with-an-unpublished-minor-version) tries to explain it but I find it does a rather poor job at it. Oh well ...

PR #1125, opened 2025-12-19
chore: Let rustfmt format code in doc comments

sbernauer:
Enable `format_code_in_doc_comments` and let `cargo +nightly-2025-10-23 fmt --all` format the code.

The unstable feature isn't perfect, but in my opinion this is a significant improvement.

### DIFF
--- a/crates/stackable-operator/crds/DummyCluster.yaml
+++ b/crates/stackable-operator/crds/DummyCluster.yaml
@@ -111,6 +111,7 @@ spec:
                               - required:
                                 - custom
                               - {}
+                              - {}
                               description: Log configuration of the container
                               properties:
                                 console:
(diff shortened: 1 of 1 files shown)


crates/stackable-telemetry/src/tracing/mod.rs
```diff
@@ -83,12 +83,11 @@ pub enum Error {
...
-///         .build()                            // |
-///         .init()?;                           // |
-///                                             // |
-///     tracing::info!("log a message");        // |
-///     Ok(())                                  // < Scope ends here, guard is dropped
+///         .service_name("test")
+///         .build()
+///         .init()?;
+///     tracing::info!("log a message");
+///     Ok(()) // < Scope ends here, guard is dropped
```
Techassi: This is a change we don't want.

crates/stackable-versioned-macros/src/lib.rs
Techassi, line 1: Most of these changes are a little weird.

crates/stackable-telemetry/src/tracing/mod.rs
NickLarsenNZ: If there's a lint that allows for aligned comments, we could enable it here?
But we also want the empty lines.

Techassi: > If there's a lint that allows for aligned comments

I have no clue if such a thing exists... But others also asked for this: https://users.rust-lang.org/t/rustfmt-and-alignment-of-trailing-comments/39121

I think the best option would be to use `#[rustfmt::skip]` here.

sbernauer: My first try with `#[rustfmt::skip]` didn't work, now it did 🤷 7690e04

Techassi: This will technically disable formatting for the actual code (struct) as well, but I guess there is no other way around it. Or can we disable doc code formatting just for these doc comments?

sbernauer: I googled a bit and could sadly not find a solution other than rustfmt::skip

Techassi: Mhm bummer, something we have to live with for now I guess.

PR #1128, opened 2025-12-22
docs(stackable-versioned): Add section about hint(...)

Techassi:
This PR adds a section about `#[versioned(hint(...))]` to the docs.

This came up after @adwk67 wanted to use this argument and pointed out that this is not covered by our documentation.


crates/stackable-versioned-macros/src/lib.rs
```diff
@@ -645,6 +645,41 @@ mod utils;
...
+///
+/// With `#[versioned(hint(...))]` it is possible to give hints to the macro
+/// that the field contains a wrapped type. Currently, these following hints
+/// are supported:
+///
+/// - `hint(option)`: Indicates that the field contains an `Option<T>`.
+/// - `hint(vec)`: Indicates that the field contains a `Vec<T>`.
+///
+/// These hints are especially useful for generated conversion functions. With
+/// these hints in place, the types are correctly mapped using `Into::into`.
```
adwk67: My understanding is that this is done by using whatever is defined for `upgrade_with` and `downgrade_with` on the field itself. Is it worth mentioning that here for clarity?

Techassi: Both `upgrade_with` and `downgrade_with` are meant to give the user complete control of the conversion. There are however cases, where the wrapped type already implements conversions (automatically generated or manually implemented), but the macro cannot reason about the wrapped types. An example:

- Without `hint(vec)`: Generated code will try to convert a `Vec<T>` by calling `old_field.into()`. This will fail, because there is no appropriate `From` implementation for `Vec<T>`, but only for `T`.
- With `hint(vec)`: The generated code now maps each value of the `Vec<T>`, which will result in the correct `Vec<U>` afterwards.

Currently, both `upgrade_with`/`downgrade_with` and `hint` kind of "conflict" with each other/produce unexpected code. I've already raised an improvement for this. I think the expected behaviour should be like this:

- No `hint`, only `upgrade_with`/`downgrade_with`: The conversion function(s) need to deal with the complete type.
- With hint and `upgrade_with`/`downgrade_with`: The inner values are mapped using the defined conversion functions instead of the default `Into::into`.

Do you want any of this mentioned in the docs?

adwk67: What confused me was having nested versioning, whereby I have `upgrade_with/downgrade_with` on the field itself, and a second conversion (instead of a `hint`) that basically called those functions in a manual mapping step in the "parent" struct. So ~it would be a case of~ I'm now using `hint` (on the parent), implicitly calling `upgrade_with/downgrade_with` on the child. At least, I _think_ that is what is happening.

adwk67: (Specifically, `upgrade_with/downgrade_with` on a field in the gitsync struct, and a `hint` on the the parent struct that includes the gitsync struct as a field).

Techassi: > What confused me was having nested versioning, whereby I have upgrade_with/downgrade_with on the field itself, and a second conversion (instead of a hint) that basically called those functions in a manual mapping step in the "parent" struct.

That's what I basically described above, because the macro cannot reason about wrapped types. Therefor it would complain about missing `From` impls, which then led you to implement manual conversions, even though they weren't needed (with the usage of `hint`).

> I'm now using hint (on the parent), implicitly calling upgrade_with/downgrade_with on the child. At least, I think that is what is happening.

Both these functions are not needed/currently used by `hint`. Here is an example of the generated code:

```rust
impl From<v1alpha1::AirflowClusterConfig> for v1alpha2::AirflowClusterConfig {
    fn from(__sv_airflowclusterconfig: v1alpha1::AirflowClusterConfig) -> Self {
        Self {
            dags_git_sync: __sv_airflowclusterconfig
                .dags_git_sync
                .into_iter()
                .map(Into::into)
                .collect(),
            // Other fields ...
        }
    }
}
```

You can expand this code yourself by either using `cargo-expand` or via rust-analyzer by placing your caret on the top-level `#[versioned]` macro, opening the command prompt (usually `Ctrl+Shift+P` in VS Code), and then selecting `rust-analyzer: Expand macro recursively at caret`.

adwk67: After we discussed this yesterday, I would maybe go with something like this:

> With these hints in place, the types are correctly mapped using `Into::into` (assuming the necessary `From` trait methods are implemented on the target types for the conversion to be done correctly).

Techassi: Sure, I can add that.

Techassi: Adjusted in efc60efc.

adwk67: Thanks!

adwk67 (approved):

PR #1132, opened 2025-12-23
fix: Only take the object_overrides by reference in ClusterResources

sbernauer:
Small optimization.

Will be released as 0.101.1


Maleware (approved):
LGTM! Thanks

PR #1135, opened 2026-01-09
chore(stackable-operator): Revert and pin k8s-openapi to 0.26.0

NickLarsenNZ:
Newer versions of k8s-openapi surface a schemars bug for optional quantities. See: https://github.com/kube-rs/kube/issues/1869

See also https://github.com/kube-rs/kube/pull/1885#issuecomment-3714312880

> [!TIP]
> Operators might need to pin the dependency too.
> See both Cargo.toml files in https://github.com/stackabletech/airflow-operator/pull/730

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- Revert and pin k8s-openapi to 0.26.0 ([#1135]).
+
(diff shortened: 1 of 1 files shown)


sbernauer (approved):
LGTM

> Operators need to pin the dependency too.

Not sure if that really is the case :thinking:

NickLarsenNZ:
> LGTM
> 
> > Operators need to pin the dependency too.
> 
> Not sure if that really is the case 🤔

I changed the wording to _might_. I think they do though (otherwise there is no bug to work around).

PR #1144, opened 2026-01-29
feat(webhook): Add graceful shutdown

Techassi:
Part of https://github.com/stackabletech/issues/issues/803.

This adds support to gracefully shutdown the `WebhookServer` and `TlsServer`. This allows callers (like our operators) to gracefully shutdown all concurrent tasks when a SIGTERM (or any other) signal is received.

### DIFF
--- a/crates/stackable-webhook/src/tls/mod.rs
+++ b/crates/stackable-webhook/src/tls/mod.rs
@@ -190,8 +190,9 @@ impl TlsServer {
                 // interval. This ensures, we always use a valid certificate for the TLS connection.
                 biased;
 
-                // As soon as this future because `Poll::Ready`, break out of the loop which cancels
-                // the certification rotation interval and stops accepting new TCP connections.
+                // Once a shutdown signal is received (this future becomes `Poll::Ready`), break out
+                // of the main loop which cancels the certification rotation interval and stops
+                // accepting new TCP connections.
                 _ = &mut shutdown_signal => {
                     tracing::trace!("received shutdown signal");
                     break;


crates/stackable-webhook/src/tls/mod.rs
```diff
@@ -161,24 +172,35 @@ impl TlsServer {
...
-                // We opt for a biased execution of arms to make sure we always check if the
-                // certificate needs rotation based on the interval. This ensures, we always use
-                // a valid certificate for the TLS connection.
+                // We opt for a biased execution of arms to make sure we always check if a
+                // shutdown signal was received or the certificate needs rotation based on the
+                // interval. This ensures, we always use a valid certificate for the TLS connection.
                 biased;
 
+                // As soon as this future because `Poll::Ready`, break out of the loop which cancels
+                // the certification rotation interval and stops accepting new TCP connections.
```
maltesander: I find this hard to understand.
```suggestion
                // Triggering Poll::Ready on this future breaks the main loop, which initiates the shutdown.
                // New TCP connections are rejected and the certificate rotation interval is cleared.
```

crates/stackable-webhook/src/lib.rs
```diff
@@ -243,6 +209,8 @@ impl WebhookServer {
             Ok(())
         };
 
+        // This either returns if one of the two futures complete with Err(_) or when both complete
```
maltesander:
```suggestion
        // This either returns if one of the two futures completes with Err(_) or when both complete
```

crates/stackable-webhook/src/tls/mod.rs
Techassi: What do you think about this:


```suggestion
                // Once a shutdown signal is received (this future becomes `Poll::Ready`), break out
                // of the main loop which cancels the certification rotation interval and stops
                // accepting new TCP connections.
```

PR #1147, opened 2026-01-30
feat(operator): Add SignalWatcher

Techassi:
Part of https://github.com/stackabletech/issues/issues/803.

> [!NOTE]
> Internally, we could use an [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) instead of a watch channel, but this would require us to add `tokio_utils` as a dependency. `CancellationToken` on it's own is also not sufficient, as the cancellation needs to be triggered by a signal (like SIGTERM). The `SignalWatcher` abstracts waiting for the signal and notifying downstream tasks all in one place.

This PR adds a new `SignalWatcher` which can be used to watch signals and multiply them to gracefully shutdown multiple concurrent tasks/futures.

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -6,9 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- BREAKING: Add support to gracefully shutdown `EosChecker`.
+  `EosChecker::run` now requires passing a shutdown signal, which is any `Future<Output = ()>` ([#1146]).
 - Add `SignalWatcher` which can be used to watch signals and multiply them to gracefully shutdown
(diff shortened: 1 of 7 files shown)


crates/stackable-operator/src/utils/signal.rs
```diff
@@ -0,0 +1,73 @@
+use snafu::{ResultExt, Snafu};
+use tokio::{
+    signal::unix::{SignalKind, signal},
+    sync::watch,
+};
+
+#[derive(Debug, Snafu)]
+#[snafu(display("failed to construct signal listener"))]
```
maltesander: Listener is already so overloaded :)
```suggestion
#[snafu(display("failed to construct signal watcher"))]
```

maltesander (dismissed):
LGTM!

PR #1183, opened 2026-03-30
fix(stackable-telemetry): Disable ANSI colors when stdout is not a terminal

lfrancke:
## Summary

- Only emit ANSI escape codes in console log output when stdout is actually a terminal
- Applies to both `Plain` and `Json` console log formats

## Problem

`tracing_subscriber::fmt::layer()` enables ANSI colors by default whenever the `ansi` feature is enabled and `NO_COLOR` is unset ([source](https://docs.rs/tracing-subscriber/0.3.23/src/tracing_subscriber/fmt/fmt_layer.rs.html#743)):

```rust
let ansi = cfg!(feature = "ansi") && env::var("NO_COLOR").map_or(true, |v| v.is_empty());
```

There is no automatic TTY detection — this is a [known upstream issue](https://github.com/tokio-rs/tracing/issues/1160) open since 2021. As a result, containerdebug (and any other consumer of stackable-telemetry) emits raw ANSI codes when running in containers where logs are collected by aggregation systems like Graylog or Vector:

```
[2m2026-03-29T18:26:23.658052Z[0m [32m INFO[0m [1mcontainerdebug run[0m...
```

## Fix

Explicitly check `std::io::stdout().is_terminal()` before configuring the subscriber, so ANSI is only used when writing to an actual terminal.

## Test plan

- [x] `cargo check -p stackable-telemetry` passes
- [x] Verify in a terminal: colored output still works
- [x] Verify in a container/pipe: ANSI codes are suppressed

### DIFF
--- a/crates/stackable-telemetry/CHANGELOG.md
+++ b/crates/stackable-telemetry/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- Only use ANSI escape sequences if stdout is a terminal/tty. Piping the output to a file will now
+  result in plain text log messages ([#1183]).
+
+[#1183]: https://github.com/stackabletech/operator-rs/pull/1183
+
 ## [0.6.2] - 2026-03-09
 
 ### Fixed
--- a/crates/stackable-telemetry/src/tracing/mod.rs
+++ b/crates/stackable-telemetry/src/tracing/mod.rs
@@ -6,7 +6,7 @@
 //!
 //! To get started, see [`Tracing`].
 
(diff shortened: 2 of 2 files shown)


NickLarsenNZ (dismissed):
LGTM

Techassi (changes requested):
This should include a changelog entry. Otherwise looks good.

Techassi (approved):

sbernauer:
Sadly this means we loose colored output, e.g. in k9s.

Is this intentional or can you think of any way to restore it? :)

image

sbernauer:
Raised https://github.com/stackabletech/operator-rs/pull/1237 with a different approach

PR #1199, opened 2026-04-21
feat(operator)!: Support dynamic product image selection

Techassi:
Part of https://github.com/stackabletech/issues/issues/716. This PR introduces the following changes:

- A new, required CLI argument and env var `IMAGE_REPOSITORY` (`--image-repository`): Set the image repository which should be used by the operator to construct image names for provisioned containers, eg. `oci.stackable.tech/sdp`.
- Adjustments to the product image selection to use dynamic registry and repository values instead of using hard-coded values.

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- BREAKING: Add two required CLI arguments and env vars to set image registry and repository ([#1199]):
+  - `IMAGE_REGISTRY` (`--image-registry`):  Sets the image registry which should be used by the
+    operator to construct image names for provisioned containers, eg. `oci.example.org`.
+  - `IMAGE_REPOSITORY` (`--image-repository`): Sets the image repository which should be used by the
+    operator to construct image names for provisioned containers, eg. `my/repository/to/operator`.
+
+### Changed
+
+- BREAKING: The product image selection mechanism via `ProductImage::resolve` now takes three
+  parameters instead of two. The new parameters are: `image_registry`, `image_repository`, and
+  `operator_version`.
+- BREAKING: The product image selection CRD interface splits up the `repo` key into `registry` and
+  `repository` for more clarity and consistency ([#1199]).
+
+[#1199]: https://github.com/stackabletech/operator-rs/pull/1199
+
 ## [0.110.1] - 2026-04-16
 
 ### Added
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -6,19 +6,14 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- BREAKING: Add two required CLI arguments and env vars to set image registry and repository ([#1199]):
-  - `IMAGE_REGISTRY` (`--image-registry`):  Sets the image registry which should be used by the
-    operator to construct image names for provisioned containers, eg. `oci.example.org`.
-  - `IMAGE_REPOSITORY` (`--image-repository`): Sets the image repository which should be used by the
-    operator to construct image names for provisioned containers, eg. `my/repository/to/operator`.
+- BREAKING: Add CLI argument and env var to set the image repository used to construct final product
+  image names: `IMAGE_REPOSITORY` (`--image-repository`), eg. `oci.example.org/my/namespace` ([#1199]).
 
 ### Changed
 
 - BREAKING: The product image selection mechanism via `ProductImage::resolve` now takes three
-  parameters instead of two. The new parameters are: `image_registry`, `image_repository`, and
-  `operator_version`.
-- BREAKING: The product image selection CRD interface splits up the `repo` key into `registry` and
-  `repository` for more clarity and consistency ([#1199]).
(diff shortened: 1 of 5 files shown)


NickLarsenNZ:
Just writing this out with some real examples:

```yaml
spec:
  image:
    productVersion: 3.3.1
    stackableVersion: 25.7.0 # Optional
    # the registry just happens to be at the root `/`
    registry: oci.stackable.tech
    # the repository structure is all the same from here (once we make it that way)
    repository: sdp/product
    # Q: So how the does UIF or bundle builder gets it's location currently?
    # A: I think this actually comes from the operator since it is the same code
```

```yaml
spec:
  image:
    productVersion: 3.3.1
    stackableVersion: 25.7.0 # Optional
    # This is a shared service, our registry (called a namespace) is under /stackable
    registry: quay.io/stackable
    # the repository structure is all the same from here (once we make it that way)
    repository: sdp/product # was just `product` in the flat mirrored structure
    # Q: So how will UIF or bundle builder gets it's sauce
    # A: I think this is the same image as the operator, so nothing related to the product
```

A simpler form (unless we have a reason to know about the registry itself would just be:

```yaml
spec:
  image:
    productVersion: 3.3.1
    stackableVersion: 25.7.0 # Optional
    # the repository structure is all the same from here (once we make it that way)
    repository: oci.stackable.tech/sdp/product
```

```yaml
spec:
  image:
    productVersion: 3.3.1
    stackableVersion: 25.7.0 # Optional
    # the repository structure is all the same from here (once we make it that way)
    repository: quay.io/stackable/sdp/product # was quay.io/stackable/product in the flat structure
```

---

Do we ever need to append anything to that, other than the product?

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
...
+    operator to construct image names for provisioned containers, eg. `oci.example.org`.
+  - `IMAGE_REPOSITORY` (`--image-repository`): Sets the image repository which should be used by the
+    operator to construct image names for provisioned containers, eg. `my/repository/to/operator`.
+
+### Changed
+
+- BREAKING: The product image selection mechanism via `ProductImage::resolve` now takes three
+  parameters instead of two. The new parameters are: `image_registry`, `image_repository`, and
+  `operator_version`.
+- BREAKING: The product image selection CRD interface splits up the `repo` key into `registry` and
```
sbernauer: Asking the obvious question: How hard would it be to use CRD versioning for this?

NickLarsenNZ:
Just another idea...

Currently we have:

```yaml
spec:
  image:
    # this is needed regardless of variant
    productVersion: 3.3.1
    # ProductImageSelection::StackableVersion
    stackableVersion: 25.7.0
    repo: oci.stackable.tech/sdp # or quay.io/stackable/sdp - NO PRODUCT INFO HERE
    # ProductImageSelection::Custom
    custom: oci.stackable.tech/sdp/hadoop:3.3.1-stackable25.7.0 # or anything you want
```

What about we just get rid of the variants, and make it _image_:

If the operator requires this `--image-repository=oci.stackable.tech/sdp`... (whether one or two options... I prefer one), it can construct the rest of the image value.
And, it can still be overridden.

```yaml
spec:
  # sucks that this is called image... maybe we should revise that since it would be a breaking change anyway.
  image:
    productVersion: 3.3.1
    # The stackableVersion field _could_ remain, but become optional if we want to allow that.
    stackableVersion: 25.7.0 # defaults to 26.3.0
    # Make image field optional (as it can still be constructed by the above information and constants in the operator).
    # image: oci.stackable.tech/sdp/hadoop:3.3.1-stackable25.7.0
    # The admin/user can still override the image:
    image: localboi/whatami
```

crates/stackable-operator/CHANGELOG.md
Techassi: This depends on what the exact final CRD interface will look like. I will come back to this once the design is finalized.

Techassi: This change is now completely non-breaking from a CRD view, marking as resolved.

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -106,44 +115,74 @@ pub enum PullPolicy {
...
+    /// ### Resolve mechanism
+    ///
+    /// The final product image is resolved in one of two ways defined by the [`ProductImageSelection`]:
+    ///
+    /// 1. When [`ProductImageSelection::Auto`] is selected by the user, the final product image
+    ///    will be constructed based on the (user) provided values.
+    /// 2. When [`ProductImageSelection::Custom`] is selected by the user, the final product image
+    ///    will be the exact value specified by the user.
+    //
+    // NOTE (@Techassi): The operator_version should probably be a Semver instead of a plain string
```
NickLarsenNZ, line 140: The version is CalVer. It looks the same, but if it were treated as SemVer, then the month would appear to indicate "minor", which our releases aren't.

That said, we could use (or make) a CalVer to use in place of the string.

NickLarsenNZ (approved):
LGTM

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: > That said, we could use (or make) a CalVer to use in place of the string.

Yea, I guess that is something we could do.

PR #1205, opened 2026-04-28
feat!: Add `(Pod)SecurityContextBuilder::with_stackable_defaults`

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/645

### Author

- [x] Changes are OpenShift compatible - To the best of my knowledge
- [x] CRD changes approved

### Reviewer

- [x] Changelog updated

### Acceptance

### Release notes (copy for operator changelogs)

```markdown
### Changed

- All product containers now run with `securityContext.runAsNonRoot` set to `true` to improve security.

### Fixed

- Previously, ArgoCD reported a diff for every Stackable customer resource deployed.
  In this release, we fix the longstanding problem of including empty `categories`, `shortNames` and `additionalPrinterColumns` in the CRDs,
  which could cause problems with GitOps tools (e.g. ArgoCD) reporting a diff in the custom resources.
  See [our internal issue](https://github.com/stackabletech/hdfs-operator/issues/626) and [the fix](https://github.com/kube-rs/kube/pull/2042) for details.
```


crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -144,14 +144,42 @@ impl SecurityContextBuilder {
...
+            pod_security_context: Self::stackable_default_pod_security_context(),
+        }
+    }
+
+    /// The Stackable's defaults for a [`PodSecurityContext`].
+    ///
+    /// It is recommended to use the [`PodSecurityContextBuilder::with_stackable_defaults`] instead
+    /// (if possible).
+    pub fn stackable_default_pod_security_context() -> PodSecurityContext {
+        todo!("Lars needs to define the exact settings he wants");
```
sbernauer: A TODO for @lfrancke :)

crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -144,14 +144,42 @@ impl SecurityContextBuilder {
...
+#[derive(Clone, Debug)]
 pub struct PodSecurityContextBuilder {
     pod_security_context: PodSecurityContext,
 }
 
 impl PodSecurityContextBuilder {
-    pub fn new() -> Self {
-        Self::default()
+    /// Construct a new [`PodSecurityContextBuilder`] that is pre-filled with Stackable's defaults.
+    pub fn with_stackable_defaults() -> Self {
```
siegfriedweber, line 171: - I prefer a vendor neutral term like `with_recommended_settings`, similar to https://search.nixos.org/options?channel=25.11&query=services.nginx.recommended.
- I would keep the `new` function and not force the settings yet: `new().with_recommended_settings()`. On the one hand, this pull request can then be merged now, and on the other hand, it is not guaranteed that we will be able to roll out this change on all operators at the same time. If we want to force the recommended settings later (which I would not do), we could create a function like `new_with_recommended_settings` and remove the `new` function.
- I would already use the builder functions here, to ensure that builder functions exist to override these settings.
- I asked for the `stackable_default_pod_security_context` function but it is not necessary and can be removed. The recommended settings can also be retrieved via `PodSecurityContextBuilder::new().with_recommended_settings().build()`.

crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -144,14 +144,42 @@ impl SecurityContextBuilder {
...
+///     // Configure any arbitrary fields
+///     .run_as_user(1234)
+///     .build();
+/// ```
+#[derive(Clone, Debug)]
 pub struct PodSecurityContextBuilder {
     pod_security_context: PodSecurityContext,
 }
 
 impl PodSecurityContextBuilder {
```
siegfriedweber, line 165: There should also be defaults for the `SecurityContextBuilder`.

https://github.com/stackabletech/operator-rs/blob/stackable-operator-0.111.1/crates/stackable-operator/src/builder/pod/security.rs#L8

crates/stackable-operator/src/builder/pod/security.rs
siegfriedweber: I would just leave the `with_recommended_settings()` function empty for now and keep the TODO in the issue, so that this pull request can be merged.

siegfriedweber (changes requested):

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: No objection against `with_recommended_settings`, so works for me. Hhowever to me this boils down to "is stackable-operator only for use or are other users also expected to use this?". I think we have different opinions in the team ^^ Discussion for later.

> If we want to force the recommended settings later (which I would not do)

I though that was the point of this PR ^^ If some defaults don't work for a tool it can manually overwrite them.

> I would keep the new function and not force the settings yet. On the one hand, this pull request can then be merged now

If we add a no-op function and rolling that out everywhere later one I have a bad feeling about e.g. setting the root filesystem to read only. If we define the defaults before merging we have a clearly breaking change. And not some operator-rs bump later on breaks everything silently. But no strong opinion.

siegfriedweber: > If some defaults don't work for a tool it can manually overwrite them.

My assumption was that the actual settings in this function are an implementation detail and should be a black box for the caller. So, the caller does not know which settings they can revert. The recommended settings could also be adapted over time and the caller should not care.

I had the NixOS option [services.nginx.recommendedTlsSettings](https://search.nixos.org/options?channel=25.11&query=services.nginx.recommended) in mind where this assumption is more or less true. For example, new ciphers are added and insecure ones are removed over time and the caller does not care (too much).

But you are right that this assumption does not necessarily hold for this function because it is highly breaking if settings are changed. So, it cannot be a black box for the caller.

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: During daily today he said "Only set runAsNonRoot", nothing else is needed for now

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: I updated the PR to only contain runAsNonRoot for now (as @lfrancke told us).
That should be fine for all operators I hope. In case some operator for *some* reasons needs to run as root, it can still explicitly enable it.

As requested I removed the `stackable_default_pod_security_context` function and used builder methods.
I think with this the PR is very minimal.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
...
 
+### Changed
+
+- BREAKING: `PodSecurityContextBuilder::new` was removed in favor of `PodSecurityContextBuilder::with_stackable_defaults` ([#XXXX]).
+  This function already sets up some defaults we want to use across the platform.
+  Currently this is `runAsNonRoot: true`, which might cause product Pods to crash and require changes.
+- BREAKING: `PodSecurityContextBuilder::run_as_non_root` now takes a `bool` instead of assuming consumers always want to set it to `true` ([#XXXX]).
+  This is needed to allow users setting it to `false` in case the new `with_stackable_defaults` functions set's it to `true`.
+
+[#XXXX]: https://github.com/stackabletech/operator-rs/pull/XXXX
```
siegfriedweber:
```suggestion
- BREAKING: `PodSecurityContextBuilder::new` was removed in favor of `PodSecurityContextBuilder::with_stackable_defaults` ([#1205]).
  This function already sets up some defaults we want to use across the platform.
  Currently this is `runAsNonRoot: true`, which might cause product Pods to crash and require changes.
- BREAKING: `PodSecurityContextBuilder::run_as_non_root` now takes a `bool` instead of assuming consumers always want to set it to `true` ([#1205]).
  This is needed to allow users setting it to `false` in case the new `with_stackable_defaults` function sets it to `true`.

[#1205]: https://github.com/stackabletech/operator-rs/pull/1205
```

crates/stackable-operator/src/builder/pod/security.rs
siegfriedweber: If needed, we can add it in a later pull request.

siegfriedweber (changes requested):

crates/stackable-operator/CHANGELOG.md
sbernauer: 77291e73e060ad3e4ea76a783c9cf05f9f6721da and f3ce81f349062cd51910a0bab23e7923e9a74b0b

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: I had to drop off for lunch, thus hadn't commited it yet. But I now pushed it as 77291e73e060ad3e4ea76a783c9cf05f9f6721da, feel free to have a look!

crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -10,16 +10,29 @@ pub struct SecurityContextBuilder {
...
+    ///
+    /// * `runAsNonRoot: true`
+    pub fn with_stackable_defaults() -> Self {
+        // We are using the builder functions to ensure that builder functions exist to override these settings.
+        let mut builder = Self {
+            security_context: SecurityContext::default(),
+        };
+
+        // Reason: Running as root is bad
+        builder.run_as_non_root(true);
```
siegfriedweber: It should default to `None`, i.e. it should not be set here. It does not make sense (and it is confusing) to set it in both the Pod and the Container.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
...
 
+### Changed
+
+- BREAKING: `PodSecurityContextBuilder::new` was removed in favor of `PodSecurityContextBuilder::with_stackable_defaults`
+  (same for `SecurityContextBuilder`) ([#1205]).
+  This function already sets up some defaults we want to use across the platform.
+  Currently this is `runAsNonRoot: true`, which might cause product Pods to crash and require changes.
+- BREAKING: `PodSecurityContextBuilder::run_as_non_root` now takes a `bool` instead of assuming consumers always want to set it to `true` ([#1205]).
+  This is needed to allow users setting it to `false` in case the new `with_stackable_defaults` function sets it to `true`.
+- BREAKING: `SecurityContextBuilder::run_as_root` is now called on the builder, not as a builder-creation function ([#1205]).
```
siegfriedweber: `SecurityContextBuilder::run_as_root` did not create the builder. It created a `SecurityContext`.

crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -10,16 +10,29 @@ pub struct SecurityContextBuilder {
...
+
+        builder
     }
 
-    pub fn new() -> Self {
-        Self::default()
+    /// Convenience function for a wide use-case.
+    ///
+    /// Please only use this is really needed.
+    pub fn run_as_root(&mut self) {
```
siegfriedweber: This is not a proper builder function. It cannot be chained. Despite its name, it does not invert `run_as_non_root`.

And it cannot be used anyway because this builder lacks a `build` function.

I would just remove this function. It is only used in stackablectl: https://github.com/stackabletech/stackablectl/blob/stackablectl-1.4.0/rust/stackablectl/src/cmds/debug.rs#L177

crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -10,16 +10,29 @@ pub struct SecurityContextBuilder {
...
-        SecurityContext {
-            run_as_user: Some(0),
-            ..SecurityContext::default()
-        }
+    /// Construct a new [`SecurityContextBuilder`] that is pre-filled with Stackable's defaults.
+    ///
+    /// Currently the defaults are:
+    ///
+    /// * `runAsNonRoot: true`
+    pub fn with_stackable_defaults() -> Self {
```
siegfriedweber, line 20: `SecurityContextBuilder` already implements `Default` without the Stackable defaults. It is not obvious which "default" applies where. And the default in the Container spec is `None` anyway, so the settings here usually do not have any effect.

siegfriedweber (changes requested):

crates/stackable-operator/CHANGELOG.md
sbernauer: Removed in 3bd1b69

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: Good point! Removed in 3bd1b69

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: Oh yeah, the default needs to go: 19b939e

crates/stackable-operator/src/builder/pod/security.rs
sbernauer: The question is: Can we assume that we at Stackable always use a PodBuilder and a ContainerBuilder?
Maybe things have changed with smooth operator?
Defaulting in both ensure we apply the defaults as soon as any of the builders are being used.

siegfriedweber: My approach wouldn't be to try to prevent every possible misuse of the API. There will always be loopholes. For example, the OpenSearch operator uses neither the `PodBuilder` nor the
`ContainerBuilder`. Since we control the code of our operators and have every intention of implementing them properly, I would instead describe the ideal state and implement that. In my opinion, the ideal state that satisfies the requirements described in this ticket would be the following:

```yaml
apiVersion: v1
kind: Pod
spec:
  securityContext:
    runAsNonRoot: true
  containers:
    - securityContext: {}
```

As part of the smooth-operator refactoring, we can ensure they are used correctly. The OpenSearch operator will also use these builders once they have been refactored.

sbernauer: Looks like you have a stronger opinion ;) bee890e76dd041153d98cf70d51a832589346fb0

siegfriedweber (approved):
LGTM

PR #1207, opened 2026-05-26
feat(operator/client): Support feature gate retrieval

Techassi:
This came up in our last versions meeting. We liked the idea of being able to check which features are enabled and on which Kubernetes version the operator is running to be able to potentially use newer Kubernetes features if available.

This PR adds associated functions to the `Client` to retrieve (enabled/disabled/all) feature gates via the `/metrics` endpoint of the Kubernetes apiserver. Each feature gate needs to be parsed from a line-based metric into a Rust struct. Parsing is achieved with the help of `winnow` (new dependency).

### DIFF
--- a/crates/stackable-operator/src/client/feature_gates.rs
+++ b/crates/stackable-operator/src/client/feature_gates.rs
@@ -30,12 +30,7 @@ impl Client {
             .await
             .context(PerformRawRequestSnafu)?;
 
-        response
-            .lines()
-            .filter(|l| l.starts_with("kubernetes_feature_enabled"))
-            .map(FeatureGate::from_str)
-            .collect::<Result<Vec<FeatureGate>, _>>()
-            .map_err(|error| ParseFeatureGateSnafu { error }.build())
+        FeatureGate::parse_from_metrics(&response)
     }
 
     /// Retrieves enabled feature gates.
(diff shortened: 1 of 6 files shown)


crates/stackable-operator/src/client/feature_gates.rs
```diff
@@ -0,0 +1,229 @@
...
+            .await
+            .context(PerformRawRequestSnafu)?;
+
+        response
+            .lines()
+            .filter(|l| l.starts_with("kubernetes_feature_enabled"))
+            .map(FeatureGate::from_str)
+            .collect::<Result<Vec<FeatureGate>, _>>()
+            .map_err(|error| ParseFeatureGateSnafu { error }.build())
+    }
```
NickLarsenNZ: I think this should be moved out so it can be tested offline with a static string

Techassi: I hope this is resolved by 43902423.

NickLarsenNZ: I'd also expect the raw string. This serves as documentation so the reader can understand what the endpoint even serves, to connect that with what we are parsing.

```
# usual prometheus style comments
kubernetes_feature_enabled{name="APIResponseCompression",stage="BETA"} 1
kubernetes_feature_enabled{name="APIServingWithRoutine",stage="ALPHA"} 0

# space above
other_metric{} 100
```

Techassi: I _guess_ I could move out the filtering and make it part of the parsing itself, but that filter is just an easy way to get rid of all unrelated lines.

I can leave a dev comment explaining the raw input tho.

NickLarsenNZ: A dev comment would suffice. But I think the test would be better (even in place of the rstests). To show what we can parse, and that we can handle unexpected stuff.

It was also the first thing I looked for so I could even understand what this does instead of reading logic not knowing what it is for.

Techassi: Take a look at b2264de6.

NickLarsenNZ: Thanks

NickLarsenNZ (approved):
LGTM

PR #1208, opened 2026-05-27
feat(operator): Add client-feature-gates gate

Techassi:
This PR adds a feature gate for changes from #1207.

Users now have to explicitly opt-in to use the new associated functions on `Client` as well as the `FeatureGate` and `FeatureStage` types.

### DIFF
--- a/crates/stackable-operator/Cargo.toml
+++ b/crates/stackable-operator/Cargo.toml
@@ -9,8 +9,9 @@ repository.workspace = true
 
 [features]
 default = ["crds"]
-full = ["crds", "certs", "time", "webhook", "kube-ws"]
+full = ["client-feature-gates", "crds", "certs", "time", "webhook", "kube-ws"]
 
+client-feature-gates = ["dep:winnow"]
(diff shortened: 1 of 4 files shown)


crates/stackable-operator/Cargo.toml
```diff
@@ -9,8 +9,9 @@ repository.workspace = true
 
 [features]
 default = ["crds"]
-full = ["crds", "certs", "time", "webhook", "kube-ws"]
+full = ["client-feature-gates", "crds", "certs", "time", "webhook", "kube-ws"]
```
NickLarsenNZ, line 12: I might have named it with something containing kube. Up to you though.

NickLarsenNZ: Eg: k8s-client-feature-gates

NickLarsenNZ (approved):
LGTM

crates/stackable-operator/Cargo.toml
Techassi: We discussed the name and decided to move forward as is.

PR #1212, opened 2026-05-29
Fix/make role and group utils pub

maltesander:
# Description

- more public v2 framework stuff adapted

### Author

- [ ] CRD changes approved

### Reviewer

- [x] Test cases added
- [ ] Changelog updated: Can be postponed until the branch `smooth-operator` is merged into `main`

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/v2/builder/pod/container.rs
+++ b/crates/stackable-operator/src/v2/builder/pod/container.rs
@@ -5,13 +5,13 @@ use std::{
 };
 
 use snafu::Snafu;
-use stackable_operator::{
-    builder::pod::container::{ContainerBuilder, FieldPathEnvVar},
-    k8s_openapi::api::core::v1::{ConfigMapKeySelector, EnvVar, EnvVarSource, ObjectFieldSelector},
-};
 use strum::{EnumDiscriminants, IntoStaticStr};
 
-use crate::v2::types::kubernetes::{ConfigMapKey, ConfigMapName, ContainerName};
+use crate::{
+    builder::pod::container::{ContainerBuilder, FieldPathEnvVar},
+    k8s_openapi::api::core::v1::{ConfigMapKeySelector, EnvVar, EnvVarSource, ObjectFieldSelector},
+    v2::types::kubernetes::{ConfigMapKey, ConfigMapName, ContainerName},
+};
 
 #[derive(Snafu, Debug, EnumDiscriminants)]
 #[strum_discriminants(derive(IntoStaticStr))]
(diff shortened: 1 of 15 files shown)


crates/stackable-operator/src/v2/mod.rs
```diff
@@ -1 +1,27 @@
...
+pub mod role_group_utils;
+pub mod role_utils;
+pub mod types;
+
+/// Has a non-empty name
+///
+/// Useful as an object reference; Should not be used to create an object because the name could
+/// violate the naming constraints (e.g. maximum length) of the object.
+pub trait HasName {
+    #[allow(dead_code)]
```
siegfriedweber, line 15: Is this directive still required?

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -124,7 +124,7 @@ impl EnvVarSet {
     /// Adds an environment variable with the given name and field path to this set
     ///
     /// An [`EnvVar`] with the same name is overridden.
-    pub fn with_field_path(mut self, name: &EnvVarName, field_path: FieldPathEnvVar) -> Self {
+    pub fn with_field_path(mut self, name: &EnvVarName, field_path: &FieldPathEnvVar) -> Self {
```
siegfriedweber, line 127: Is there a reason for this change? The usability is slightly better without the reference.

siegfriedweber (changes requested):
Which change makes the role and role group utils public?

There are build errors (on my machine):

```
error[E0433]: cannot find `stackable_operator` in the crate root
   --> crates/stackable-operator/src/v2/role_utils.rs:222:21
    |
222 |     #[derive(Debug, Fragment, PartialEq)]
    |                     ^^^^^^^^ could not find `stackable_operator` in the list of imported crates
    |
    = note: this error originates in the derive macro `Fragment` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this module through its public re-export
    |
200 +     use crate::v2::role_utils::fragment;
    |
```

crates/stackable-operator/src/v2/mod.rs
maltesander: removed.

crates/stackable-operator/src/v2/builder/pod/container.rs
maltesander: That is clippy, it is not consumed in that function, so it can be a reference.

siegfriedweber (approved):

PR #1222, opened 2026-06-12
Relax the length constraints of cluster, role and role group name

siegfriedweber:
# Description

Relax the length constraints of cluster, role and role group name

Some operators use long role names like `dagprocessor` and `middlemanager`. Increasing the length of role names would require decreasing the length of cluster or role group name. These length constraints are already so tight that they could break existing cluster configurations. Instead, the length constraints were relaxed: `ClusterName` is now up to 40 characters, `RoleName` and `RoleGroupName` are up to 63 characters each, according to RFC 1123 label names. In turn, if the qualified role group name length exceeds its limit of 52 characters, it is truncated and a hash is added.

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/v2/role_group_utils.rs
+++ b/crates/stackable-operator/src/v2/role_group_utils.rs
@@ -77,6 +77,7 @@ impl ResourceNames {
             "{}-{}-{}",
             self.cluster_name, self.role_name, self.role_group_name,
         );
+        // `concatenated_name` contains only ASCII characters.
         let sanitized_name = Self::ensure_max_length(
             concatenated_name,
             QualifiedRoleGroupName::MAX_LENGTH,
@@ -92,8 +93,12 @@ impl ResourceNames {
     ///
     /// # Panics
     ///
-    /// Panics if `max_length < 1 /* character */ + 1 /* dash */ + hash_length`.
+    /// Panics if `resource_name` contains non-ASCII characters or if
+    /// `max_length < 1 /* character */ + 1 /* dash */ + hash_length`.
+    ///
+    /// Kubernetes object names cannot contain non-ASCII characters.
     fn ensure_max_length(resource_name: String, max_length: usize, hash_length: usize) -> String {
+        assert!(resource_name.is_ascii());
         assert!(max_length >= 1 /* character */ + 1 /* dash */ + hash_length);
 
         if resource_name.len() <= max_length {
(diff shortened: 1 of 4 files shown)


crates/stackable-operator/src/v2/role_group_utils.rs
```diff
@@ -47,11 +73,59 @@ impl ResourceNames {
...
+    /// Panics if `max_length < 1 /* character */ + 1 /* dash */ + hash_length`.
+    fn ensure_max_length(resource_name: String, max_length: usize, hash_length: usize) -> String {
+        assert!(max_length >= 1 /* character */ + 1 /* dash */ + hash_length);
+
+        if resource_name.len() <= max_length {
+            resource_name
+        } else if hash_length == 0 {
+            let mut truncated_name = resource_name;
+            truncated_name.truncate(max_length);
+            truncated_name
```
maltesander, line 109: The `hash_length` is a constant and never 0?

maltesander (commented):
LGTM, there are some stale compile asserts not using the `qualified_role_group_name` assert "The string `<cluster_name>-<hash>`..." but "The string `<cluster_name>-<role_name>-<role_group_name>`...":
- role_group_config_map
- stateful_set_name
- listener_name

adwk67 (commented):
Testing with superset/resources test that failed with current smooth-op branch.
Nit: does this change behaviour - e.g. where previously a name exceeding the allowed overall length would result in an error, the name is now silently adapted? If so, we should add something to the change log, along with the other breaking changes.

crates/stackable-operator/src/v2/role_group_utils.rs
siegfriedweber: I wanted the function to be self-contained. It could be called from somewhere else with different parameters and if `hash_length` is set to 0, it would create invalid Kubernetes object names if I do not handle this case.

siegfriedweber:
> Nit: does this change behaviour - e.g. where previously a name exceeding the allowed overall length would result in an error, the name is now silently adapted? If so, we should add something to the change log, along with the other breaking changes.

If a name exceeded the length limit, then the cluster spec was not rejected by the operator. It just applied the resources until Kubernetes complained. Usually, some Services and ConfigMaps were applied but the StatefulSet was not. The operator then went into an error loop.

This is in my opinion not a change but a new feature. The operator can now handle more specs than before.

adwk67 (approved):
Local failing tests now pass.

maltesander (approved):
LGTM!

PR #1226, opened 2026-06-18
feat: Support floating tag in product image selection

Techassi:
Part of https://github.com/stackabletech/internal-issues/issues/169.

This PR adds support for auto updating product images. This is done by using a floating tag for product images instead of fully-qualified ones.

Significant changes are:

- Add a new `stackableVersionStrategy` field to the product image selection CRD interface:

  ```yaml
  productImage:
    stackableVersionStrategy:
      default: Exact
      description: |-
        Configure the Stackable version strategy. Defaults to `Exact`.

        Currently, two variants are supported:

        - `Exact`, which uses the exact, fully-qualified, canonical version of a product image.
        - `LatestPatch`, referencing a floating tag which always points to the latest patch version
          in the current release line. The current release line is either automatically derived by
          the operator based on its own version, or can be overridden with `stackableVersion`.

        A potential newer image is only pulled when Pods are rotated or their containers are
        restarted. Pods are NOT rotated and containers are NOT restarted automatically when a new
        image is available. This behaviour makes this a passive update mechanism, rather than an
        active one.

        It should be noted that when this field is set to `LatestPatch`, the operator automatically
        uses `Always` as the pull policy for product images. If set to `Exact`, `IfNotPresent` is
        used. Explicitly setting `pullPolicy` takes precedence.

        ### Examples

        - The `stackableVersion` field is not set, the operator falls back to its own version, eg.
          26.7.0. If this field is set to `LatestPatch`, the `26.7` floating tag will be used for
          product images, else, `26.7.0` will be used.
        - The `stackableVersion` field is set to `26.3.0`. If this field is set to `LatestPatch`,
          the `26.3` floating tag will be used for product images, else, `26.3.0` will be used.
      enum:
      - Exact
      - LatestPatch
      type: string
  ```

- Remove the default of `pullPolicy`. This better aligns with the floating tag feature, because if enabled, the pull policy is automatically set to `Always`. The current default is (as far as I can see) a leftover from the past. Even the doc comment states that `Always` is the default, as we support floating tags (not true).

- The `stackableVersion` field is now a proper semantic version, instead of a plain string. This is needed, to correctly extract the floating tag if the auto update is enabled.

[Release notes](https://github.com/stackabletech/operator-rs/pull/1226#issuecomment-5603667429)


crates/stackable-operator/crds/DummyCluster.yaml
```diff
@@ -1872,9 +1894,9 @@ spec:
                     type: string
                   stackableVersion:
                     description: |-
-                      Stackable version of the product, e.g. `23.4`, `23.4.1` or `0.0.0-dev`.
+                      Stackable version of the product, e.g. `26.7.0` or `0.0.0-dev`.
 
-                      If not specified, the operator will use its own version, e.g. `23.4.1`. When using a nightly
+                      If not specified, the operator will use its own version, e.g. `26.7.1`. When using a nightly
```
sbernauer, line 1892: We should mention the new bool here I'd say

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -98,18 +119,28 @@ pub struct ResolvedProductImage {
...
-/// We default to `Always`, as we use floating tags for our release lines.
-/// This means the tag 23.4 starts of pointing to the same image 23.4.0 does, but switches to 23.4.1 after the releases of 23.4.1.
+/// TODO: Update comment
 ///
 /// ### See
 ///
 /// - <https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy>
 /// - <https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2291-L2300>
+#[derive(
+    Clone,
```
sbernauer, line 141: nit: That way we can avoid a clone (clippy will tell you)
```suggestion
    Copy,
    Clone,
```

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -98,18 +119,28 @@ pub struct ResolvedProductImage {
...
+    Debug,
+    Default,
```
sbernauer: I think I would prefer to not derive Default for this, as there are so many cases and it's complicated. There is no "fits it all" default. A unwrap_or_default() looks so innocent, but should IMHO should be an explicit decision

sbernauer:
A few points on the CRD change:

* I like it in general, it's pretty simple to configure and understand for users.
* Naming things is hard :sweat:. I dislike `autoUpdate` because it's sounds like we are actively doing stuff and I would prefer it to be a bit more speaking. Ideas could we `patchLevelUpdates` or even more descritptive `patchLevelFloatingTag` - which makes it clear we are not actually updating anything ourselves, but just point at a floating tag. We might also to add a "real" update mechanism in the future (periodically checking for new versions and restarting), which could than get the autoUpdate (or whatever) name.
* We should clearly document that this will not automatically pull in updates, but only on Pod restarts.
** Also mention TLS cert rotation fixes thís implictely.

Comments on the implementations:

* For 0.0.0-dev images IMHO we want to default pullPolicy Always, as it's a giant floating tag.
** I had to revert a few test asserts Always -> IfNotDefault changes
* When a ":latest" image is used it's also a floating tag, so we should default to Always as well. That's what k8s seems to do as well.
** I also had to revert custom_without_tag and custom_with_colon_in_repo_and_without_tag test changes
* We should also warn in case the user configures a floating tag but a "wrong" pull policy.

I updated the tests and the refactored the implementation to turn them green again (one is left as an exercise for the reader). I also added some explanatory comments.
I send you the patch via Slack. Happy about comments!

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Good spot! Addressed in 83531c76.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Yeah, true. This is "double breaking" then: The default is no longer `Always` + there is now _no_ `Default` impl. But I agree that this is the right direction.

Addressed in 83531c76.

Techassi:
> Naming things is hard

Agreed. Like discussed, I now went with `useFloatingTag` instead.

> We should clearly document that this will not automatically pull in updates, but only on Pod restarts.

Yes, I think I sufficiently touched on this in my updated doc comment.

> Also mention TLS cert rotation fixes thís implictely.

I would say that this is implicitly covered by my current doc comment. It basically says, that a container/pod needs to restart in order to have an effect.

All your comments on the implementation should be addressed in 83531c76.

crates/stackable-operator/crds/DummyCluster.yaml
Techassi: Could you explain what you want to see here? I think the documentation on `useFloatingTag` is rather complete and mentions how `stackableVersion` is used.

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -98,22 +124,42 @@ pub struct ResolvedProductImage {
...
 }
 
+impl PullPolicy {
+    /// Returns the appropriate [`PullPolicy`] based on if a floating tag is used.
+    fn from_is_floating_tag(is_floating_tag: bool) -> Self {
+        if is_floating_tag {
+            Self::Always
+        } else {
+            Self::IfNotPresent
+        }
```
sbernauer, line 167:
```suggestion
        if is_floating_tag {
            // Floating tags only really make sense in case new versions are pulled on startup.
            Self::Always
        } else {
            // Fixed (e.g. full SDP version) tags should never move and thus only need to be pulled if not present.
            Self::IfNotPresent
        }
```

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -137,41 +183,52 @@ impl ProductImage {
...
+                    app_version_label_value,
                     image_pull_policy,
-                    pull_secrets,
                 })
             }
             ProductImageSelection::Auto(AutoProductImage {
-                product_version,
                 stackable_version,
+                product_version,
+                use_floating_tag,
```
sbernauer: Any reason for the re-ordering (also in other places)?

crates/stackable-operator/crds/DummyCluster.yaml
sbernauer: Maybe something like
> If not specified, the operator will use its own version, e.g. `26.7.1` (or the `26.7` floating tag in case useFloatingTag is set to true).
> When using a nightly operator or a PR version, it will use the nightly `0.0.0-dev` image.

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -297,7 +383,7 @@ mod tests {
             image: "oci.stackable.tech/sdp/superset:1.4.1-stackable0.0.0-dev".to_string(),
             app_version_label_value: "1.4.1-stackable0.0.0-dev".parse().expect("static app version label is always valid"),
             product_version: "1.4.1".to_string(),
-            image_pull_policy: "Always".to_string(),
+            image_pull_policy: "IfNotPresent".to_string(),
```
sbernauer: IMHO even if the user specified 0.0.0-dev it's a floating tag after all:
```diff
diff --git a/crates/stackable-operator/src/commons/product_image_selection.rs b/crates/stackable-operator/src/commons/product_image_selection.rs
index fafdcab..e8a481f 100644
--- a/crates/stackable-operator/src/commons/product_image_selection.rs
+++ b/crates/stackable-operator/src/commons/product_image_selection.rs
@@ -253,7 +253,6 @@ impl ProductImage {
                                 version = *ZERO_ZERO_ZERO_DEV
                             );
 
-                            is_floating_tag = true;
                             &*ZERO_ZERO_ZERO_DEV
                         } else {
                             operator_version
@@ -261,6 +260,11 @@ impl ProductImage {
                     }
                 };
 
+                // 0.0.0-dev is one giant floating tag
+                if stackable_version == &*ZERO_ZERO_ZERO_DEV {
+                    is_floating_tag = true;
+                }
+
                 let stackable_version = if *use_floating_tag {
                     is_floating_tag = true;
                     stackable_version.floating()
@@ -383,7 +387,7 @@ mod tests {
             image: "oci.stackable.tech/sdp/superset:1.4.1-stackable0.0.0-dev".to_string(),
             app_version_label_value: "1.4.1-stackable0.0.0-dev".parse().expect("static app version label is always valid"),
             product_version: "1.4.1".to_string(),
-            image_pull_policy: "IfNotPresent".to_string(),
+            image_pull_policy: "Always".to_string(),
             pull_secrets: None,
         }
     )]
```

sbernauer (commented):
I have to admin I prefer `patchLevelFloatingTag` as it makes it a bot more clear what the floatingTag is about

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Addressed in c0c90316.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: No not really, other than my OCD :see_no_evil:

I can revert if needed.

crates/stackable-operator/crds/DummyCluster.yaml
Techassi: Done in f760b36a.

crates/stackable-operator/src/commons/product_image_selection.rs
sbernauer: IIRC there even is a clippy lint that when constructing a struct you should use the order of the fields as it's defined in the struct.
That has product_version, stackable_version, use_floating_tag and repo, which I like :)

crates/stackable-operator/crds/DummyCluster.yaml
```diff
@@ -1887,12 +1888,41 @@ spec:
...
                       operator or a PR version, it will use the nightly `0.0.0-dev` image.
+
+                      If this is used in combination with `useFloatingTag`, the correct floating tag is computed
+                      automatically, e.g. `26.7.0` becomes `26.7` for the product image.
                     nullable: true
                     type: string
+                  useFloatingTag:
```
sbernauer: Bumping this up so we don't forget

> I have to admin I prefer `patchLevelFloatingTag` as it makes it a bot more clear what the floatingTag is about.
Also we might want to have a different floating tag for some other functionality in the future (making things up: Only baseOS updates or product version bugfix bumps), in which case the abbreviated field name causes problems (which we don't have when we express **what** floating tag the user configures).

sbernauer:
Bumping the [bump](https://github.com/stackabletech/operator-rs/pull/1226#discussion_r3512602114) as I suggest to rename it before voting:

> I have to admin I prefer `patchLevelFloatingTag` as it makes it a bot more clear what the floatingTag is about.
Also we might want to have a different floating tag for some other functionality in the future (making things up: Only baseOS updates or product version bugfix bumps), in which case the abbreviated field name causes problems (which we don't have when we express **what** floating tag the user configures).

Techassi:
> Bumping the [bump](https://github.com/stackabletech/operator-rs/pull/1226#discussion_r3512602114) as I suggest to rename it before voting:
> 
> > I have to admin I prefer `patchLevelFloatingTag` as it makes it a bot more clear what the floatingTag is about

I did not forgot about this, but _personally_ I'm fine with the current name. I hope that others voting on this will prefer one or the other, otherwise it is 1 vs 1.

Techassi:
@soenkeliebau mentioned that we might want to remove the term "floating tag" from the field name so that we don't explicitly mention the underlying mechanism (which could change down he line).

I agree that it makes sense, but naming things is hard. A few alternative names might be:

- `useLatestPatchVersion`
- `autoUpdatePatchVersion`
- `useRollingVersion`

soenkeliebau:
Late to the party, sorry. @Techassi already represented my concerns above. 

I have since considered this a bit more and can also see a problem with patchLevel since this could be confusing whether it refers to the product version or the stackable version as it sits at the same level in the CRD I believe.

CRD could look like this:

```
image:
  productVersion: 1.2.3
  useLatestPatchVersion: true
```
Which looks like it makes a lot of sense, but the parameters refer to entirely different things, even though they sound totally similar.

Also, K8s tends towards enums rather than bools, its even [specified in their docs](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#primitive-types) :) 


> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

If I may propose a totally new approach (sorry!), how about `stackableVersionPolicy` with two initial allowed values:
- `exact`
- `latestPatch`

and the default being `exact`?

That would give us this:
```
image:
  productVersion: 1.2.3
  stackableVersionPolicy: latestPatch
```

Maybe even `stackableImageVersionPolicy` but at some point it gets long ...

Techassi:
Oh I like the idea with the enum. Two things:

- I like `stackableVersionPolicy` as the field name. Using `stackableImageVersionPolicy` is in my opinion overly verbose and also doubles information (it is already implied we are talking about the `image` by the parent key.
- I like both variants, however per Kubernetes conventions, they should use PascalCase:
  - `Exact`
  - `LatestPatch`

soenkeliebau:
Indeed, sorry. I am simply unable to keep it straight in my head which case is to be used where :)

Edit to mention a random idea that just came to me:
Maybe we can even extend the CRD with a `productVersionPolicy` at some point..

Techassi:
Okay, today we decided to go ahead with @soenkeliebau's enum proposal. The exact structure will look like this:

```yaml
image:
  stackableVersionPolicy: Exact # or LatestPatch
```

~~The PR description will be updated accordingly as well.~~ **Done.**

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -66,11 +66,43 @@ pub struct AutoProductImage {
...
+    ///
+    /// ### Examples
+    ///
+    /// - The `stackableVersion` field is not set, the operator falls back to its own version, eg.
+    ///   26.7.0. If this field is set to `LatestPatch`, the `26.7` floating tag will be used for
+    ///   product images, else, `26.7.0` will be used.
+    /// - The `stackableVersion` field is set to `26.3.0`. If this field is set to `LatestPatch`,
+    ///   the `26.3` floating tag will be used for product images, else, `26.3.0` will be used.
+    #[serde(default)]
+    stackable_version_strategy: StackableVersionStrategy,
```
adwk67: This is `stackableVersionPolicy` in the PR notes (https://github.com/stackabletech/operator-rs/pull/1226#issuecomment-5507040356) or has that been changed?

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -66,11 +66,43 @@ pub struct AutoProductImage {
...
 
-    /// Stackable version of the product, e.g. `23.4`, `23.4.1` or `0.0.0-dev`.
+    /// Stackable version of the product, e.g. `26.7.0` or `0.0.0-dev`.
     ///
-    /// If not specified, the operator will use its own version, e.g. `23.4.1`. When using a nightly
+    /// If not specified, the operator will use its own version, e.g. `26.7.1`. When using a nightly
     /// operator or a PR version, it will use the nightly `0.0.0-dev` image.
-    stackable_version: Option<String>,
+    ///
+    /// If this is used in combination with `useFloatingTag`, the correct floating tag is computed
```
adwk67: `useFloatingTag` is stale

crates/stackable-shared/src/semver.rs
```diff
@@ -0,0 +1,56 @@
...
+pub static ZERO_ZERO_ZERO_DEV: LazyLock<semver::Version> = LazyLock::new(|| semver::Version {
+    major: 0,
+    minor: 0,
+    patch: 0,
+    pre: semver::Prerelease::new("dev").expect("static prerelease must be valid"),
+    build: semver::BuildMetadata::EMPTY,
+});
+
+pub trait VersionExt {
+    /// Returns the floating version as a [`String`], eg. `26.7.0` -> `26.7`
```
adwk67:
```suggestion
    /// Returns the floating version as a [`String`], e.g. `26.7.0` -> `26.7`.
```

crates/stackable-shared/src/semver.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// Currently, `0.0.0-dev` and `0.0.0-pr*` versions are considered floating.
+    fn is_floating(&self) -> bool;
+}
+
+impl VersionExt for ::semver::Version {
+    fn floating(&self) -> String {
+        if self.is_floating() {
+            self.to_string()
+        } else {
+            format!("{major}.{minor}", major = self.major, minor = self.minor)
```
adwk67: I think we should at least document what happens to pre-release versions, as 26.7.0-rc1 will be stripped to 26.7. This _feels_ like this could be a problem, but I'm not 100% sure.

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -66,11 +66,43 @@ pub struct AutoProductImage {
...
+    /// automatically, e.g. `26.7.0` becomes `26.7` for the product image.
```
adwk67: This is breaking, as what was previously advertised as valid (e.g. 26.7) is not any longer. What happens with existing CRs that use yy.mm - it will be accepted here but won't deserialization fail in the operator?

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -98,22 +130,44 @@ pub struct ResolvedProductImage {
     pub pull_secrets: Option<Vec<LocalObjectReference>>,
 }
 
-#[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize, AsRefStr)]
-#[serde(rename = "PascalCase")]
-/// We default to `Always`, as we use floating tags for our release lines.
-/// This means the tag 23.4 starts of pointing to the same image 23.4.0 does, but switches to 23.4.1 after the releases of 23.4.1.
+/// TODO: Update comment
```
adwk67: Something like: "The default value is now derived from whether or not the tag is floating"?

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -182,23 +247,54 @@ impl ProductImage {
...
+                    stackable_version.to_string()
+                };
+
+                let image_pull_policy = match pull_policy {
+                    Some(pull_policy) => {
+                        if is_floating_tag && *pull_policy != PullPolicy::Always {
+                            tracing::warn!(
+                                pull_policy.configured = %pull_policy,
+                                stackable_version,
+                                r#"product image pull policy is not "Always" but a floating tag is \
```
adwk67: This is a raw string: will the back-slashes still get processed?

crates/stackable-operator/src/commons/product_image_selection.rs
```diff
@@ -137,41 +191,52 @@ impl ProductImage {
...
             ProductImageSelection::Custom(CustomProductImage {
                 custom,
                 product_version,
             }) => {
                 let image_ref = ImageRef::parse(custom);
-                let image_tag_or_hash = image_ref
-                    .tag
-                    .or(image_ref.hash)
-                    .unwrap_or_else(|| "latest".to_string());
+                let image_tag_or_hash = image_ref.tag.or(image_ref.hash).unwrap_or_else(|| {
```
adwk67: This will result in `:latest` as being non-floating which contradicts the comment above in lines 207-208.

adwk67 (commented):
Mainly small things. I'm not sure about the implicit breaking behaviour, though: is that going to bite us? (e.g. existing CRs with e.g. 26.7)

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: My mind must have replaced "Policy" with "Strategy". I'm honestly indifferent about it. What do you think?

CC'ing @soenkeliebau

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Ah yes, good spot. I will update.

crates/stackable-shared/src/semver.rs
Techassi: Yes, you are correct. I actually fixed this for boil recently. See https://github.com/stackabletech/docker-images/pull/1601 and https://github.com/stackabletech/docker-images/pull/1606.

I will fix it here as well.

I will also raise a follow-up PR to move this into it's own crate which encodes our CalVer based versioning so that we can use it across our Rust codebase.

crates/stackable-operator/src/commons/product_image_selection.rs
adwk67: I'm fine with "Strategy"

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Previously, everything was valid. Which also means that users can already produce invalid tags (which then lead to image pull errors).

As we currently don't tag our images with YY.MM (only with YY.MM.X), this was previously possible, but would have led to image pull errors. So I would say this is fine as is and we now do correct thing.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Yes, thank you for the reminder. It will indeed be something along those lines.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Good catch, I _think_ they won't get processed (which means this would need updating). I will do some testing.

crates/stackable-operator/src/commons/product_image_selection.rs
adwk67: Ah, yes, thanks for the clarification. :+1:

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: > This will result in `:latest `as being non-floating

Am I missing something here, because I don't see it? In the `unwrap_or_else` combinator, we set `is_floating_tag = true`, which in turn leads to the correct `pullPolicy` selection further down.

However, one issue I just spotted is that we should print the same warning here as below. I will pull out that particular piece of code into its own function.

adwk67: Ah, maybe I mis-read that. I assumed that `image_ref.tag` would pick up `:latest` as the tag and we would never reach the `unwrap_or_else` bit.

Techassi: The behaviour to before is the same here. If users specify a completely `custom` image, but without a ref, we default to `:latest`. And `:latest` is considered floating by us.

adwk67: And so _explicitly_ stating e.g. `my-custom-image:latest` will result the image as being treated as non-floating?

crates/stackable-operator/src/commons/product_image_selection.rs
soenkeliebau: TL/DR: I think it should be `policy`

I did a little bit of research and reading and while there doesn't seem to be a hard rule, the established pattern seems to kind of be this: 

> Policy in Kubernetes is used for a flat enum that answers whether/when something is allowed or happens, Strategy is used for how an operation is carried out, and it's almost always a struct with a type discriminator plus type-specific parameters.

This is from Claude, I tried finding a hard quote to prove this, but was unable to find something stating it explicitly. 
https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#primitive-types calls it "policy options" in the last bullet of the list, but that is kinda implicit.

Flux uses `imagePolicy` for something similar: https://fluxcd.io/flux/components/image/imagepolicies/

And looking some examples from k8s core: 

 Policy:
- [imagePullPolicy](https://kubernetes.io/docs/reference/kubernetes-api/core/pod-v1/#Container)
- [restartPolicy](https://kubernetes.io/docs/reference/kubernetes-api/core/pod-v1/#PodSpec)
- [concurrencyPolicy](https://kubernetes.io/docs/reference/kubernetes-api/batch/cron-job-v1/#CronJobSpec)

Strategy:
- [strategy](https://kubernetes.io/docs/reference/kubernetes-api/apps/deployment-v1/#DeploymentStrategy)
- [updateStrategy](https://kubernetes.io/docs/reference/kubernetes-api/apps/stateful-set-v1/#StatefulSetUpdateStrategy)

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Updated in 87100f77.

crates/stackable-shared/src/semver.rs
Techassi: Fixed in e903648c.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Updated comment added in 9292a887.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Adjusted in 728377ce.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Ah, good catch. I think that's currently the case indeed, which is wrong. I will fix this and add a unit test for this.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: I will change it to "Policy" and will leave a dev comment referring to some of these materials.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Fixed in 61671b6d.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Renamed in fdd1b821.

adwk67 (dismissed):
LGTM!

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: Comment added in 7f6dc953.

crates/stackable-operator/src/commons/product_image_selection.rs
Techassi: I would be happy to enable that lint in a follow-up PR.

The lint in question is this one: https://rust-lang.github.io/rust-clippy/stable/index.html?search=field#inconsistent_struct_constructor

Techassi: I opened an issue for this: https://github.com/stackabletech/operator-rs/issues/1273.

adwk67 (approved):

Techassi:
### Release Notes

```adoc
==== New platform features

===== General

* It is possible to automatically select the latest patch version of an SDP release line using the new `spec.image.stackableVersionPolicy` field.
  It currently supports two values:
** `Exact`: Use the exact version specified or automatically derived.
   This is the current behaviour and the default value to ensure backwards compatibility.
** `LatestPatch`: Use the latest patch version of an SDP release line.

+
Consult the documentation page on xref:xref:concepts:product-image-selection.adoc[product image selection] for more details.
Implemented in https://github.com/stackabletech/operator-rs/pull/1226[operator-rs#1226].
```

PR #1224, opened 2026-06-17
chore: Bump kube and k8s-openapi as well as other crates

sbernauer:
# Description

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

### Author

- [x] Changes are OpenShift compatible
- [x] CRD changes approved
- [x] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [x] Integration tests passed (for non trivial changes)
- [x] Changes need to be "offline" compatible

### Reviewer

- [ ] Code contains useful comments
- [ ] Code contains useful logging statements
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)

### Acceptance

- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added

PR #1230, opened 2026-06-22
Fix the changelog

siegfriedweber:
# Description

Fix the changelog

The entries for #1224 and #1227 were added to the version 0.112.0 instead of the "Unreleased" section.


sbernauer (approved):

lfrancke:
Thanks for catching it.

PR #1228, opened 2026-06-19
feat(versioned)!: Require crd(doc = "...") to override the CRD root description

lfrancke:
# Description

Currently all generated CRDs are of this style "Auto-generated derived type for Foo via `CustomResource`"
This should allow us to use the [upstream feature](https://github.com/kube-rs/kube/pull/1359) to change this documentation.

PR #1231, opened 2026-06-22
chore: Release stackable-operator 0.113.0, stackable-versioned 0.11.0

siegfriedweber:
This PR releases stackable-operator 0.113.0 and stackable-versioned 0.11.0:

# stackable-operator

### Added

- Add documentation for the `roleGroups` field, which now shows up as a description in the generated CRDs ([#1227]).
- Add root descriptions to all CRDs defined in this crate (`AuthenticationClass`, `Listener`,
  `ListenerClass`, `PodListeners`, `S3Bucket`, `S3Connection`, `Scaler`), now that the `versioned`
  macro requires a `doc` argument ([#1228]).

### Changed

- BREAKING: Bump to kube `4.0.0`, k8s-openapi `0.28.0` and enable the Kubernetes 1.36 feature ([#1224]).

# stackable-versioned

### Added

- BREAKING: Add a required `#[versioned(crd(doc = "..."))]` argument that sets the root
  description of the generated CRD. It is required so that every CRD carries a meaningful
  description instead of the generic auto-generated one ([#1228]).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3084,7 +3084,7 @@ dependencies = [
 
 [[package]]
 name = "stackable-operator"
-version = "0.112.0"
+version = "0.113.0"
 dependencies = [
  "base64",
  "clap",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3186,7 +3186,7 @@ dependencies = [
 
 [[package]]
 name = "stackable-versioned"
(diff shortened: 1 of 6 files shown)


sbernauer:
LGTM, but we also need to release stackable-versioned

maltesander (approved):
LGTM!

PR #1235, opened 2026-06-29
ci: Use prek, bump action version

Techassi:
What is says on the tin.


Techassi:
We need to remove `pre-commit` as a required status check and mark `prek` as required instead.

**EDIT:** Done.

NickLarsenNZ (approved):
LGTM

PR #1233, opened 2026-06-26
chore: Import containerdebug

Techassi:
> [!CAUTION]
> This requires changes in `stackable-base` in https://github.com/stackabletech/docker-images.

> [!NOTE]
> It should be noted, that referenced PR numbers link to the **wrong** repository (this one) instead of the `containerdebug` repository.

This PR imports https://github.com/stackabletech/containerdebug as a workspace member with the complete history.

<details>
<summary>Import commands</summary>

```shell
cd containerdebug
git format-patch -o /tmp/patches --root src CHANGELOG.md Cargo.toml LICENSE README.md build.rs
cd operator-rs
git am --directory crates/containerdebug --3way /tmp/patches/*.patch
```

</details>

A PR which adds the new workspace member to the pre-commit config file will be raised later. It will also include general improvements to reduce the runtime of pre-commit when committing.

PR #1236, opened 2026-06-29
refactor: Use prek workspaces

Techassi:
Triggered by #1233.

prek supports nested configs (called workspace mode) which enables us to use indivudal, crate specific config files. Those hooks are only triggered if there are changes in the subdirectoy which drastically cuts down on runtime during commit and in CI.

Currently, most of these configs are copy+pasted and we ideally want some kind of central config + input based on the prek project/workspace member to avoid this duplicated code. This was already raised as an issue upstream.

Migration commands:

```shell
$ prek install
Hook already exists at `.git/hooks/pre-commit`, moved it to `.git/hooks/pre-commit.legacy`
Migration mode: prek will also run legacy hook `.git/hooks/pre-commit.legacy`. Use `--overwrite` to remove legacy hooks.
prek installed at `.git/hooks/pre-commit`

$ prek install --overwrite
Overwriting existing hook at `.git/hooks/pre-commit`
prek installed at `.git/hooks/pre-commit`
```

_It is encouraged to use `--overwrite` so that legacy hooks are not run to avoid duplicate work_


NickLarsenNZ (commented):
Does pre-commit support this?

Most users currently run pre-commit. If only prek supports this, then we probably need some kind of announcement before this gets merged.

Techassi:
No, this is prek specific, see https://prek.j178.dev/workspace/.

So yes, this would need to be announced.

NickLarsenNZ:
Let's wait until everyone has had time to update.

NickLarsenNZ (approved):
LGTM

PR #1241, opened 2026-07-02
feat(operator): Add `Role::{fixed,estimated}_replica_count`

sbernauer:
# Description

Needed for https://github.com/stackabletech/nifi-operator/pull/953, we can then also use it in https://github.com/stackabletech/trino-operator/blob/3158402b23edde1633b2639eb50c67bb1b41889c/rust/operator-binary/src/crd/mod.rs#L918

### Author

- [x] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/role_utils.rs
+++ b/crates/stackable-operator/src/role_utils.rs
@@ -434,6 +434,18 @@ where
             })
             .sum()
     }
+
+    /// Returns the estimated total number of replicas across all role groups.
+    ///
+    /// Unlike [`Self::fixed_replica_count`], this always returns a value: a role group with an unset
+    /// (i.e. [`None`]) replica count is assumed to run a single replica. Use this when a best-effort
+    /// estimate is needed even though the exact number of replicas is not hard-coded.
+    pub fn estimated_replica_count(&self) -> u32 {
+        self.role_groups
+            .values()
+            .map(|rg| u32::from(rg.replicas.unwrap_or(1)))
+            .sum()
+    }
 }
 
 impl<Config, ConfigOverrides, RoleConfig>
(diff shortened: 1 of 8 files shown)


crates/stackable-operator/src/role_utils.rs
```diff
@@ -403,6 +403,51 @@ where
...
 }
 
+impl<Config, ConfigOverrides, RoleConfig, CommonConfig>
+    Role<Config, ConfigOverrides, RoleConfig, CommonConfig>
+where
+    RoleConfig: Default + JsonSchema + Serialize,
+    CommonConfig: Default + JsonSchema + Serialize,
+    ConfigOverrides: Default + JsonSchema + Serialize,
+{
+    /// Returns [`Some<u32>`] in case the number of nodes is hard-coded to a certain value.
```
Techassi:
```suggestion
    /// Returns [`Some<u32>`] in case the number of replicas is hard-coded to a certain value.
```

crates/stackable-operator/src/role_utils.rs
```diff
@@ -403,6 +403,51 @@ where
...
+    ///
+    /// This is the case when all `replicas` are set to [`Some<u16>`], in which case they are simply
+    /// summed.
```
Techassi, line 416: This especially mentions _all_ `replicas` fields. The function however doesn't assert that this is actually the case in all role groups.

So either this comment needs adjusting, or the function body needs to change.

crates/stackable-operator/src/role_utils.rs
```diff
@@ -403,6 +403,51 @@ where
...
+
+        self.role_groups
+            .values()
+            .map(|rg| match rg.replicas {
+                None => None,
+                Some(0) if treat_zero_as_none => None,
+                // The individual replicas are [`u16`]s, so a [`u32`] sum has plenty of space.
+                Some(replicas) => Some(u32::from(replicas)),
+            })
+            .sum()
```
Techassi, line 437: I'm curious: Does `sum` flatten the `Option`s produced by the `map` above?

sbernauer: `sum` on an iterator of Options returns None if there is at least one None present in the iterator. If all are Some the sum is returned.

E.g. the `replica_counts_with_one_replica_unset` test showcases this

crates/stackable-operator/src/role_utils.rs
sbernauer: As soon as a single rolegroup has None (or 0 for that matter), None is returned. See your other comment in https://github.com/stackabletech/operator-rs/pull/1241/changes#r3512315258

crates/stackable-operator/src/role_utils.rs
```diff
@@ -654,4 +699,74 @@ mod tests {
             ]
         );
     }
+
+    #[test]
+    fn replica_counts_with_all_replicas_set() {
+        let role = construct_role_with_replicas([Some(3), Some(2), Some(5)]);
+
+        assert_eq!(role.fixed_replica_count(false), Some(10));
```
lfrancke: Maybe it's just me and this is non-blocking but these kinds of "true/false" boolean parameters are annoying. What does it mean? I now need to check the function. Can we use an enum `ZeroReplicas` instead? TreatAsNone vs. Count or something like that?

sbernauer: I had the same idea but wondered if this is an overkill: d3cc0a2121ba7653ac6c23b6e4c63b5ddd37dc67

crates/stackable-operator/src/role_utils.rs
Techassi: Ah, I see. Thanks!

Techassi (approved):

PR #1247, opened 2026-07-08
refactor!: Make `Role::{fixed,estimated}_replica_count` standalone

sbernauer:
# Description

Needed in https://github.com/stackabletech/nifi-operator/pull/953

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/role_utils.rs
+++ b/crates/stackable-operator/src/role_utils.rs
@@ -403,62 +403,6 @@ where
     }
 }
 
-impl<Config, ConfigOverrides, RoleConfig, CommonConfig>
-    Role<Config, ConfigOverrides, RoleConfig, CommonConfig>
-where
-    RoleConfig: Default + JsonSchema + Serialize,
-    CommonConfig: Default + JsonSchema + Serialize,
-    ConfigOverrides: Default + JsonSchema + Serialize,
-{
-    /// Returns [`Some<u32>`] in case the number of replicas is hard-coded to a certain value.
-    ///
-    /// This is the case when all `replicas` are set to [`Some<u16>`], in which case they are simply
-    /// summed.
-    ///
-    /// The argument `zero_replicas_counting` is a safety mechanism, which allows the caller to
-    /// decide if an explicit replica count of `0` should be treated as [`None`]. It also means that
-    /// [`None`] is returned in case no roleGroups are configured at all.
-    pub fn fixed_replica_count(&self, zero_replicas_counting: ZeroReplicasCounting) -> Option<u32> {
-        // An empty role has no fixed replica count when zeros are treated as None.
-        if zero_replicas_counting == ZeroReplicasCounting::TreatAsNone
-            && self.role_groups.is_empty()
-        {
-            return None;
-        }
-
-        self.role_groups
-            .values()
-            .map(|rg| match rg.replicas {
-                None => None,
-                Some(0) if zero_replicas_counting == ZeroReplicasCounting::TreatAsNone => None,
-                // The individual replicas are [`u16`]s, so a [`u32`] sum has plenty of space.
-                Some(replicas) => Some(u32::from(replicas)),
-            })
-            .sum()
-    }
-
-    /// Returns the estimated total number of replicas across all role groups.
-    ///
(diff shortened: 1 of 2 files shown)


crates/stackable-operator/src/role_utils.rs
```diff
@@ -594,6 +538,56 @@ impl<K: Resource> Display for RoleGroupRef<K> {
...
+/// This is the case when all `replicas` are set to [`Some<u16>`], in which case they are simply
+/// summed.
+///
+/// The argument `zero_replicas_counting` is a safety mechanism, which allows the caller to
+/// decide if an explicit replica count of `0` should be treated as [`None`]. It also means that
+/// [`None`] is returned in case no roleGroups are configured at all.
+pub fn fixed_replica_count<I: IntoIterator<Item = Option<u16>>>(
+    replicas: I,
+    zero_replicas_counting: ZeroReplicasCounting,
+) -> Option<u32> {
```
Techassi, line 555: (1/3) You could do something like this instead of peeking.


```suggestion
pub fn fixed_replica_count<I>(
    replicas: I,
    zero_replicas_counting: ZeroReplicasCounting,
) -> Option<u32>
where
    I: IntoIterator<Item = Option<u16>>,
    I::IntoIter: ExactSizeIterator,
{
```

crates/stackable-operator/src/role_utils.rs
```diff
@@ -594,6 +538,56 @@ impl<K: Resource> Display for RoleGroupRef<K> {
...
+    let mut replicas = replicas.into_iter().peekable();
```
Techassi, line 556: (2/3) This would then become


```suggestion
    let mut replicas = replicas.into_iter();
```

crates/stackable-operator/src/role_utils.rs
```diff
@@ -594,6 +538,56 @@ impl<K: Resource> Display for RoleGroupRef<K> {
...
+
+    // An empty role has no fixed replica count when zeros are treated as None.
+    if zero_replicas_counting == ZeroReplicasCounting::TreatAsNone && replicas.peek().is_none() {
```
Techassi, line 559: (3/3) And finally this would be


```suggestion
    if zero_replicas_counting == ZeroReplicasCounting::TreatAsNone && replicas.len() == 0 {
```

Sadly, `ExactSizeIterator::is_empty` is not stable yet.

Techassi (commented):
Just a small idea, otherwise lgtm.

crates/stackable-operator/src/role_utils.rs
sbernauer: TIL about `ExactSizeIterator`.
No strong opinion, however this has a higher requirement for the caller. I guess this is fine, as operators will most likely work on `Vec`s, on the other hand it's a function API decision driven by "internas" of the function, which only needs to know if the iter.length == 0, it doesn't need to know the exact length.
I let you decide.

Techassi: > TIL about `ExactSizeIterator`.

Hehe, you welcome!

> which only needs to know if the iter.length == 0, it doesn't need to know the exact length.

Yeah, there sadly is no such trait for this exact requirement - `ExactSizeIterator` is the closest we can get to it.

I would say both approaches have their pros and cons. I guess it boils down to "which iterator do we want to support here". I would say it is _fine_ to only allow iterators with a well-known size in 99% of cases. But you never know...

sbernauer: I think I'd go better safe than sorry and keep the current signature.
@NickLarsenNZ also suggested going with Option 1 ^^

Techassi: Addressed in c24c510.

crates/stackable-operator/src/role_utils.rs
Techassi: Addressed in c24c510.

crates/stackable-operator/src/role_utils.rs
Techassi: Addressed in c24c510.

Techassi (approved):

PR #1249, opened 2026-07-14
feat: Take dependencies into account when sorting environment variables

siegfriedweber:
# Description

Take dependencies into account when sorting environment variables

For example, the environment variables

```properties
ENV1=reference to $(ENV2)
ENV2=some value
ENV3=some value
```

are sorted in this order when iterated or converted to a vector:

```
ENV2
ENV1
ENV3
```

Part of stackabletech/issues#866

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/v2/builder/pod/container.rs
+++ b/crates/stackable-operator/src/v2/builder/pod/container.rs
@@ -1,6 +1,7 @@
 use std::{
-    collections::{BTreeMap, BTreeSet, btree_map},
+    collections::{BTreeMap, BTreeSet},
     str::FromStr,
+    vec,
 };
 
 use regex::Regex;
@@ -165,6 +166,15 @@ impl From<EnvVarSet> for Vec<EnvVar> {
         let mut vec: Self = value.0.values().cloned().collect();
         vec.sort_by_key(|env_var| env_var_closure.sort_key(env_var));
         vec
+    }
+}
+
+impl IntoIterator for EnvVarSet {
+    type IntoIter = vec::IntoIter<Self::Item>;
+    type Item = EnvVar;
+
+    fn into_iter(self) -> Self::IntoIter {
+        Vec::from(self).into_iter()
     }
 }
 
@@ -492,15 +502,6 @@ impl<'a> EnvVarDependencyResolver<'a> {
     }
 }
 
-impl IntoIterator for EnvVarSet {
-    type IntoIter = btree_map::IntoValues<EnvVarName, Self::Item>;
-    type Item = EnvVar;
-
-    fn into_iter(self) -> Self::IntoIter {
-        self.0.into_values()
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use std::str::FromStr;
@@ -692,89 +693,67 @@ mod tests {
     }
 
     #[test]
-    fn test_envvarset_with_references() {
+    fn test_vec_envvar_from_envvarset() {
         let env_var_set = EnvVarSet::new()
-            .with_value(&EnvVarName::from_str_unsafe("ENV1"), "value1")
-            // valid reference to a later variable
-            .with_value(&EnvVarName::from_str_unsafe("ENV2"), "$(ENV3)")
-            // valid reference to a later variable
-            .with_value(&EnvVarName::from_str_unsafe("ENV3"), "$(ENV4)")
-            // valid reference to an earlier variable
-            .with_value(&EnvVarName::from_str_unsafe("ENV4"), "$(ENV1)")
-            // invalid reference
-            .with_value(&EnvVarName::from_str_unsafe("ENV5"), "$(ENV?)")
-            // Same keys are allowed in Kubernetes, but not in `EnvVarSet`.
-            // The existing ENV1 is overridden.
-            .with_value(&EnvVarName::from_str_unsafe("ENV6"), "value6")
-            .with_value(&EnvVarName::from_str_unsafe("ENV6"), "$(ENV6)")
-            // multiple references
-            .with_value(
-                &EnvVarName::from_str_unsafe("ENV7"),
-                "$(ENV5) $(ENV8) $(ENV2)",
-            )
-            // multiple references with escaped and invalid references
-            .with_value(
-                &EnvVarName::from_str_unsafe("ENV8"),
-                "$(ENV1) $$(ENV9) $() $(ENV2)",
-            )
-            // No value
-            .with_field_path(&EnvVarName::from_str_unsafe("ENV9"), &FieldPathEnvVar::Name);
+            .with_value(&EnvVarName::from_str_unsafe("ENV1"), "$(ENV2)")
+            .with_value(&EnvVarName::from_str_unsafe("ENV2"), "value 2")
+            .with_value(&EnvVarName::from_str_unsafe("ENV3"), "value 3");
 
         assert_eq!(
             vec![
                 EnvVar {
-                    name: "ENV1".to_owned(),
-                    value: Some("value1".to_owned()),
+                    name: "ENV2".to_owned(),
+                    value: Some("value 2".to_owned()),
                     value_from: None
                 },
                 EnvVar {
-                    name: "ENV4".to_owned(),
-                    value: Some("$(ENV1)".to_owned()),
+                    name: "ENV1".to_owned(),
+                    value: Some("$(ENV2)".to_owned()),
                     value_from: None
                 },
                 EnvVar {
                     name: "ENV3".to_owned(),
-                    value: Some("$(ENV4)".to_owned()),
+                    value: Some("value 3".to_owned()),
                     value_from: None
-                },
-                EnvVar {
-                    name: "ENV2".to_owned(),
-                    value: Some("$(ENV3)".to_owned()),
-                    value_from: None
-                },
-                EnvVar {
-                    name: "ENV5".to_owned(),
-                    value: Some("$(ENV?)".to_owned()),
-                    value_from: None
-                },
-                EnvVar {
-                    name: "ENV6".to_owned(),
-                    value: Some("$(ENV6)".to_owned()),
-                    value_from: None
-                },
-                EnvVar {
-                    name: "ENV8".to_owned(),
-                    value: Some("$(ENV1) $$(ENV9) $() $(ENV2)".to_owned()),
-                    value_from: None
-                },
-                EnvVar {
-                    name: "ENV7".to_owned(),
-                    value: Some("$(ENV5) $(ENV8) $(ENV2)".to_owned()),
-                    value_from: None
-                },
-                EnvVar {
-                    name: "ENV9".to_owned(),
-                    value: None,
-                    value_from: Some(EnvVarSource {
-                        field_ref: Some(ObjectFieldSelector {
-                            field_path: FieldPathEnvVar::Name.to_string(),
-                            ..ObjectFieldSelector::default()
-                        }),
-                        ..EnvVarSource::default()
-                    }),
                 },
             ],
             Vec::from(env_var_set)
         );
+    }
+
+    #[test]
+    fn test_envvarset_intoiterator() {
+        let env_var_set = EnvVarSet::new()
+            .with_value(&EnvVarName::from_str_unsafe("ENV1"), "$(ENV2)")
+            .with_value(&EnvVarName::from_str_unsafe("ENV2"), "value 2")
+            .with_value(&EnvVarName::from_str_unsafe("ENV3"), "value 3");
+
+        let mut iter = env_var_set.into_iter();
+
+        assert_eq!(
+            Some(EnvVar {
+                name: "ENV2".to_owned(),
+                value: Some("value 2".to_owned()),
+                value_from: None
+            }),
+            iter.next()
+        );
+        assert_eq!(
+            Some(EnvVar {
+                name: "ENV1".to_owned(),
+                value: Some("$(ENV2)".to_owned()),
+                value_from: None
+            }),
+            iter.next()
+        );
+        assert_eq!(
+            Some(EnvVar {
+                name: "ENV3".to_owned(),
+                value: Some("value 3".to_owned()),
+                value_from: None
+            }),
+            iter.next()
+        );
+        assert_eq!(None, iter.next());
     }
 }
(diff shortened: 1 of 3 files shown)


crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -346,4 +709,69 @@ mod tests {
...
+                },
+            ],
+            Vec::from(env_var_set)
+        );
+    }
+
+    #[test]
+    fn test_envvarset_intoiterator() {
+        let env_var_set = EnvVarSet::new()
+            .with_value(&EnvVarName::from_str_unsafe("ENV1"), "$(ENV2)")
```
NickLarsenNZ, line 762: What is doing the substitution?

I would have expected `${...}` instead of command substitution `$(...)` if we were talking about sh/bash. I'm assuming this is for custom interpolation that we perform.

NickLarsenNZ (dismissed):
LGTM, but left a comment to check my understanding.

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -10,14 +16,29 @@ use crate::{
     v2::types::kubernetes::{ConfigMapKey, ConfigMapName, ContainerName},
 };
 
+/// Pattern for an escaped dollar sign (`$$`)
+static ESCAPED_DOLLAR_SIGN_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"\$\$").expect("should be a valid regular expression"));
```
Techassi: I would go as far as and say it _must_ be a valid regex.


```suggestion
    LazyLock::new(|| Regex::new(r"\$\$").expect("static string must be a valid regular expression"));
```

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -144,16 +177,346 @@ impl EnvVarSet {
 
 impl From<EnvVarSet> for Vec<EnvVar> {
     fn from(value: EnvVarSet) -> Self {
-        value.0.values().cloned().collect()
+        let dependency_resolver =
+            EnvVarDependencyResolver::new(&value, ENV_VAR_DEPENDENCY_RESOLVER_MAX_RECURSION_DEPTH);
+
+        let mut vec: Self = value.0.values().cloned().collect();
```
Techassi: There is no need to clone. We can consume `value`.


```suggestion
        let mut vec: Self = value.0.into_values().collect();
```

Techassi: Oh we borrow `value` above and need to hold onto it **after** we collect the values into the Vec.

This indicates that the design of `EnvVarDependencyResolver` is sub optimal.

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -144,16 +177,346 @@ impl EnvVarSet {
...
 
     fn into_iter(self) -> Self::IntoIter {
-        self.0.into_values()
+        Vec::from(self).into_iter()
+    }
+}
+
+/// Resolves dependencies between environment variables and provides sort keys which take these
+/// dependencies into account
+pub struct EnvVarDependencyResolver<'a> {
```
Techassi, line 226: I feel like this whole struct should be build around owning the `EnvVarSet`. This would get rid of a whole bunch of clones all over the place. This additionally makes sense, because it is only used in a context in which the underlying `EnvVarSet` is consumed.

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -144,16 +177,346 @@ impl EnvVarSet {
...
+    ///     Some(BTreeSet::from(["ENV4".to_owned()])),
+    ///     resolver.calculate_closure(&env_var3)
+    /// );
+    /// assert_eq!(Some(BTreeSet::new()), resolver.calculate_closure(&env_var4));
+    /// assert_eq!(None, resolver.calculate_closure(&env_var5));
+    /// assert_eq!(None, resolver.calculate_closure(&env_var6));
+    /// assert_eq!(None, resolver.calculate_closure(&env_var7));
+    /// assert_eq!(None, resolver.calculate_closure(&env_var8));
+    /// ```
+    pub fn calculate_closure(&self, env_var: &EnvVar) -> Option<BTreeSet<String>> {
```
Techassi: What's the reason this needs to be public?

Techassi (changes requested):

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: Environment variables in Kubernetes are referenced with `$(...)`, see https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/.

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: Doc tests can only access public functions. I wanted to test each function in isolation because that made it easier to achieve high test coverage. I know this isn't idiomatic, but Rust doesn't offer an alternative I'd like better.

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: changed in 9a3111d076eef844589418b81efdece841e9700d

"Should" is often used in these cases. If I write a unit test that is **EXPECTED** to panic, I use the attribute `#[should_panic]` even if I know that it **MUST** panic as long as nobody changes the code.

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: `EnvVarDependencyResolver` resolves the dependencies and provides sort keys for a later sort. Which design would you prefer?

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: > I feel like this whole struct should be build around owning the EnvVarSet.

If `EnvVarDependencyResolver` would own the `EnvVarSet`, then it would not be available to the sorting algorithm anymore.

> This additionally makes sense, because it is only used in a context in which the underlying `EnvVarSet` is consumed.

An `EnvVarSet` should be usable for several containers or role-groups. So, consuming it, does not feel right for me.

> This would get rid of a whole bunch of clones all over the place.

The code could be written completely without clones. But every avoided clone makes the code a bit harder to read. The `EnvVarSets` are small and every `EnvVar` usually contains a short name and a short value. So the clones are cheap.

However, I changed the code to use references (05660fae204b96d53cc2c3536c07733bf95a0554). The `IntoIterator` for `EnvVarSet` still clones the `EnvVar`s, because this would be the alternative:

```rust
impl IntoIterator for EnvVarSet {
    type IntoIter = vec::IntoIter<Self::Item>;
    type Item = EnvVar;

    fn into_iter(self) -> Self::IntoIter {
        let sorted_indices: Vec<usize> = {
            let dependency_resolver = EnvVarDependencyResolver::new(
                &self,
                ENV_VAR_DEPENDENCY_RESOLVER_MAX_RECURSION_DEPTH,
            );

            let mut indexed_env_vars: Vec<(usize, &EnvVar)> =
                self.0.values().enumerate().collect();
            indexed_env_vars
                .sort_by_cached_key(|(_, env_var)| dependency_resolver.sort_key(env_var));

            indexed_env_vars
                .into_iter()
                .map(|(index, _)| index)
                .collect()
        };

        let mut env_vars: Vec<Option<EnvVar>> = self.0.into_values().map(Some).collect();

        sorted_indices
            .into_iter()
            .map(|index| {
                env_vars[index]
                    .take()
                    .expect("sorting must yield each index exactly once")
            })
            .collect::<Vec<_>>()
            .into_iter()
    }
}
```

crates/stackable-operator/src/v2/builder/pod/container.rs
NickLarsenNZ: TIL :+1:

crates/stackable-operator/src/v2/builder/pod/container.rs
NickLarsenNZ: What about `pub(crate)` (I assume that still doesn't give it enough visibility)?

siegfriedweber: `pub(crate)` is not sufficient for doc tests.

If we would own `EnvVar`, then e.g. `referenced_env_vars` would be a public function there and it would be accepted that it is tested in isolation. We could wrap `EnvVar`, but that would be overkill.

crates/stackable-operator/src/v2/builder/pod/container.rs
Techassi: "Should" always indicates that there is _some_ condition (maybe unknown to us), that _can_ fail. I want to make it 100 percent clear that this expression **must** be able to be parsed as a regular expression. We, as developers, guarantee that the static string provided is valid and therefore **must** be be able to be parsed.

I'm fine with using "should" in situations which we cannot fully control, like reading an env var we expect, or trying to parse dynamic/user input.

crates/stackable-operator/src/v2/builder/pod/container.rs
Techassi: > If `EnvVarDependencyResolver` would own the `EnvVarSet`, then it would not be available to the sorting algorithm anymore.

External sorting: yes - internal sorting: no. The `EnvVarDependencyResolver` could work in the following way:

1. Figure out the dependencies (which are the sort keys)
2. Do the actual sorting on the owned `EnvVarSet`

> An `EnvVarSet` should be usable for several containers or role-groups. So, consuming it, does not feel right for me.

We could still make it work by introducing a public function on `EnvVarDependencyResolver` which consumes it, and returns the inner (sorted) `EnvVarSet`. This set can then be used (and probably cloned) by multiple consumers.

```rust
impl EnvVarDependencyResolver {
    pub fn new(set: EnvVarSet) -> Self;

    pub fn sort(self) -> EnvVarSet;
}
```

crates/stackable-operator/src/v2/builder/pod/container.rs
Techassi: See https://github.com/stackabletech/operator-rs/pull/1249#discussion_r3613064775.

crates/stackable-operator/src/v2/builder/pod/container.rs
Techassi: > Doc tests can only access public functions

Wait, really? I'm very surprised by this, because I really thought private functions could also be tested with doc tests...

siegfriedweber: Doc tests are compiled as separate units and linked to the library:

> Note that they will still link against only the public items of your crate; if you need to test private items, you need to write a unit test.
>
> https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests

see also [rust-lang/rust#60820](https://www.github.com/rust-lang/rust/issues/60820)

Techassi: Oh wow, thanks - TIL.

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: `sort` cannot return an `EnvVarSet`. An `EnvVarSet` is a `BTreeMap` which always iterates in _key_ order. The entire point of the dependency resolver is to find a _topological_ order. However, `sort` could return a `Vec<EnvVar>`.

Your first comment was about "a whole bunch of clones". After 05660fae204b96d53cc2c3536c07733bf95a0554, which clones are left that you are not satisfied with?

Techassi: > `sort` cannot return an `EnvVarSet`. An `EnvVarSet` is a `BTreeMap` which always iterates in key order. The entire point of the dependency resolver is to find a _topological_ order. However, sort could return a `Vec<EnvVar>`.

I see, makes sense. My bad. The current implementation already looks way better. I have one small improvement in mind, which I will submit as a suggestion in a follow-up review.

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -142,18 +185,353 @@ impl EnvVarSet {
...
 
-impl From<EnvVarSet> for Vec<EnvVar> {
-    fn from(value: EnvVarSet) -> Self {
-        value.0.values().cloned().collect()
+impl<'a> From<&'a EnvVarSet> for Vec<&'a EnvVar> {
+    fn from(value: &'a EnvVarSet) -> Self {
+        value.into_iter().collect()
+    }
+}
+
```
Techassi: We can additionally add this:


```suggestion
impl From<EnvVarSet> for Vec<EnvVar> {
    fn from(value: EnvVarSet) -> Self {
        value.into_iter().collect()
    }
}
```

crates/stackable-operator/src/v2/builder/pod/container.rs
```diff
@@ -142,18 +185,353 @@ impl EnvVarSet {
...
+    type IntoIter = vec::IntoIter<Self::Item>;
     type Item = EnvVar;
 
     fn into_iter(self) -> Self::IntoIter {
-        self.0.into_values()
+        Vec::from(&self)
+            .into_iter()
+            .cloned()
+            .collect::<Vec<_>>()
+            .into_iter()
```
Techassi: Based on the suggestion above, this can be shortened into:


```suggestion
        Vec::from(self).into_iter()
```

Techassi (dismissed):
LGTM, just a minor (2-part) suggestion.

crates/stackable-operator/src/v2/builder/pod/container.rs
siegfriedweber: Your suggestions introduced a call cycle between `From<EnvVarSet> for Vec<EnvVar>` and `IntoIterator for EnvVarSet`. I fixed it in de636c0d1c02b569be31c2280c0bb3cc4fa67a9b, but the code is not shorter than before.

Techassi (approved):

PR #1255, opened 2026-07-29
feat: Add an optional task appender to AutomaticContainerLogConfig

Churi12:
# Description

Adds a task appender to AutomaticContainerLogConfig, next to console and file.

Airflow renders a separate task handler, which is the one that feeds the log view in its web UI. Its level was not configurable, and there was no way to express it: loggers is a map so any key works, but console and file are named fields, so a sibling could not be added downstream. Setting the airflow.task logger level instead is not equivalent, since a logger threshold applies to every destination at once, whereas the point is to send DEBUG to the log files and to Vector while showing only INFO in the UI.

The field is optional and left unset by default, so the configuration rendered by every other product is unchanged. Only products with a separate task log destination use it, which today means Airflow alone.

The Merge impl for this struct is hand written rather than derived, so the field had to be added there too. Without it a level set on a role group is silently dropped, so there is a test covering both directions, and it does fail if the merge block is removed.

The consumer side is stackabletech/airflow-operator#829, which is what motivated this.

One thing worth a maintainer opinion, since it is a shared struct: a plain sibling field is the smallest change that matches stackabletech/airflow-operator#650, which is why I went with it. The alternatives would be a generic appenders map, or keeping this out of the shared struct entirely and letting Airflow carry it. Happy to rework if you would rather have one of those. I looked at additional_config as an escape hatch first, but it only exists on create_logback_config, and Airflow renders its own Python template rather than using any of the log4j or logback builders, so it does not help here.

### Author

- [x] Changes are OpenShift compatible: not applicable, no resource or image changes
- [ ] CRD changes approved: this adds an optional field to a shared struct, see the note above

### Reviewer

- [ ] Changelog updated

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Add an optional `task` appender to `AutomaticContainerLogConfig`, next to `console` and
+  `file`. It is only used by products which have a separate task log destination, currently
+  Airflow, where it sets the log level shown in the web UI. It defaults to unset, so the
+  rendered configuration of every other product is unchanged ([#1255]).
+
+[#1255]: https://github.com/stackabletech/operator-rs/pull/1255
+
 ## [0.114.0] - 2026-07-22
 
 ### Added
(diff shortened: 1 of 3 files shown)


Churi12:
Closing this. Discussion is on stackabletech/airflow-operator#829: the preference is to not add a field to a shared struct that only Airflow would use, and to carry the level in a magic airflow.task key in loggers instead. That needs no CRD change at all, so this is not needed.

Thanks @sbernauer for the steer.

PR #992, opened 2025-03-28
feat: Add Deployments to ClusterResources

razvan:
# Description

Part of : https://github.com/stackabletech/spark-k8s-operator/issues/284

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

```[tasklist]
# Author
- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

PR #1256, opened 2026-07-31
fix: Add Deployment to delete_orphaned_resources

Maleware:
Allowing OPA being a deployment, we should take care of orphaned deployment clean-ups as it will be possible to switch from `Deployment` to `DaemonSet`

Fixup of https://github.com/stackabletech/operator-rs/pull/992


sbernauer:
This is actually a bugfix of https://github.com/stackabletech/operator-rs/pull/992, good catch!

Maleware:
That's good to know! Thanks for adding that. Then we might be good to merge it, I wanted to think about wether this has impact somewhere else, but doesn't seem like.

NickLarsenNZ (approved):
LGTM

sbernauer:
A changelog entry would be great :)

PR #1258, opened 2026-08-04
chore: Release stackable-operator `0.115.0`

sbernauer:
### Changed

- BREAKING: `PodSecurityContextBuilder::new` was removed in favor of `PodSecurityContextBuilder::with_stackable_defaults`
  (same for `SecurityContextBuilder`) ([#1205]).
  This function already sets up some defaults we want to use across the platform.
  Currently this is `runAsNonRoot: true` for `PodSecurityContextBuilder`, which might cause product Pods to crash and require changes.
- BREAKING: `PodSecurityContextBuilder::run_as_non_root` now takes a `bool` instead of assuming consumers always want to set it to `true` ([#1205]).
  This is needed to allow users setting it to `false` in case the new `with_stackable_defaults` function sets it to `true`.
- BREAKING: `SecurityContextBuilder::run_as_root` has been removed ([#1205]).
- BREAKING: Bump `kube` to `4.2.0` ([#1257]).
  - This fixes a long-standing issue, where `additionalPrinterColumns`, `categories` and `shortNames` where always included in the CRD, which lead to ArgoCD thinking the CRs where out of sync.

[#1205]: https://github.com/stackabletech/operator-rs/pull/1205
[#1257]: https://github.com/stackabletech/operator-rs/pull/1257

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3670,7 +3670,7 @@ dependencies = [
 
 [[package]]
 name = "stackable-operator"
-version = "0.114.0"
+version = "0.115.0"
 dependencies = [
  "base64 0.23.0",
  "clap",
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [0.115.0] - 2026-08-04
+
 ### Changed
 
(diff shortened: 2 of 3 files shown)


siegfriedweber (commented):
Does your change in https://github.com/stackabletech/operator-rs/pull/1257/changes#diff-bc7736d26396b13b125507bb70f16bb26bfcefab22b1203338ab7e2092821ecf also require a stackable-versioned release?

sbernauer:
I'd say no, as it's a purely internal change. IMO we only need to call out bumps to re-exported crates (kube, k8s-openapi), as they are technically breaking.

siegfriedweber (approved):
LGTM

PR #1260, opened 2026-08-05
feat: Add `Key::new_trimmed` helper function

sbernauer:
# Description

I need this helper function to fix a bug in commons-operator: https://github.com/stackabletech/commons-operator/pull/443

### Author

- [ ] CRD changes approved

### Reviewer

- [x] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/kvp/key.rs
+++ b/crates/stackable-operator/src/kvp/key.rs
@@ -3,7 +3,7 @@ use std::{fmt::Display, ops::Deref, str::FromStr, sync::LazyLock};
 use regex::Regex;
 use snafu::{ResultExt, Snafu, ensure};
 
-use crate::utils::length_enforcement::ensure_max_length;
+use crate::utils::length_enforcement::ensure_max_string_length;
 
 const KEY_PREFIX_MAX_LEN: usize = 253;
 const KEY_NAME_MAX_LEN: usize = 63;
@@ -139,7 +139,7 @@ impl Deref for Key {
 impl Key {
     /// (Optionally) shortens the `prefix` and `name` to make sure they produce a valid [`Key`].
     ///
-    /// See [`ensure_max_length`] for details on the shortening algorithm.
+    /// See [`ensure_max_string_length`] for details on the shortening algorithm.
     pub fn shortened_to_valid_length(
         prefix: Option<&str>,
         name: impl Into<String>,
@@ -148,7 +148,7 @@ impl Key {
         // name, shortening won't make it one. In particular, a prefix must end in a letters-only
         // TLD, but the appended hash adds a hyphen and probably digits, very likely being an
         // invalid result.
-        let name = ensure_max_length(name, KEY_NAME_MAX_LEN, 8);
+        let name = ensure_max_string_length(name, KEY_NAME_MAX_LEN, 8);
 
         let key = match prefix {
             Some(prefix) => format!("{prefix}/{name}"),
--- a/crates/stackable-operator/src/kvp/key.rs
+++ b/crates/stackable-operator/src/kvp/key.rs
@@ -137,7 +137,10 @@ impl Deref for Key {
 }
 
 impl Key {
-    pub fn with_enforced_length(
+    /// (Optionally) shortens the `prefix` and `name` to make sure they produce a valid [`Key`].
+    ///
+    /// See [`ensure_max_length`] for details on the shortening algorithm.
+    pub fn shortened_to_valid_length(
         prefix: impl Into<String>,
         name: impl Into<String>,
     ) -> Result<Self, KeyError> {
--- a/crates/stackable-operator/src/kvp/key.rs
+++ b/crates/stackable-operator/src/kvp/key.rs
@@ -3,6 +3,8 @@ use std::{fmt::Display, ops::Deref, str::FromStr, sync::LazyLock};
 use regex::Regex;
 use snafu::{ResultExt, Snafu, ensure};
 
+use crate::utils::length_enforcement::ensure_max_length;
+
 const KEY_PREFIX_MAX_LEN: usize = 253;
 const KEY_NAME_MAX_LEN: usize = 63;
 
@@ -135,6 +137,16 @@ impl Deref for Key {
 }
 
 impl Key {
+    pub fn with_enforced_length(
+        prefix: impl Into<String>,
+        name: impl Into<String>,
+    ) -> Result<Self, KeyError> {
+        let prefix = ensure_max_length(prefix, KEY_PREFIX_MAX_LEN, 8);
+        let name = ensure_max_length(name, KEY_NAME_MAX_LEN, 8);
+
+        Self::from_str(&format!("{prefix}/{name}"))
+    }
+
     /// Retrieves the key's prefix.
     ///
     /// ```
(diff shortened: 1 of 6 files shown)


crates/stackable-operator/src/v2/macros/attributed_string_type.rs
```diff
@@ -6,6 +6,9 @@ use strum::{EnumDiscriminants, IntoStaticStr};
 /// Duplicates the private constant [`crate::kvp::LABEL_VALUE_MAX_LEN`]
 pub const MAX_LABEL_VALUE_LENGTH: usize = 63;
 
+/// Maximum length of annotation names
+pub const MAX_ANNOTATION_NAME_LENGTH: usize = 63;
```
siegfriedweber: This constant belongs in `kvp::key` and should be named `MAX_KEY_NAME_LENGTH`.

Rather than exposing the value publicly, it might be cleaner to offer a `sanitize` function that takes a key prefix and name and returns a valid `kvp::key::Key`.

crates/stackable-operator/src/v2/role_group_utils.rs
```diff
@@ -99,7 +99,11 @@ impl ResourceNames {
     /// `max_length < 1 /* character */ + 1 /* dash */ + hash_length`.
     ///
     /// Kubernetes object names cannot contain non-ASCII characters.
-    fn ensure_max_length(resource_name: String, max_length: usize, hash_length: usize) -> String {
+    pub fn ensure_max_length(
+        resource_name: String,
+        max_length: usize,
+        hash_length: usize,
+    ) -> String {
```
siegfriedweber: This function was written for resource names, where the assertions always hold. If it's now used to shorten arbitrary strings (which could also come from the user) then you can `expect` the operator to crash.

That said, I do see the usefulness. I'd suggest making it safe for arbitrary strings (which isn't trivial) and moving it to a string utility module.

siegfriedweber (changes requested):

crates/stackable-operator/src/v2/macros/attributed_string_type.rs
sbernauer: I removed the constant and added the requested function. However, I didn't call it `sanitize`, as to me that gives off security vibes, but happy to discuss the name :)

crates/stackable-operator/src/v2/role_group_utils.rs
sbernauer: Moved it into `crates/stackable-operator/src/utils/length_enforcement.rs`.

> which isn't trivial

However, I'm pretty sure I missed something, will test non-asci strings now

sbernauer: I pushed a commit that adds UTF-8 support: f0a76b3
It work's, but also feels a bit awkward to use

crates/stackable-operator/src/kvp/key.rs
```diff
@@ -135,6 +137,19 @@ impl Deref for Key {
...
 
 impl Key {
+    /// (Optionally) shortens the `prefix` and `name` to make sure they produce a valid [`Key`].
+    ///
+    /// See [`ensure_max_length`] for details on the shortening algorithm.
+    pub fn shortened_to_valid_length(
+        prefix: impl Into<String>,
+        name: impl Into<String>,
+    ) -> Result<Self, KeyError> {
+        let prefix = ensure_max_length(prefix, KEY_PREFIX_MAX_LEN, 8);
```
siegfriedweber: Shortening the prefix is pointless: if it isn't already a valid DNS subdomain name, shortening won't make it one. In particular, a prefix must end in a letters-only TLD, but the appended hash adds a hyphen and probably digits, guaranteeing an invalid result.

A unit test for both the shortened and unshortened case would be good.

crates/stackable-operator/src/utils/length_enforcement.rs
```diff
@@ -0,0 +1,164 @@
...
+///
+/// It is recommended to only use ASCII characters, but this function also handles UTF-8: Multi-byte
+/// characters are never split up, so the result can be shorter than the maximum length.
+///
+/// If the truncation does not leave any character then only the hash is returned.
+///
+/// # Panics
+///
+/// Panics if `max_length_bytes < 1 /* character */ + 1 /* dash */ + hash_length`.
+pub fn ensure_max_length(
```
siegfriedweber: **nit:** `utils::length_enforcement::ensure_max_length` reads as generic, but it only works on strings. What would the corresponding function for collections like vectors be called?

siegfriedweber (changes requested):

crates/stackable-operator/src/kvp/key.rs
sbernauer: I'm glad that you know this stuff better than I do 😅 
Fixed in a9cbd52. Aalso made the prefix optional, so that callers can only pass the name

crates/stackable-operator/src/utils/length_enforcement.rs
sbernauer: I renamed it to `ensure_max_string_length` in 0c8b4fc. But totally happy to name it to something else if you recommend

crates/stackable-operator/src/utils/length_enforcement.rs
```diff
@@ -0,0 +1,175 @@
...
+            "a--f7a0".to_owned(),
+            ensure_max_string_length("a--defgh".to_owned(), 7, 4)
+        );
+
+        // A hash_length longer than the produced hash string may not produce the desired result.
+        // Just use sensible values!
+        assert_eq!(
+            "aaaaaaaaa-d476ce01c3787bcab054a2cf48d6af6dd303a0eb549e21a74125132f79d90c36".to_owned(),
+            ensure_max_string_length("a".repeat(1011), 1010, 1000)
+        );
```
siegfriedweber:
```
thread 'utils::length_enforcement::test::ensure_max_string_length_ascii' (60970) panicked at crates/stackable-operator/src/utils/length_enforcement.rs:20:5:
We hash using sha256, so we don't produce more than 64 bytes
```

siegfriedweber (changes requested):

crates/stackable-operator/src/utils/length_enforcement.rs
sbernauer: Pff, who would write such an insane test? :P
Removed in 03d8966

siegfriedweber (dismissed):
LGTM

crates/stackable-operator/src/kvp/key.rs
```diff
@@ -135,6 +137,26 @@ impl Deref for Key {
...
 
 impl Key {
+    /// Shortens `name` if needed, so that it does not exceed the maximum key name length.
+    ///
+    /// The `prefix` is used as-is: If it isn't already a valid DNS subdomain name, shortening won't
+    /// make it one. In particular, a prefix must end in a letters-only TLD, but the appended hash
+    /// adds a hyphen and probably digits, very likely being an invalid result.
+    ///
+    /// See [`ensure_max_string_length`] for details on the shortening algorithm.
+    pub fn shortened_to_valid_length(
```
Techassi: Maybe `new_trimmed` is a slightly shorter alternative name for this.

sbernauer: Works for me: fcf48c9

siegfriedweber (approved):

PR #1261, opened 2026-08-06
feat: Add functions for recommended labels

siegfriedweber:
# Description

Add functions for recommended labels to `v2::kvp::label`

The functions use specific types, e.g. `ClusterName` and `ProductName`, and are infallible.

**Examples:**

- stackabletech/airflow-operator#838
  only simple changes, but a lot of them
- stackabletech/opensearch-operator#171
  only few changes, thanks to the contexts `RoleBuilder` and `RoleGroupBuilder` provide

Part of stackabletech/internal-issues#167

### Reviewer

- [x] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -8,7 +8,10 @@ All notable changes to this project will be documented in this file.
 
 - Add the Cargo feature `kube-cel` that enables the `cel` feature on the `kube` crate ([#1259]).
 - Add `length_enforcement::ensure_max_string_length` and `Key::shortened_to_valid_length` helper functions ([#1260]).
-- [v2] Add functions for recommended labels to `v2::kvp::label` ([#1261]).
+
+### Changed
+
+- BREAKING: [v2] Improve functions for recommended labels in `v2::kvp::label` ([#1261]).
 
 [#1259]: https://github.com/stackabletech/operator-rs/pull/1259
 [#1260]: https://github.com/stackabletech/operator-rs/pull/1260
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -7,8 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Add the Cargo feature `kube-cel` that enables the `cel` feature on the `kube` crate ([1259]).
+- [v2] Add functions for recommended labels to `v2::kvp::label` ([#1261]).
 
 [1259]: https://github.com/stackabletech/operator-rs/pull/1259
+[1261]: https://github.com/stackabletech/operator-rs/pull/1261
 
 ## [0.115.0] - 2026-08-04
 
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -6,11 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Add the Cargo feature `kube-cel` that enables the `cel` feature on the `kube` crate ([1259]).
+- Add the Cargo feature `kube-cel` that enables the `cel` feature on the `kube` crate ([#1259]).
 - [v2] Add functions for recommended labels to `v2::kvp::label` ([#1261]).
 
-[1259]: https://github.com/stackabletech/operator-rs/pull/1259
-[1261]: https://github.com/stackabletech/operator-rs/pull/1261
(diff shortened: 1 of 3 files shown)


crates/stackable-operator/CHANGELOG.md
```diff
@@ -6,9 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Add the Cargo feature `kube-cel` that enables the `cel` feature on the `kube` crate ([1259]).
+- Add the Cargo feature `kube-cel` that enables the `cel` feature on the `kube` crate ([#1259]).
+- [v2] Add functions for recommended labels to `v2::kvp::label` ([#1261]).
```
adwk67: I think it is more accurate to have this under `Changed` as we have removed/replaced API functions.

crates/stackable-operator/src/v2/kvp/label.rs
```diff
@@ -1,151 +1,291 @@
...
+) -> ControllerName {
+    let mut full_controller_name = format!(
+        "{}_{}",
+        operator_name.to_label_value(),
+        controller_name.to_label_value()
+    );
+    full_controller_name.truncate(ControllerName::MAX_LENGTH);
+
+    ControllerName::from_str(&full_controller_name).expect(
+        "is a valid ControllerName because it satifies both the character and length constraints",
```
adwk67:
```suggestion
        "is a valid ControllerName because it satisfies both the character and length constraints",
```
The claim about the character constraint is not yet fully fulfilled - e.g. a truncated value could have trailing non-alphanumerics (it's checked above with Label::try_from but it could panic before it even gets there).

crates/stackable-operator/src/v2/kvp/label.rs
```diff
@@ -1,151 +1,291 @@
...
+        .expect("produces a valid label because the statically defined key is valid and the value implements NameIsValidLabelValue")
+}
+
+/// Creates the `app.kubernetes.io/component` label with the given role as value.
+pub fn label_app_kubernetes_io_component(role_name: &RoleName) -> Label {
+    Label::component(&role_name.to_label_value())
+        .expect("produces a valid label because the value implements NameIsValidLabelValue")
+}
+
+/// Creates the `app.kubernetes.io/role-group` label with the the given role group as value.
```
adwk67:
```suggestion
/// Creates the `app.kubernetes.io/role-group` label with the given role group as value.
```

crates/stackable-operator/src/v2/kvp/label.rs
```diff
@@ -1,151 +1,291 @@
...
-    impl HasName for Cluster {
-        fn to_name(&self) -> String {
-            self.object_meta
-                .name
-                .clone()
-                .expect("should be set in Cluster::new")
-        }
-    }
+    #[test]
+    fn role_selector_is_subset_of_recommended_role_labels() {
```
adwk67, line 262: Some tests are prefixed with "test_" and some aren't - we are very inconsistent on this but can at least keep them consistent in the same file

adwk67 (changes requested):

crates/stackable-operator/CHANGELOG.md
siegfriedweber: It is even a breaking change.

848d18d4e3bde63e8f8701783b4433e9a74f4ae4

crates/stackable-operator/src/v2/kvp/label.rs
siegfriedweber: Good catch!

Now, I do not try to fix the full controller name and just return the operator name: ea6e645831402bcf3c6aa2b617ba019333ab2d01

crates/stackable-operator/src/v2/kvp/label.rs
siegfriedweber: I revised all comments: ea6e645831402bcf3c6aa2b617ba019333ab2d01

crates/stackable-operator/src/v2/kvp/label.rs
siegfriedweber: This was actually a deliberate decision on my part, even if I know that our code conventions state otherwise. Usually, I write a test case for each function named `test_<function_name>` and if further tests are required to test the interoperability of these functions, I just write a descriptive name. If I would omit the `test_` prefix then the function names would become something like `<function_under_test>_does_the_right_thing`.

But code conventions are code conventions and therefore I renamed all functions: 5d21e6cc8d8cdef002538f08243786cd738138fe

adwk67 (approved):
LGTM

PR #1264, opened 2026-08-14
feat: Adds a warning when a defined object override was not applied

sweb:
# Description

Adds a warning when a defined object override was not applied.

~~Optional: Adds the base namespace as default when the namespace of the override is not provided~~ I removed this as it seems too much of a change. 

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance


crates/stackable-operator/src/cluster_resources.rs
```diff
@@ -657,6 +664,8 @@ impl<'a> ClusterResources<'a> {
     ///
     /// * `client` - The client which is used to access Kubernetes
     pub async fn delete_orphaned_resources(self, client: &Client) -> Result<()> {
+        self.warn_about_unmatched_object_overrides();
```
Maleware, line 675: Thanks for the call. I understand it better now. I'd like to have a comment here, why the decision was to put in `delete_orphaned_resources` something along the lines:

Did it here because last step as it's only about resources which didn't had a match. Did so to not clutter Snafu warnings directly related to malformed CRDs.

Maleware (commented):
Up to the comment, LGTM thanks!

crates/stackable-operator/src/cluster_resources.rs
sweb: addressed in https://github.com/stackabletech/operator-rs/pull/1264/commits/9a2adee270b91d580cde02631dbcf7b4206dba0e

@Maleware I thought a bit more about this - I think you are right that we should warn earlier. My initial thinking was that an object override would not make an apply step fail but that is wrong. When I tried to add it earlier I ran into the issue that you actually only get to match against all possible overrides, once all applies have been done. There are some checks we could do before, but not all of them.

To keep this small, I'd leave it at this, but I think this deserves further improvements - but from what I looked at, that would be more invasive than this change.

crates/stackable-operator/src/cluster_resources.rs
```diff
@@ -681,6 +698,40 @@ impl<'a> ClusterResources<'a> {
...
+                .unwrap_or("<not set>");
+
+            warn!(
+                "The objectOverride at index {index} (apiVersion: {api_version:?}, kind: \
+                {kind:?}, metadata.name: {name:?}, metadata.namespace: {namespace:?}) did not \
+                match any object created for this cluster and therefore had no effect. Please \
+                check that apiVersion, kind and metadata.name are correct and that \
+                metadata.namespace is set to {cluster_namespace:?}.",
+                cluster_namespace = self.namespace,
+            );
```
Techassi: We generally prefer providing dynamic values as fields/attributes (by mentioning them before the message text in the `warn!` macro). This allows consumers (like OpenTelemetry collectors) to access those fields as structured data instead of only having access to a message string.

crates/stackable-operator/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- `ClusterResources` now warns about `objectOverrides` entries that did not match any of the objects it created ([#1264]).
+- BREAKING: To enable this, `apply_deep_merge` now returns whether the merge matched the base object and `ObjectOverrides::apply_to`
+  returns the indices of the entries that matched.
```
Techassi: We usually add changes as a single entry (or split place multiple entries into multiple sections).

So I would recommend combining this entry into a single one or split it up into "Added" and "Changed" section.

crates/stackable-operator/src/cluster_resources.rs
sweb: that makes sense - thank you! Addressed in https://github.com/stackabletech/operator-rs/pull/1264/commits/c48736ce4943110fc03b78a25271e111c4eafd68

crates/stackable-operator/CHANGELOG.md
sweb: addressed in https://github.com/stackabletech/operator-rs/pull/1264/commits/6033b8771af748e43f8c9ee878fc708847a25b8a

sweb:
@Techassi could you take another look?

crates/stackable-operator/src/cluster_resources.rs
Techassi: This is already better, but I would recommend following semantic conventions.

- The only official semantic convention is `k8s.namespace.name`.
- The `k8s.object.api_version`, `k8s.object.kind`, `k8s.object.name` attributes are not official, but still make sense.
- Instead of using a plain `index` attribute, we could make it clearer what the index is talking about. I propose `object_override.index` or similar.
- For (product) cluster related attributes we have no concrete conventions (but probably should).

```suggestion
            warn!(
                object_override.index = index,
                k8s.object.api_version = api_version,
                k8s.object.kind = kind,
                k8s.object.name = name,
                k8s.namespace.name = namespace,
                cluster_namespace = self.namespace,
                "objectOverride did not match any object created for this cluster and therefore had \
                no effect. Please check that apiVersion, kind and metadata.name are correct and that \
                metadata.namespace matches the cluster namespace."
            );
```

Techassi (changes requested):

crates/stackable-operator/src/cluster_resources.rs
sweb: @Techassi thanks - that makes sense!

Techassi (dismissed):

Techassi (approved):

crates/stackable-operator/src/cluster_resources.rs
Maleware: All right, thanks!

lfrancke:
I know this is in operator-rs but once it lands in the operators...should this have release notes?

PR #1265, opened 2026-08-21
fix: wait_for_termination returns actual exit status

sweb:
# Description

since `set -e` is the last part of `wait_for_termination` it always has exit code 0. With this, it returns the exit code of the process it is waiting for instead. `[ "${term_child_pid}" ]` blows up when the child pid is not set yet, so gets changed to `[ -n "${term_child_pid:-}" ]`.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-operator/src/crd/git_sync/v1alpha1_impl.rs
+++ b/crates/stackable-operator/src/crd/git_sync/v1alpha1_impl.rs
@@ -494,7 +494,7 @@ mod tests {
 
   handle_term_signal()
   {
-      if [ "${term_child_pid}" ]; then
+      if [ -n "${term_child_pid:-}" ]; then
           kill -TERM "${term_child_pid}" 2>/dev/null
       else
           term_kill_needed="yes"
@@ -509,9 +509,14 @@ mod tests {
           kill -TERM "${term_child_pid}" 2>/dev/null
       fi
       wait ${term_child_pid} 2>/dev/null
+      term_child_status=$?
       trap - TERM
-      wait ${term_child_pid} 2>/dev/null
+      if [ "${term_child_status}" -gt 128 ]; then
+          wait ${term_child_pid} 2>/dev/null
+          term_child_status=$?
+      fi
       set -e
+      return ${term_child_status}
   }
 
   prepare_signal_handlers
@@ -563,7 +568,7 @@ volumeMounts:
 
   handle_term_signal()
   {
-      if [ "${term_child_pid}" ]; then
+      if [ -n "${term_child_pid:-}" ]; then
           kill -TERM "${term_child_pid}" 2>/dev/null
       else
           term_kill_needed="yes"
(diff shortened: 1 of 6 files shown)


lfrancke:
Ah...we just had the same/similar in https://github.com/stackabletech/stackable-utils/pull/130

It's an annoying trap.

sweb:
> Ah...we just had the same/similar in [stackabletech/stackable-utils#130](https://github.com/stackabletech/stackable-utils/pull/130)
> 
> It's an annoying trap.

in this context, it is not a huge deal - from what I checked it does not interfere with OOM based restarts or something like that - but getting the actual exit codes may be nice and should be an improvement.

crates/stackable-operator/src/crd/git_sync/v1alpha1_impl.rs
```diff
@@ -509,9 +509,14 @@ mod tests {
           kill -TERM "${term_child_pid}" 2>/dev/null
       fi
       wait ${term_child_pid} 2>/dev/null
+      term_child_status=$?
       trap - TERM
-      wait ${term_child_pid} 2>/dev/null
+      if [ "${term_child_status}" -gt 128 ]; then
```
maltesander, line 514: Can we add a flag / comment for the 128?

maltesander (changes requested):
A couple of things. Almost all downstream products require changes:
- https://github.com/stackabletech/trino-operator/blob/77bcf6da54ac49fb2d896388f791611f79984a50/rust/operator-binary/src/controller/build/command.rs#L145
- https://github.com/stackabletech/nifi-operator/blob/4d6cea98741fb34c29d28ae4e13931dcb317d817/rust/operator-binary/src/controller/build/resource/statefulset.rs#L404

Will abort under `set -e` with non-zero code, neither `create_vector_shutdown_file_command` or other lines after that will be executed (e.g. the vector shutdown file etc.). So `|| product_exit_code=$?` should be added and rolled out with an operator-rs release properly?

And i think a similar problem for git-sync containers https://github.com/stackabletech/operator-rs/blob/4e5b18690a2fd121bc926dfd8df2748d85bb4a23/crates/stackable-operator/src/crd/git_sync/v1alpha1_impl.rs#L355

---

It could make sense to stop that test script duplication (i think its in over 10 places) and e.g. use insta (already in the repo) as dev/test dependency? Something like:

```
insta::assert_snapshot!(
    serde_yaml::to_string(&git_sync_resources.git_sync_containers.first()).unwrap()
);
```

crates/stackable-operator/src/crd/git_sync/v1alpha1_impl.rs
sweb: addressed here: https://github.com/stackabletech/operator-rs/pull/1265/commits/f3a4feee6a72690879865f771bb7d82a6c49df36

still looking into `insta`.

PR #1271, opened 2026-09-04
fix(operator): Add missing SecurityContextBuilder::build function

Techassi:
This was discovered when I implemented <https://github.com/stackabletech/stackablectl/pull/453> and had to deal with the `SecurityContextBuilder` changes from <https://github.com/stackabletech/operator-rs/pull/1205>.

### DIFF
--- a/crates/stackable-operator/src/builder/pod/security.rs
+++ b/crates/stackable-operator/src/builder/pod/security.rs
@@ -9,6 +9,8 @@ pub struct SecurityContextBuilder {
     security_context: SecurityContext,
 }
 
+// FIXME (@Techassi): These associated function should take `self`` instead of `&mut self` for
+// better chainability.
 impl SecurityContextBuilder {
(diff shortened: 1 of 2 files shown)


crates/stackable-operator/src/builder/pod/security.rs
```diff
@@ -9,6 +9,8 @@ pub struct SecurityContextBuilder {
     security_context: SecurityContext,
 }
 
+// FIXME (@Techassi): These associated functions should take `self` instead of `&mut self` for
```
lfrancke, line 12: What's stopping you from changing that now?

Techassi: Technically, nothing. But that would be a breaking API change. I'm happy to do it in a follow-up PR.

Techassi: I noted it down locally. I will now mark this thread as resolved.

NickLarsenNZ (approved):
LGTM

PR #1272, opened 2026-09-08
feat: Add `/ready` endpoint to WebhookServer

xeniape:
# Description

Part of https://github.com/stackabletech/issues/issues/828

This PR adds a `/ready` endpoint to the WebhookServer to be used by startup probes in the operators. It was added to the WebhookServer as per requirements of the ticket above, for the future and if more readiness checks come up unrelated to the CRDs and/or WebhookServer, it might be reasonable to start a separate server instead. Currently, if no webhooks were defined, the ready endpoint also would not be served (not the case with the current operators, as they all have webhooks).

As part of this the `crd_established` signal was also adjusted (timeout removed) because the timeout there would bite with the probes configuration. Instead of shutting down the whole controller on a crd_established timeout, the startup probe would now handle the checking and restarting.

The implementation was tested with the zookeeper-operator, changes there would then only include creating and passing a HealthCheckRegistry and marking checks as passed on crd_established signals. (few lines) -> part of the rollout later
After the rollout to the operators, the probes need to be added to the operator Deployments.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/crates/stackable-shared/src/health.rs
+++ b/crates/stackable-shared/src/health.rs
@@ -1,0 +1,114 @@
+use std::{
+    fmt::Display,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+};
+
+/// A single named check contributing to one health endpoint.
+///
+/// A check only passes once [`HealthCheck::mark_passed`] has been called.
+#[derive(Clone)]
+pub struct HealthCheck {
+    name: String,
+    passed: Arc<AtomicBool>,
+}
+
+impl HealthCheck {
+    fn new(name: impl Into<String>) -> Self {
+        Self {
+            name: name.into(),
+            passed: Arc::new(AtomicBool::new(false)),
+        }
+    }
+
+    pub fn mark_passed(&self) {
+        self.passed.store(true, Ordering::Release);
+    }
+
+    fn passed(&self) -> bool {
+        self.passed.load(Ordering::Acquire)
+    }
+}
+
+/// A set of checks to be used for a health endpoint a probe can call.
+///
+/// # Example
+///
+/// ```
+/// use stackable_shared::health::HealthCheckRegistry;
+///
+/// let mut startup_checks = HealthCheckRegistry::new();
+/// let crds_established = startup_checks.register("crds-established");
+///
+/// assert!(!startup_checks.all_passed());
+/// crds_established.mark_passed();
+/// assert!(startup_checks.all_passed());
+/// ```
+#[derive(Default)]
+pub struct HealthCheckRegistry {
+    checks: Vec<HealthCheck>,
+}
+
+impl HealthCheckRegistry {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Registers a new [`HealthCheck`] with the provided name and returns it.
+    pub fn register(&mut self, name: impl Into<String>) -> HealthCheck {
+        let check = HealthCheck::new(name);
+        self.checks.push(check.clone());
+        check
+    }
+
+    /// Returns `true` if all the registered health checks have passed or no health checks are
+    /// registered.
+    pub fn all_passed(&self) -> bool {
+        self.checks.iter().all(HealthCheck::passed)
+    }
+}
+
+impl Display for HealthCheckRegistry {
+    /// Renders one line per check, with the check's name and status only. Anything else, error
+    /// causes in particular, must not end up in a response to an unauthenticated endpoint.
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.checks.is_empty() {
+            return writeln!(f, "[ok] no checks registered");
+        }
+
+        for check in &self.checks {
+            let status = if check.passed() { "ok" } else { "pending" };
+            writeln!(f, "[{status}] {name}", name = check.name)?;
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn passed_on_empty_registry() {
+        let registry = HealthCheckRegistry::new();
+
+        assert!(registry.all_passed());
+    }
+
+    #[test]
+    fn passed_only_once_every_check_is() {
+        let mut registry = HealthCheckRegistry::new();
+        let crds = registry.register("crds-established");
+        let migration = registry.register("database-migrated");
+
+        assert!(!registry.all_passed());
+
+        crds.mark_passed();
+        assert!(!registry.all_passed());
+
+        migration.mark_passed();
+        assert!(registry.all_passed());
+    }
+}
--- a/crates/stackable-webhook/src/lib.rs
+++ b/crates/stackable-webhook/src/lib.rs
@@ -11,13 +11,17 @@
 //!
 //! For usage please look at the [`WebhookServer`] docs as well as the specific [`Webhook`] you are
 //! using.
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::{
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    sync::Arc,
+};
 
 use ::x509_cert::Certificate;
-use axum::{Router, routing::get};
+use axum::{Router, http::StatusCode, routing::get};
 use futures_util::TryFutureExt;
 use k8s_openapi::ByteString;
 use snafu::{ResultExt, Snafu};
+use stackable_shared::health::HealthCheckRegistry;
 use stackable_telemetry::AxumTraceLayer;
 use tokio::{sync::mpsc, try_join};
 use tower::ServiceBuilder;
@@ -54,6 +58,7 @@ pub enum WebhookServerError {
 /// ### Example usage
 ///
 /// ```
+/// use stackable_shared::health::HealthCheckRegistry;
 /// use stackable_webhook::{WebhookServer, WebhookServerOptions, webhooks::Webhook};
 /// use tokio::time::{Duration, sleep};
 ///
@@ -65,7 +70,10 @@ pub enum WebhookServerError {
 ///     webhook_namespace: "my-namespace".to_owned(),
 ///     webhook_service_name: "my-operator".to_owned(),
 /// };
-/// let webhook_server = WebhookServer::new(webhooks, webhook_options).await.unwrap();
+/// let readiness_checks = HealthCheckRegistry::new();
+/// let webhook_server = WebhookServer::new(webhooks, webhook_options, readiness_checks)
+///     .await
+///     .unwrap();
 /// let shutdown_signal = sleep(Duration::from_millis(100));
 ///
 /// webhook_server.run(shutdown_signal).await.unwrap();
@@ -111,6 +119,7 @@ impl WebhookServer {
     pub async fn new(
         webhooks: Vec<Box<dyn Webhook>>,
         options: WebhookServerOptions,
+        readiness_checks: HealthCheckRegistry,
     ) -> Result<Self> {
         tracing::trace!("create new webhook server");
 
@@ -132,12 +141,26 @@ impl WebhookServer {
             router = webhook.register_routes(router);
         }
 
+        // Create the route handler for the startup probe.
+        let readiness_checks = Arc::new(readiness_checks);
+        let ready_route = move || async move {
+            let status = if readiness_checks.all_passed() {
+                StatusCode::OK
+            } else {
+                StatusCode::SERVICE_UNAVAILABLE
+            };
+            // The response body carries check names and their status. Error causes etc. go to the
+            // log, never into a response to not leak internal information to the public endpoint.
+            (status, readiness_checks.to_string())
+        };
+
         let router = router
             // Enrich spans for routes added above.
             // Routes defined below it will not be instrumented to reduce noise.
             .layer(trace_service_builder)
-            // The health route is below the AxumTraceLayer so as not to be instrumented
-            .route("/health", get(|| async { "ok" }));
+            // The health and ready routes are below the AxumTraceLayer so as not to be instrumented
+            .route("/health", get(|| async { "ok" }))
+            .route("/ready", get(ready_route));
 
         tracing::debug!("create TLS server");
         let (tls_server, cert_rx) = TlsServer::new(router, &options)
--- a/crates/stackable-operator/CHANGELOG.md
+++ b/crates/stackable-operator/CHANGELOG.md
@@ -8,7 +8,14 @@ All notable changes to this project will be documented in this file.
 
 - BREAKING: `ClusterResources` now warns about `objectOverrides` entries that did not match any of the objects it created.  To enable this, the signatures of `apply_deep_merge` and `ObjectOverrides::apply_to` needed to be adjusted ([#1264]).
 
+### Removed
+
+- BREAKING: Removed `timeout_duration` parameter from `signal::crd_established`. It now waits indefinitely and
+  the timeout should be handled with a startup probe instead. As a consequence `DEFAULT_CRD_ESTABLISHED_TIMEOUT`
+  also got removed ([#]).
+
 [#1264]: https://github.com/stackabletech/operator-rs/pull/1264
+[#]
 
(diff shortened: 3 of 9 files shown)


crates/stackable-webhook/src/lib.rs
```diff
@@ -111,6 +119,7 @@ impl WebhookServer {
     pub async fn new(
         webhooks: Vec<Box<dyn Webhook>>,
         options: WebhookServerOptions,
+        readiness_checks: HealthCheckRegistry,
```
Techassi, line 122: The readiness checks being part of the public API makes me feel like the whole `HealthCheck` and `HealthCheckRegistry` machinery should be part of the stackable-webhook crate.

crates/stackable-webhook/src/lib.rs
```diff
@@ -132,12 +141,26 @@ impl WebhookServer {
...
+        let readiness_checks = Arc::new(readiness_checks);
+        let ready_route = move || async move {
+            let status = if readiness_checks.all_passed() {
+                StatusCode::OK
+            } else {
+                StatusCode::SERVICE_UNAVAILABLE
+            };
+            // The response body carries check names and their status. Error causes etc. go to the
+            // log, never into a response to not leak internal information to the public endpoint.
+            (status, readiness_checks.to_string())
```
Techassi, line 154: We could potentially implement axum's `IntoResponse` trait here which internally calls `to_string()`.

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+};
+
+/// A single named check contributing to one health endpoint.
+///
+/// A check only passes once [`HealthCheck::mark_passed`] has been called.
+#[derive(Clone)]
+pub struct HealthCheck {
+    name: String,
+    passed: Arc<AtomicBool>,
+}
```
Techassi, line 16: The whole machinery could be simplified a little. I will add my suggestions to each piece of code individually.

Here, we only need access to the underlying AtomicBool. I also renamed the struct to better reflect what it is: a handle to mark a readiness check as ready.

```suggestion
pub struct ReadinessHandle(Arc<AtomicBool>)
```

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    pub fn mark_passed(&self) {
+        self.passed.store(true, Ordering::Release);
+    }
+
+    fn passed(&self) -> bool {
+        self.passed.load(Ordering::Acquire)
+    }
+}
```
Techassi, line 33: The impl block can also be simplified a bunch:

```suggestion
impl ReadinessHandle {
    pub fn ready(self) {
        self.0.store(true, Ordering::Release);
    }
}
```

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+/// let crds_established = startup_checks.register("crds-established");
+///
+/// assert!(!startup_checks.all_passed());
+/// crds_established.mark_passed();
+/// assert!(startup_checks.all_passed());
+/// ```
+#[derive(Default)]
+pub struct HealthCheckRegistry {
+    checks: Vec<HealthCheck>,
+}
```
Techassi, line 52: We now use a Vec over a tuple instead:

```suggestion
// This has to be an AtomicBool as we could otherwise not share references to it.
pub struct ReadinessChecks(Vec<(String, Arc<AtomicBool>)>);
```

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl HealthCheckRegistry {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
```
Techassi, line 58: If both a `new` method and a Default impl exist, the Default impl should delegate to `new`, not the other way around (see [CLippy lint](https://rust-lang.github.io/rust-clippy/stable/index.html?search=new_#new_without_default)). In this case, I would argue we can even remove the `new` method.

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+        self.checks.push(check.clone());
+        check
+    }
+
+    /// Returns `true` if all the registered health checks have passed or no health checks are
+    /// registered.
+    pub fn all_passed(&self) -> bool {
+        self.checks.iter().all(HealthCheck::passed)
+    }
+}
```
Techassi, line 71: Because we split the health check into its name and AtomicBool, we only have to clone the Arc (increase its reference count). I also renamed the `all_passed` method to `all_ready`.

```suggestion
impl ReadinessChecks {
    /// Registers a new readiness check with the provided name.
    ///
    /// The returned handle can be used to mark the check as ready.
    pub fn register(&mut self, name: impl Into<String>) -> ReadinessHandle {
        let ready = Arc::new(AtomicBool::default());
        // Store an reference counted clone of the same underlying AtomicBool in the list of checks.
        // Both the handle and the item in the list refer to the same AtomicBool.
        self.0.push((name.into(), ready.clone()));
        ReadinessHandle(ready)
    }

    /// Returns `true` if all the registered checks are ready or no checks are registered.
    pub fn all_ready(&self) -> bool {
        self.0
            .iter()
            .all(|(_, ready)| ready.load(Ordering::Acquire))
    }
}
```

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+            return writeln!(f, "[ok] no checks registered");
+        }
+
+        for check in &self.checks {
+            let status = if check.passed() { "ok" } else { "pending" };
+            writeln!(f, "[{status}] {name}", name = check.name)?;
+        }
+        Ok(())
+    }
+}
```
Techassi, line 87: We now have to adjust the handling and loading of the statuses:

```suggestion
impl Display for ReadinessChecks {
    /// Renders one line per check, with the check's name and status only. Anything else, error
    /// causes in particular, must not end up in a response to an unauthenticated endpoint.
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        if self.0.is_empty() {
            return writeln!(f, "[ok] no readiness checks registered");
        }

        for (name, ready) in &self.0 {
            let status = if ready.load(Ordering::Acquire) {
                "ready"
            } else {
                "pending"
            };
            writeln!(f, "[{status}] {name}")?;
        }
        Ok(())
    }
}
```

crates/stackable-shared/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        assert!(!registry.all_passed());
+
+        crds.mark_passed();
+        assert!(!registry.all_passed());
+
+        migration.mark_passed();
+        assert!(registry.all_passed());
+    }
+}
```
Techassi, line 114:
```suggestion
mod test {
    use super::*;

    #[test]
    fn ready_on_empty_registry() {
        let checks = ReadinessChecks::default();

        assert!(checks.all_ready());
    }

    #[test]
    fn ready_only_once_every_check_is() {
        let mut checks = ReadinessChecks::default();
        let crds = checks.register("crds-established");
        let migration = checks.register("database-migrated");

        assert!(!checks.all_ready());

        crds.ready();
        assert!(!checks.all_ready());

        migration.ready();
        assert!(checks.all_ready());
    }
}
```

Techassi (changes requested):
I like the overall mechanism of this, nicely done! I only have a few suggestions to slightly optimize it. And I also have a few thoughts on where code is located.

All my suggestions target only code, so when (and if) they are applied, the doc tests need to be adjusted as well.

crates/stackable-webhook/src/lib.rs
xeniape: Yeah, wasn't completely sure where to put it. I decided it based on the possibility of moving the whole health checks into a separate health server unrelated to the webhook server in the future, since there might be other things later on like "migration done" check or "external server reachable" check (just some random examples) which have nothing to do with our webhook here.

crates/stackable-webhook/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+};
+
+/// A single named check contributing to one health endpoint.
+///
+/// A check only passes once [`HealthCheck::mark_passed`] has been called.
+#[derive(Clone)]
+pub struct HealthCheck {
+    name: String,
+    passed: Arc<AtomicBool>,
+}
```
xeniape: I'll look into all the improvement suggestions now regarding the code, but with the naming I would keep it that way unless you strongly disagree. The health checks could possibly also be used for liveness probes etc. later on, it was meant as a general construct for that. It also is more closely to the Kubernetes naming of those checks which are then also used by readiness/liveness/startup probes etc.

xeniape: Hm not sure what the code suggestion improves here. For me it reads worse, the name and status are basically "separated" then instead of being part of the same concept struct. Also in case we want to extend that later (I don't know, add timestamp field when health check last passed or something) it would be easier with the current approach, but feel free to ignore that point since it's just a "maybe scenario".

crates/stackable-webhook/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    pub fn mark_passed(&self) {
+        self.passed.store(true, Ordering::Release);
+    }
+
+    fn passed(&self) -> bool {
+        self.passed.load(Ordering::Acquire)
+    }
+}
```
xeniape: This also kind of reads worse for me (`self.0` instead of `self.passed`). Also `ready` implies that it's a check rather than a write operation.

crates/stackable-webhook/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+/// let crds_established = startup_checks.register("crds-established");
+///
+/// assert!(!startup_checks.all_passed());
+/// crds_established.mark_passed();
+/// assert!(startup_checks.all_passed());
+/// ```
+#[derive(Default)]
+pub struct HealthCheckRegistry {
+    checks: Vec<HealthCheck>,
+}
```
xeniape: This is kind of related to https://github.com/stackabletech/operator-rs/pull/1272#discussion_r4060845537 also, and I think some of other suggestions might be derived from the decision which approach to use here. I first started out with the Vec version, but again, it seemed better readable and extensible to me with the current implemented version and it didn't seem to have any performance downsides etc. Or what would be the benefits I might not see to using the Vec version?

crates/stackable-webhook/src/health.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl HealthCheckRegistry {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
```
xeniape: The lint you linked was about *not* having a default implementation. And the current code also aligns with all the other places in the code already using it the same way. I assume dropping the `new` method would mean we construct with `HealthCheckRegistry::default()` then? Which to me also also sounds a bit worse in the readability direction ("what are the defaults here?").

crates/stackable-webhook/src/lib.rs
Techassi: This goes back to the thing I mentioned in a DM. We most likely want to widen the scope of this crate to **not** only be about webhooks, but about **all** HTTP related stuff.

It already is kind of built to be extensible anyway, just not to fully support all arbitrary HTTP handlers yet.

crates/stackable-webhook/src/health.rs
Techassi: In my opinion it is fine to separate the name/identifier from the actual check (which can of course contains more than just a bool indicating if the check passed or not) as long as the check it self does not need it (which is currently the case).

crates/stackable-webhook/src/lib.rs
xeniape: Oh I see! Yeah, if that is the plan for that crate then it sounds like the right place for it, I will move it.

crates/stackable-webhook/src/health.rs
Techassi: > This also kind of reads worse for me (`self.0` instead of `self.passed`)

If it is only about the name of the field, then we could bring that back. But having a single field struct is kind of a smell - that's why I opted to use the tuple form.

> Also `ready` implies that it's a check rather than a write operation.

I'm fine making it more explicit, like `mark_ready` - I just liked the short name.

crates/stackable-webhook/src/lib.rs
xeniape: Oh I see! Yeah, if that is the plan for that crate then it sounds like the right place for it, I will move it.

crates/stackable-webhook/src/health.rs
Techassi: > I first started out with the Vec version, but again, it seemed better readable and extensible to me with the current implemented version and it didn't seem to have any performance downsides etc.

Am I missing something here? The current implementation **and** my suggestion both use a Vec internally - just the inner type differs.

Are you talking about using a named field vs the tuple approach?

crates/stackable-webhook/src/health.rs
Techassi: > I assume dropping the new method would mean we construct with `HealthCheckRegistry::default()` then?

Yes, exactly.

> Which to me also also sounds a bit worse in the readability direction ("what are the defaults here?").

I mean it is literally the definition of the `Default` impl to say "I don't care about the exact defaults, I just want a new default instance of this type". If you care about individual values, then the `new` function is usually the way to go.

Also, what is currently the difference between calling `new` or `default`? They are literally running the same piece of code. Users currently already don't know what defaults are being used when calling `new`.

crates/stackable-webhook/src/health.rs
xeniape: Yes I meant the difference between the type (in the current implementation) and the tuple (your suggestion), sorry.

crates/stackable-webhook/src/health.rs
xeniape: I guess `new` to me reads more like an "empty" registry, while `default` might be already prefilled with something. Which theoretically could change later and the user of that function wouldn't notice it. But maybe that's too philosophical :smile: I can change it, just making sure it's fine to divert from the current code convention.

crates/stackable-webhook/src/lib.rs
xeniape: Changed everything we talked about in a call in https://github.com/stackabletech/operator-rs/pull/1272/commits/e572370b3708f6371fd3a7460807ab98f6a58197 (also incorporating the other comments)

crates/stackable-webhook/src/health.rs
```diff
@@ -0,0 +1,164 @@
...
+//!
+//! The naming here follows the Kubernetes convention of a health check / health check registry and
+//! their usage by the different probes.
+//!
+//! ## References
+//!
+//! - <https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go#L41>
+//! - <https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/healthz.go#L34>
+//! - <https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go#L205>
+//! - <https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go#L330>
```
Techassi: These references are very nice, but I would advise to use links to a specific tag or even commit to ensure these links stay valid even if code changes on `master`.

Techassi (dismissed):

crates/stackable-webhook/src/lib.rs
Techassi: Done in e572370.

crates/stackable-webhook/src/health.rs
xeniape: https://github.com/stackabletech/operator-rs/pull/1272/commits/c330da176ab000a1c53a71f3ccab4365f3254e83

Techassi (approved):

PR #1275, opened 2026-09-10
refactor(operator): Improve pull policy handling/warning

Techassi:
Follow-up of #1226, part of https://github.com/stackabletech/internal-issues/issues/169.

This adds the same warning to the `custom` path which previously was missing this. We now define the pull policy handling/determination in a single place as a dedicated function.


NickLarsenNZ (approved):
LGTM

PR #1277, opened 2026-09-16
chore: Make cargo doc fail on warnings

Techassi:
Triggered by https://github.com/stackabletech/issues/issues/848#issuecomment-4915255906, similar to https://github.com/stackabletech/operator-templating/pull/650.

`cargo doc` hooks run by prek now error on warnings.


NickLarsenNZ (approved):
Hmm, I find warnings useful for deprecating things without a breaking change.
I'll approve it since we already do that, but we should maybe look at relaxing that rule (it could be a separate check that is allowed to fail so at least there is feedback).

